PR #129, opened 2025-04-16
chore(internal): configure releases

meorphis:
configures openai-ruby releases similar to how we have them set-up for other languages, e.g. https://github.com/openai/openai-node/blob/next/.github/workflows/create-releases.yml

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   release:
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -1,7 +1,5 @@
 name: Create releases
 on:
-  schedule:
-    - cron: '0 5 * * *' # every day at 5am UTC
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/create-releases.yml
```diff
@@ -0,0 +1,42 @@
+name: Create releases
+on:
+  schedule:
+    - cron: '0 5 * * *' # every day at 5am UTC
```
ms-jpq, line 4: not sure if we need to publish at 5 AM every day haha, but everything else looks sound!

ms-jpq (approved):
can you please remove the daily cron job and its all good!

.github/workflows/create-releases.yml
```diff
@@ -1,7 +1,5 @@
 name: Create releases
 on:
-  schedule:
```
ms-jpq: nit: no biggie, but having a manual release trigger wouldnt hurt!

I think that's what the [cron job](https://github.com/openai/openai-python/blob/ed53107e10e6c86754866b48f8bd862659134ca8/.github/workflows/create-releases.yml#L4) in the python-sdk functionally does actually

PR #134, opened 2025-04-29
Fix warning: JSON.fast_generate is deprecated

sferik:
This library will break as soon as version 3 of the JSON gem is released.


ms-jpq:
thank you so much!

PR #135, opened 2025-04-29
Bump minimum supported Ruby version to 3.2

sferik:
Ruby 3.1 is no longer supported as of March 31, 2025. See https://endoflife.date/ruby.


ms-jpq:
thank you! that is very reasonable.

PR #137, opened 2025-05-02
Remove unnecessary require statement

sferik:
Now that https://github.com/openai/openai-ruby/pull/135 has been merged, it's safe to remove this `require` statement. In Ruby 3.2, the Set class is built-in and is automatically loaded when you use the `Set` constant or call `Enumerable#to_set`.


ms-jpq:
sounds nice! 👍

PR #139, opened 2025-05-12
Fix all linter warnings and set stricter fail level

sferik:


ms-jpq:
thanks @sferik , I don't know how I missed your PR here last week. sorry about the really slow turn around!

Issue #126, opened 2025-04-16
Beta Feedback and Feature Requests!


ms-jpq:
# General discussion for beta feedback and feature requests

- Collection for general feedback and feature requests so we can prioritize.

- Please feel empowered to start a new issue for more substantial areas of discussion.

- Prioritization will be adjusted in accordance with community interest

agarcher:
I'm having a trouble getting the gem to load properly in a rails environment. When adding the gem and trying to blindly use it, I was getting `NameErrors` because the `OpenAI` namespace wasn't initialized (the gem wasn't auto-loaded). I was able to workaround this with a `require "openai"` statement in `application.rb` or by creating an initializer with the require that also accessed the constant (a noop call to `OpenAI` which is kind of weird). I think the second workaround stopped working with the latest changes since yesterday.

However, these workarounds don't solve the loading issues when tapioca tries to load the rails application to build rbis.

```
❯ bundle exec tapioca dsl                                                                                                                             Loading DSL extension classes... Done
Loading Rails application...
Tapioca attempted to load the Rails application after encountering a `config/application.rb` file, but it failed. If your application uses Rails please ensure it can be loaded correctly before generating RBIs. If your application does not use Rails and you wish to continue RBI generation please pass
 `--no-halt-upon-load-error` to the tapioca command in sorbet/tapioca/config.yml or in CLI.
uninitialized constant OpenAI
bundler: failed to load command: tapioca (/Users/agarcher/.rbenv/versions/ruby-3.3.0/bin/tapioca)
...
```

The rest of the error just shows the first place in my application that it hits a reference to the `OpenAI` namespace, which is a schema class I define that extends `OpenAI < BaseModel`

Any suggestions for how to properly load the gem in a rails environment?

I tried adding `require "bundler/setup"` as well, as per the usage instructions, but it didn't help.

agarcher:
@ms-jpq this gets messy quickly because I also reference `OpenAI::BaseModel` `OpenAI::ArrayOf`, etc. etc. I would need to define stubs of these classes, and I would need to do them only when running tapioca (so I don't conflict with the real gem at runtime). If I'm understanding correctly, it seems like you only need the workaround when the user is running `tapioca gem`? Perhaps the `grep` on the workaround could be more specific?

agarcher:
[...]

PR #149, opened 2025-05-23
Only skip loading classes on 'tapioca gem'

agarcher:
Addresses the class loading issue when running `bundle exec tapioca dsl` in an application that depends on this gem as discussed in https://github.com/openai/openai-ruby/issues/126#issuecomment-2905149023

### DIFF
--- a/lib/openai.rb
+++ b/lib/openai.rb
@@ -14,10 +14,15 @@ require "securerandom"
 require "stringio"
 require "time"
 require "uri"
+
 # We already ship the preferred sorbet manifests in the package itself.
 # `tapioca` currently does not offer us a way to opt out of unnecessary compilation.
-if Object.const_defined?(:Tapioca) && caller.chain([$PROGRAM_NAME]).chain(ARGV).grep(/tapioca/)
-  return
+if Object.const_defined?(:Tapioca)
+  call_stack_array = caller.chain([$PROGRAM_NAME]).chain(ARGV).to_a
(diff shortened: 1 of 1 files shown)


lib/openai.rb
```diff
@@ -14,10 +14,15 @@
...
 require "uri"
+
 # We already ship the preferred sorbet manifests in the package itself.
 # `tapioca` currently does not offer us a way to opt out of unnecessary compilation.
-if Object.const_defined?(:Tapioca) && caller.chain([$PROGRAM_NAME]).chain(ARGV).grep(/tapioca/)
-  return
+if Object.const_defined?(:Tapioca)
+  call_stack_array = caller.chain([$PROGRAM_NAME]).chain(ARGV).to_a
+  last_tapioca_index = call_stack_array.rindex { |line| line.include?("tapioca") }
+  if last_tapioca_index && last_tapioca_index < call_stack_array.length - 1 && call_stack_array[last_tapioca_index + 1].strip == "gem"
```
ms-jpq, line 23: humm, I think think this might be a little too targeted?

we also want to skip introspection during `tapioca init`.

I am going to prototype this a little more and see what can be done!

ms-jpq (commented):
thank you so much for the quick PR on fixing this issue. I will take a deeper look into how tapioca works before I merge a fix to this issue though.

Since they did promise we could hack into their internal introspection but "no promises if its not going to break one day"

agarcher:
Closing as this was addressed in the last release. Thank you!

Issue #180, opened 2025-08-18
Sorbet has violations when executing via Tapioca


SheldonNunes:
I am getting the following Internal Error when trying to perform tapioca typechecking on this package

```
##### INTERNAL ERROR #####

There are parse errors in the generated RBI files.

This seems related to a bug in Tapioca.
Please open an issue at https://github.com/Shopify/tapioca/issues/new with the following information:

Tapioca v0.17.7

Command:
  bin/tapioca gem

Gems:
 ....
  openai (0.17.1)
 ....

Errors:
  sorbet/rbi/gems/openai@0.15.0.rbi:156: An `attr_accessor` cannot be `void` (3516)
  sorbet/rbi/gems/openai@0.15.0.rbi:159: An `attr_accessor` cannot be `void` (3516)
  sorbet/rbi/gems/openai@0.15.0.rbi:162: An `attr_accessor` cannot be `void` (3516)
  sorbet/rbi/gems/openai@0.15.0.rbi:165: An `attr_accessor` cannot be `void` (3516)
  sorbet/rbi/gems/openai@0.15.0.rbi:168: An `attr_accessor` cannot be `void` (3516)

##########################
```

It appears that there are some `void` `attr_accessor` attributes which are causing issues with Tapioca.

ms-jpq:
closing as addressed by next release 👍 .

thank you for the report!

PR #181, opened 2025-08-19
Fix APIConnectionError attr_accessor rbi definition

SheldonNunes:
* Closes https://github.com/openai/openai-ruby/issues/180

This PR fixes an error being encountered when trying to execute tapioca/sorbet on this gem.

In the event of `APIConnectionError` the attr_accessors listed will return nil (not void)


SheldonNunes:
Closing this PR as I can see that the fix is included in the recent RBI regeneration

* https://github.com/openai/openai-ruby/pull/184

PR #185, opened 2025-08-22
Add example for using openai-ruby with Azure OpenAI

Cosmo:
This PR adds examples showing how to use `openai-ruby` with Azure OpenAI using the new [V1 API](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle?tabs=key).

It demonstrates:
- The required environment variables
- How to add the `api-version` query parameter
- Requesting the Chat Completion API
- Requesting the Images API

### DIFF
--- a/examples/azure_openai.rb
+++ b/examples/azure_openai.rb
@@ -1,0 +1,38 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+# typed: strong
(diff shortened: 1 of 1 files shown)


ms-jpq:
humm, i actually can't test this without getting myself an azure account, will do that first.

PR #262, opened 2026-03-27
feat: expose HTTP response metadata

ktamulonis:
## Summary

Adds an explicit raw-response modifier for generated HTTP operations:

```ruby
raw = openai.models.with_raw_response.retrieve("gpt-5.2")

puts(raw.status)
puts(raw.headers["openai-processing-ms"])
puts(raw.request_id)
puts(raw.read)

model = raw.parse
```

`OpenAI::RawResponse[T]` exposes the HTTP status, normalized immutable headers,
request ID, buffered body, and a lazy cached parsed value without changing the
operation's arguments or normal error behavior.

## API contract

- Ordinary top-level models and pages retain the `_request_id` convenience
  introduced by #352.
- `raw.request_id` and `raw.parse._request_id` are the same value when the
  parsed result is a model or page.
- Raw responses provide metadata for model, page, primitive, binary, and `nil`
  results.
- Full response headers are not attached to models, pages, nested values, or
  helper-assembled streaming snapshots.
- Higher-level streaming helpers continue to expose `status` and `headers`
  directly on their stream objects.
- Generated resource navigation and HTTP method signatures are preserved in
  Ruby, RBI, and RBS.

This follows the explicit raw-response pattern used by the other OpenAI SDKs
while keeping transport metadata separate from API-schema models.

## Generator ownership

The generated resource wrappers and type signatures are owned by the Castiron
source change in openai/openai#1272460. That companion PR prevents future SDK
generation from overwriting this API.

## Validation

- Ruby 4.0.6: 567 tests, 2,903 assertions
- Ruby 3.3.12: 567 tests, 2,903 assertions
- Raw-response contract: 9 tests, 1,001 assertions
- RuboCop: 2,608 files, no offenses
- Sorbet: no errors
- RBS: 1,214 files validated
- Gem build and package-content inspection
- Thermo-nuclear code-quality review


lib/openai/internal/type/base_model.rb
```diff
@@ -235,12 +235,21 @@ def hash = [self.class, @data].hash
...
-        # @param request_id [String, nil]
+        # @param headers [Hash{String=>String}]
         # @return [self]
-        def _set_request_id(request_id)
-          @_request_id = request_id
+        def _set_response_metadata(headers)
+          @_request_id = headers["x-request-id"]
+          @response_headers = headers
           self
         end
```
HAYDEN-OAI, line 245: Could we also keep this metadata out of Psych's `YAML.dump(response)` path? `response.to_yaml` uses the custom helper, but `YAML.dump` bypasses it and serializes instance variables, so this new `@response_headers` hash appears in the output. That contradicts the README's YAML guarantee and can expose headers from custom/proxied endpoints. An `encode_with` hook (or narrowing the documented guarantee) would close that gap.

HAYDEN-OAI (approved):
Overall looks good. I left one low-priority serialization-boundary comment inline.

lib/openai/internal/type/base_model.rb
jbeckwith-oai: Fixed in `4f33ccd3`. I added a Psych `encode_with` boundary that explicitly serializes only the model’s existing `data` and `coerced` state. This preserves the pre-feature object representation while excluding both `_request_id` and `response_headers`. The client regression test now exercises `YAML.dump(response)` directly and verifies that neither metadata field appears. Local Ruby 3.3/4.0 suites, RuboCop, Sorbet, RBS validation, and gem packaging all pass.

jbeckwith-oai:
Thank you for the response-header observability contribution. We are closing this PR while consolidating response metadata, HTTP logging, and timeout behavior into one replacement OpenAI Ruby PR. The original requirement and contributor attribution will be preserved in the consolidated change.

PR #275, opened 2026-05-19
Fix byte-limited reads for Enumerator-backed request bodies

### LINKED ISSUE #274: ReadIOAdapter#read returns too many bytes for Enumerator bodies
## Summary

`OpenAI::Internal::Util::ReadIOAdapter#read(max_len)` can return more than `max_len` bytes when wrapping an `Enumerator`. The buffered enumerator path currently slices with an inclusive range, so `read(3)` can return 4 bytes and `read(0)` can consume data.

## Reproduction

```ruby
adapter = OpenAI::Internal::Util::ReadIOAdapter.new(["abc", "def"].to_enum) {}

p adapter.read(0)
p adapter.read(1)
p adapter.read(3)
p adapter.read(99)
p adapter.read(1)
```

## Expected

```ruby
""
"a"
"bcd"
"ef"
nil
```

## Actual

```ruby
""
"ab"
"cdef"
""
nil
```

## Impact

This adapter is used for streamed request bodies, including multipart uploads. `IO#read(max_len)` is byte-oriented, so returning more than `max_len` bytes can violate callers' expectations when an upload body is backed by an `Enumerator`. The same code path should use byte lengths rather than character lengths for buffered reads.

fallintoplace:
## Summary

- make Enumerator-backed `ReadIOAdapter#read(max_len)` return at most `max_len` bytes
- use byte-oriented buffering so multibyte strings follow `IO#read` semantics
- preserve buffered bytes when an unbounded read follows a partial read
- match IO behavior for zero-length reads, EOF, negative lengths, and `out_string`
- add regression coverage for bounded reads, multibyte data, buffering, and output buffers

## Why

This adapter feeds streamed request bodies into IO consumers such as `IO.copy_stream`. Returning more bytes than requested can break those consumers and corrupt multipart upload boundaries; character-oriented slicing also makes the result depend on string encoding instead of the requested byte count.

Fixes #274.

## Test plan

- `bundle exec rake TEST=test/openai/internal/util_test.rb test`
- `bundle exec rake lint`

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -421,18 +421,38 @@ module OpenAI
         #
         # @param max_len [Integer, nil]
         #
-        # @return [String]
+        # @return [String, nil]
         private def read_enum(max_len)
           case max_len
           in nil
-            @stream.to_a.join
+            # `loop` rescues StopIteration, but this method handles it below.
+            # rubocop:disable Style/InfiniteLoop
+            @buf << @stream.next.b while true
+            # rubocop:enable Style/InfiniteLoop
+          in Integer if max_len.negative?
+            raise ArgumentError, "negative length #{max_len} given"
           in Integer
-            @buf << @stream.next while @buf.length < max_len
-            @buf.slice!(..max_len)
+            @buf << @stream.next.b while @buf.bytesize < max_len
+            read_buffer(max_len)
           end
         rescue StopIteration
+          return @buf.slice!(0..) if max_len.nil?
+
           @stream = nil
-          @buf.slice!(0..)
+          return nil if @buf.bytesize.zero?
+
+          read_buffer(max_len)
+        end
+
+        # @api private
+        #
+        # @param max_len [Integer]
+        #
+        # @return [String]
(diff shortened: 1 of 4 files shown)


jbeckwith-oai:
Thank you so much for the careful and comprehensive work here, @fallintoplace. Your analysis of byte-oriented `max_len` reads, preserving buffered data across bounded and unbounded reads, EOF and `out_string` semantics, and negative lengths became the backbone of the canonical fix.

Since this PR was opened, `ReadIOAdapter` moved into its SDK-owned file and `main` advanced substantially. We carried your work forward in #354, rebased it onto current `main`, combined it with the mixed-encoding fix for #317, preserved native `IO`/`StringIO` delegation, and added end-to-end and randomized contract validation. That replacement is now fully green across the Ruby matrix, lint, signatures, packaging, and CodeQL.

I’m closing this PR only because #354 supersedes it—not because the contribution wasn’t valuable. It materially shaped the final implementation and gave us the complete contract we needed. Really appreciate the time, rigor, and care you put into this. Thank you! 🙏

PR #289, opened 2026-06-22
[codex] Add AWS Bedrock provider authentication

HAYDEN-OAI:
## Summary

- add an `OpenAI::Providers.bedrock` provider for Amazon Bedrock's OpenAI-compatible API
- support bearer credentials, static AWS credentials, refreshable credential providers, named profiles, and the standard AWS credential chain, including `~/.aws/credentials`
- sign each request attempt with SigV4 and prevent signed requests from following redirects or authenticating a different origin
- isolate provider routing and authentication from ambient OpenAI client configuration
- add RBI/RBS types, setup documentation, a runnable example, a shared SigV4 fixture, and focused regression tests

## Why

The Ruby SDK did not have a supported provider-authentication path for AWS Bedrock, so applications could not use their normal AWS credentials file or runtime credential chain with the standard OpenAI client. This brings Ruby in line with the provider architecture used by the Node and Python SDKs.

Linear: [SDK-78](https://linear.app/openai/issue/SDK-78/support-for-aws-credentials-file-ruby)

## User impact

Ruby users can now construct the normal client with `OpenAI::Providers.bedrock(...)` and use AWS environment credentials, shared credentials/config files, named profiles, SSO or assume-role profiles, workload identities, explicit static credentials, or refreshable providers. AWS authentication loads `aws-sdk-core` as an optional dependency; bearer authentication does not require it.

Provider authentication is applied immediately before every transport attempt, so retries receive fresh credentials and a fresh signature over the exact serialized body and retry headers.

## Validation

- focused Bedrock, client, and workload-identity suites: 60 runs, 226 assertions, 0 failures, 0 errors
- Bedrock implementation coverage: 98.7% lines, 96.5% branches, 100% methods
- provider registry coverage: 100% lines, branches, and methods
- RuboCop and type checks pass
- `git diff --check` passes
- gem package isolation check passes

The full generated resource suite requires the local Prism mock server on `localhost:4010`; without it, those tests stop on expected connection errors. CI provisions that dependency.


lib/openai/providers/bedrock.rb
```diff
@@ -0,0 +1,480 @@
...
+          require("aws-sdk-core")
+          require("aws-sigv4")
+        rescue LoadError => e
+          raise OpenAI::Errors::Error.new(MISSING_DEPENDENCY_MESSAGE), cause: e
+        end
+
+        def resolve_base_url(base_url, region)
+          return base_url if base_url
+          raise OpenAI::Errors::Error, MISSING_REGION_MESSAGE if region.nil?
+          "https://bedrock-mantle.#{region}.api.aws/openai/v1"
```
jliccini: From Codex: The default Bedrock base URL should be `https://bedrock-mantle.<region>.api.aws/v1`, without the `/openai` segment. [AWS's Bedrock Mantle documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html) sets `OPENAI_BASE_URL` to that `/v1` root and shows requests at `$OPENAI_BASE_URL/models`; as written, default-configured clients will request the undocumented `/openai/v1/...` path. Please update this default and the corresponding documentation, tests, and SigV4 fixture.

jliccini (commented):
Inline review of the default Bedrock endpoint.

jliccini (approved):
Approving but noting that the URL thing Codex found might be important. I am not a Bedrock expert but the Ruby lgtm

lib/openai/providers/bedrock.rb
HAYDEN-OAI: thanks for flagging this! fixed in bc99d8c3. the provider now derives https://bedrock-mantle.<region>.api.aws/v1, and i updated the docs, canonical url tests, and regenerated sigv4 fixture. the focused suites pass with 60 runs and 226 assertions.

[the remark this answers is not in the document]
HAYDEN-OAI: good catch, thank you! fixed in bc99d8c3 by switching the canonical bedrock endpoint to /v1 and updating the docs, endpoint coverage, and shared sigv4 fixture. the focused suites pass with 60 runs and 226 assertions.

lib/openai/client.rb
```diff
@@ -229,25 +242,66 @@ class Client < OpenAI::Internal::Transport::BaseClient
...
+      webhook_secret: OpenAI::Internal::OMIT,
+      provider: nil,
+      base_url: OpenAI::Internal::OMIT,
       max_retries: self.class::DEFAULT_MAX_RETRIES,
       timeout: self.class::DEFAULT_TIMEOUT_IN_SECONDS,
       initial_retry_delay: self.class::DEFAULT_INITIAL_RETRY_DELAY,
       max_retry_delay: self.class::DEFAULT_MAX_RETRY_DELAY
     )
+      provider_runtime = nil
+      if provider
```
jliccini: From Codex: This truthiness check treats `provider: false` as though the argument were absent, bypassing the invalid-provider validation. Since the runtime domain is `Handle | nil`, could this use `unless provider.nil?` so every non-`nil` invalid value is rejected? With ambient OpenAI credentials configured, silently treating `false` as absent could also route through the normal OpenAI client unexpectedly.

lib/openai/providers/bedrock.rb
```diff
@@ -0,0 +1,480 @@
...
+        explicit_api_key = !api_key.equal?(OpenAI::Internal::OMIT) && !api_key.nil?
+        normalized_api_key = Bedrock.normalize_optional_string(api_key)
+        if explicit_api_key && normalized_api_key.nil?
+          raise OpenAI::Errors::Error, "The Bedrock bearer credential must not be empty."
+        end
+        if explicit_api_key && token_provider
+          raise OpenAI::Errors::Error,
+                "The `api_key` and `token_provider` options are mutually exclusive. Configure only one."
+        end
+        if token_provider && !token_provider.respond_to?(:call)
```
jliccini: From Codex: These protocol-object validations use truthiness, while the mode-selection code below uses non-`nil` presence (`!token_provider.nil?` / `!credentials_provider.nil?`). Consequently, `token_provider: false` skips validation but counts as an explicit bearer mode before falling through to AWS auth; `credentials_provider: false` similarly counts as an AWS mode but falls back to the default provider. Please test presence consistently with `nil?` and reject any non-`nil` value that does not implement the required protocol.

jliccini (commented):
Ruby truthiness and optional-value validation review.

lib/openai/providers/bedrock.rb
```diff
@@ -0,0 +1,480 @@
...
+          )
+
+          request.merge(
+            headers: OpenAI::Internal::Util.normalized_headers(headers, signature.headers),
+            follow_redirects: false,
+            provider_auth: SIGV4_AUTH_MARKER
+          )
+        rescue OpenAI::Errors::Error
+          raise
+        rescue StandardError => e
```
jliccini, line 171: From Codex: Because this rescue covers the entire method, it also converts internal programming errors from `request.fetch`, URI/header processing, or `request.merge` into credential-resolution errors. Could we narrow the rescue to credential resolution / `@signer.sign_request` (and preferably AWS-specific exception types), allowing malformed internal requests and SDK regressions to surface normally? The test that removes `:method` currently appears to codify a `KeyError` being misreported as an authentication failure.

jliccini (commented):
Ruby exception-boundary review.

lib/openai/client.rb
```diff
@@ -218,6 +227,10 @@ class Client < OpenAI::Internal::Transport::BaseClient
     #
     # @param webhook_secret [String, nil] Defaults to `ENV["OPENAI_WEBHOOK_SECRET"]`
     #
+    # @param provider [OpenAI::Internal::Provider::Handle, nil] Configure a supported
```
jliccini: From Codex: This is a public initializer parameter, and the RBI/RBS declarations also expose `OpenAI::Internal::Provider::Handle`. Once consumers type against that name, the `Internal` type effectively becomes part of the public compatibility surface. Would it be cleaner to expose an opaque public type such as `OpenAI::Provider`, while keeping the provider definition/runtime implementation internal?

lib/openai/providers/bedrock.rb
```diff
@@ -0,0 +1,480 @@
...
+        token_provider: nil,
+        access_key_id: nil,
+        secret_access_key: nil,
+        session_token: nil,
+        profile: nil,
+        credentials_provider: nil
+      )
+        normalized_region = Bedrock.normalize_optional_string(region)
+        if !region.nil? && normalized_region.nil?
+          raise OpenAI::Errors::Error, "The Bedrock AWS `region` must not be empty."
```
jliccini: From Codex: This and the other invalid Bedrock option combinations raise the generic `OpenAI::Errors::Error`, while adjacent `Client` argument validation uses `ArgumentError`. Since these represent caller configuration mistakes rather than API failures, could we use `ArgumentError` consistently, or introduce a dedicated configuration exception inheriting from `ArgumentError`? That gives callers more precise and idiomatic rescue semantics.

jliccini (commented):
Ruby public-type and exception-contract review.

lib/openai/client.rb
HAYDEN-OAI: yes, fixed in 35b96160. the opaque public type is now OpenAI::Provider, while the definition and runtime registry stay internal. i updated the yard, rbi, and rbs contracts, and the provider constructor remains private.

lib/openai/client.rb
HAYDEN-OAI: good call, fixed in 35b96160. client initialization now checks provider presence with nil? and rejects false or any foreign value before ambient openai configuration can be used. i added regression coverage for provider: false.

lib/openai/providers/bedrock.rb
HAYDEN-OAI: fixed in 35b96160, thank you! bedrock option validation and client provider conflicts now raise ArgumentError for caller configuration mistakes. credential resolution and request authentication failures remain OpenAI::Errors::Error, and the tests cover that boundary.

lib/openai/providers/bedrock.rb
HAYDEN-OAI: agreed, fixed in 35b96160. credential wrapping now lives in narrow resolve_token and sign_request helpers. request fetch, uri/header processing, and merge errors bubble unchanged, with regression coverage for malformed bearer and sigv4 requests.

lib/openai/providers/bedrock.rb
HAYDEN-OAI: fixed in 35b96160, thank you! both protocol options now use non-nil presence consistently, and false is rejected alongside other values that do not implement call or credentials. the validation matrix now covers both false cases.

PR #293, opened 2026-06-26
Preserve nil timeouts through request transport

fallintoplace:
## Summary
- preserve `timeout: nil` as "no timeout" instead of coercing it to `0.0`
- skip deadline and `x-stainless-timeout` handling when a request disables timeouts
- reset pooled `Net::HTTP` socket timeouts and add regressions for default and per-request nil timeout behavior

## Why
The README documents `timeout: nil` as disabling timeouts, but the transport currently calls `to_f` on the timeout value. In Ruby, `nil.to_f` becomes `0.0`, which turns the computed deadline into "right now" rather than removing the deadline.

## Validation
- `ASDF_RUBY_VERSION=3.4.9 asdf exec bundle exec ruby -Itest test/openai/client_test.rb --name=/timeout/`
- `ASDF_RUBY_VERSION=3.4.9 asdf exec bundle exec ruby -Itest test/openai/client_test.rb`

### DIFF
--- a/lib/openai/client.rb
+++ b/lib/openai/client.rb
@@ -223,7 +223,7 @@ module OpenAI
     #
     # @param max_retries [Integer] Max number of retries to attempt after a failed retryable request.
     #
-    # @param timeout [Float]
+    # @param timeout [Float, nil] Pass `nil` to disable request timeouts.
     #
     # @param initial_retry_delay [Float]
     #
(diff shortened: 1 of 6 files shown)


jbeckwith-oai:
Thank you for identifying and fixing the nil-timeout behavior. We are closing this PR while consolidating the timeout fix with the related HTTP response and diagnostics work in one replacement OpenAI Ruby PR. The implementation and contributor attribution will be preserved.

PR #294, opened 2026-06-26
Fix ReadIOAdapter max_len handling

fallintoplace:
## Summary
- fix `ReadIOAdapter#read_enum` so `read(max_len)` returns exactly `max_len` bytes instead of `max_len + 1`
- return `nil` when an enumerator-backed adapter is exhausted for bounded reads, matching `IO#read(max_len)`
- add a regression that reads across chunk boundaries and verifies exact-length slices through EOF

## Why
`ReadIOAdapter#read_enum` currently uses an inclusive slice when draining its buffer. That means a bounded read can return one extra byte, which is risky in the request-body path because the adapter feeds `Net::HTTP`.

While validating the fix, the same path also showed an EOF mismatch for bounded reads: after the final chunk, the adapter returned `""` once before returning `nil`. Tightening that behavior keeps the adapter aligned with `IO#read(max_len)` semantics.

## Validation
- `ASDF_RUBY_VERSION=3.4.9 asdf exec bundle exec ruby -Itest test/openai/internal/util_test.rb --name=/read/`
- `ASDF_RUBY_VERSION=3.4.9 asdf exec bundle exec ruby -Itest test/openai/internal/util_test.rb`

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -442,21 +442,23 @@ module OpenAI
         #
         # @return [String, nil]
         def read(max_len = nil, out_string = nil)
-          case @stream
-          in nil
-            nil
-          in IO | StringIO
-            @stream.read(max_len, out_string)
-          in Enumerator
-            read = read_enum(max_len)
-            case out_string
-            in String
-              out_string.replace(read)
+          read =
+            case @stream
             in nil
-              read
+              nil
+            in IO | StringIO
+              return @stream.read(max_len, out_string).tap(&@blk)
+            in Enumerator
+              read_enum(max_len)
             end
-          end
-            .tap(&@blk)
+
+          case out_string
+          in String
+            out_string.replace(read || "")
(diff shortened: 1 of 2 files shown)


jbeckwith-oai:
Thank you again, @fallintoplace, for isolating the `max_len + 1` read and bounded-EOF/output-buffer problems so clearly. The focused reproduction here made the contract failure easy to understand and gave us an important independent check on the broader adapter repair.

#354 now includes these corrections together with the complete byte-oriented buffering behavior from #275 and the mixed UTF-8/binary chunk fix for #317, all ported to the current SDK-owned adapter file on latest `main`. It also adds native length-conversion coverage and has passed the full Ruby 3.3/3.4/4.0 matrix, lint, signature validation, packaging, and CodeQL.

I’m closing this as superseded by #354. Your work here absolutely contributed to the final result, and I’m very grateful for the thoughtful diagnosis, tests, and follow-through. Thank you! 🙏

PR #311, opened 2026-07-16
Fix multiple image edit uploads

### LINKED ISSUE #201: Editing multiple images is broken
When using `client.images.edit(image:)` if you pass multiple files for the `image:` keyword argument you get a 400 error from the OpenAI API that says you must provide image in the request. The issue is that the multipart form formatting for arrays does not match the format that the OpenAI API expects.

## Code to reproduce

```ruby
client = OpenAI::Client.new


file_1 = OpenAI::FilePart.new(File.read("file1.png"), content_type: "image/png", filename: "file1.png")
file_2 = OpenAI::FilePart.new(File.read("file2.png"), content_type: "image/png", filename: "file2.png")

client.images.edit({ prompt: "do something", image: [file_1, file_2] })
```

## How to fix
There is a naive fix in this commit https://github.com/851-labs/openai-ruby/commit/9bba7349d8a0f3727eb495627f27eff413a0017b.

jbeckwith-oai:
## Summary

- encode multiple image edit uploads using the API's bracketed `image[]` multipart field
- preserve the existing scalar `image` field for single-image edits
- cover both non-streaming and streaming image edits with multipart regression tests

## Root cause

The generated multipart encoder flattened arrays of file inputs but reused the scalar field name for every part. The Images API therefore received duplicate `image` parameters instead of an `image[]` array and returned `duplicate_parameter`.

This applies the array field name only to image edit requests, avoiding a behavior change for unrelated multipart array parameters.

Fixes #201.

## Validation

- `ruby -Itest test/openai/resources/images_test.rb --name '/test_edit_(encodes_multiple_images_as_an_array_multipart_field|preserves_the_scalar_image_multipart_field)/'`
- `ruby -Itest test/openai/internal/util_test.rb`
- `rubocop --force-exclusion lib/openai/resources/images.rb test/openai/resources/images_test.rb`
- live API contract check: the request now reaches image validation (`invalid_image_file`) instead of failing with `duplicate_parameter`

### DIFF
--- a/lib/openai/resources/images.rb
+++ b/lib/openai/resources/images.rb
@@ -87,6 +87,7 @@ module OpenAI
       # @see OpenAI::Models::ImageEditParams
       def edit(params)
         parsed, options = OpenAI::ImageEditParams.dump_request(params)
+        parsed = prepare_edit_body(parsed)
         if parsed[:stream]
           message = "Please use `#edit_stream_raw` for the streaming use case."
           raise ArgumentError.new(message)
@@ -148,6 +149,7 @@ module OpenAI
       # @see OpenAI::Models::ImageEditParams
       def edit_stream_raw(params)
         parsed, options = OpenAI::ImageEditParams.dump_request(params)
+        parsed = prepare_edit_body(parsed)
         unless parsed.fetch(:stream, true)
           message = "Please use `#edit` for the non-streaming use case."
           raise ArgumentError.new(message)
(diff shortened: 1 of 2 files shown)


jbeckwith-oai:
This targeted fix was useful for validating the `image[]` API contract and, importantly, the requirement that a single image remain the scalar `image` field.

That behavior is now included in the shared multipart serializer change in #348, alongside primitive arrays and recursive nested objects. I’m closing this narrower PR in favor of the consolidated implementation so we only carry one fix through review.

PR #322, opened 2026-07-31
feat: require Ruby 3.3 or newer

jbeckwith-oai:
## Summary

- require Ruby 3.3 or newer and remove Ruby 3.2 from the supported CI matrix
- add a repository versioning policy covering breaking changes, Ruby support, dependencies, and release maintenance
- retain Ruby 4.0.6 as the contributor default and Ruby 4.0 for routine lint, package, and release jobs
- include the versioning policy in the built gem documentation

## Why

Ruby 3.2 reached upstream end of life on April 1, 2026. The SDK should support maintained CRuby releases by default, with post-EOL grace requiring an explicit exception. The published v0.75.x line remains available to Ruby 3.2 users without a separate maintenance promise.

Because the gem is pre-1.0, this runtime-floor change is a minor release. The `feat:` classification allows release-please to produce v0.76.0 under the repository's existing configuration.

## Impact

Ruby 3.2 users remain on v0.75.x. New SDK releases require Ruby 3.3 or newer. Ruby 3.3, 3.4, and 4.0 remain blocking CI targets.

## Validation

- `./scripts/test` on Ruby 3.3.12, 3.4.10, and 4.0.6: 468 tests / 1,514 assertions on each runtime
- `./scripts/lint` on Ruby 4.0.6
- `bundle exec rake build:gem` on Ruby 4.0.6
- verified the built gem contains `README.md` and `VERSIONING.md`
- verified Ruby 3.2.11 rejects the built gem and Ruby 3.3.12 installs it
- verified published v0.75.0 still installs on Ruby 3.2.11

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,77 @@
+# Versioning policy
+
+The OpenAI Ruby SDK follows [Semantic Versioning](https://semver.org/). This
+policy covers the public gem API, supported Ruby runtimes, dependencies, and
+type definitions.
+
+## Gem releases
+
+| Change | While the gem is `0.x` | After `1.0` |
+| --- | --- | --- |
+| Backwards-compatible bug or security fix | Patch | Patch |
+| Additive endpoint, optional argument, response field, or public type | Minor | Minor |
+| Higher minimum Ruby version | Minor | Major by default |
+| Other user-visible breaking change | Minor, with migration guidance | Major |
+
+Breaking changes include removing or renaming public behavior and incompatible
+changes to required arguments, return types, exception families,
+serialization, retries, or authentication. A dependency upgrade follows the
+same rule: its version number does not determine the SDK release; its effect on
+SDK users does.
+
+Corrections to non-runtime RBI or RBS definitions are normally patches. Use a
+minor release when a type-only change is likely to create substantial new type
+errors for users.
+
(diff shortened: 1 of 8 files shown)


VERSIONING.md
```diff
@@ -0,0 +1,77 @@
...
+branches](https://www.ruby-lang.org/en/downloads/branches/) for the upstream
+lifecycle.
+
+The current supported versions are Ruby 3.3, 3.4, and 4.0. The authoritative
+minimum is `required_ruby_version` in `openai.gemspec`.
+
+When the minimum Ruby version changes:
+
+1. Ship the change in the release category defined above, never in a patch.
+2. Name the final compatible gem release line in the README and release notes.
```
HAYDEN-OAI, line 43: [P3] Include the final Ruby 3.2-compatible line in generated release notes. This policy requires both the README and release notes to name the final compatible line, but the only release-please input is `feat: require Ruby 3.3 or newer`, so the generated v0.76.0 changelog/release notes will omit `v0.75.x`. Include that final compatible line in the releasable commit/PR title or explicitly add it to the release notes so users on Ruby 3.2 know which version to pin.

HAYDEN-OAI (approved):
Approved: Ruby 3.3 minimum, supported-runtime coverage, versioning policy, and gem packaging are consistent. One non-blocking release-note follow-up is noted inline.

PR #324, opened 2026-08-03
fix: preserve binary multipart chunks

### LINKED ISSUE #317: Encoding::CompatibilityError during multipart file upload is swallowed and reported as a generic APIConnectionError ("Connection error.")
Uploading a file (e.g. via `client.audio.transcriptions.create`) alongside any other multipart text field that contains a non-ASCII UTF-8 character can raise `Encoding::CompatibilityError` while the gem is building the multipart request body. This gets caught by a blanket rescue in the transport layer and re-raised as a generic `OpenAI::Errors::APIConnectionError` with the message "Connection error.", indistinguishable from an actual network failure.

We hit this in production doing audio transcription with a prompt field built from lesson text containing a curly apostrophe (', U+2019)

This happened when trying to use code similar to this:

```
require "openai"

client = OpenAI::Client.new(api_key: ENV["OPENAI_API_KEY"])

client.audio.transcriptions.create(
  model: "gpt-4o-transcribe",
  file: File.open("path/to/audio_file.mp3", "rb"),
  prompt: "Transcribe the users’s audio verbatim." # U+2019, plain curly apostrophe
)
# => OpenAI::Errors::APIConnectionError: Connection error.
```

### How to reproduce

```
require "openai"

utf8_chunk = "café".dup.force_encoding("UTF-8")   # any non-ASCII UTF-8 text field
binary_chunk = [0xFF, 0xFE, 0x00, 0x01].pack("C*") # any non-ASCII binary file bytes

enum = Enumerator.new do |y|
  y << utf8_chunk
  y << binary_chunk
end

adapter = OpenAI::Internal::Util::ReadIOAdapter.new(enum) {}

loop do
  chunk = adapter.read(2)
  break if chunk.nil?
end

# => 'OpenAI::Internal::Util::ReadIOAdapter#read_enum': incompatible character encodings: UTF-8 and BINARY (ASCII-8BIT) (Encoding::CompatibilityError)
```

### Suggested fix

Force every chunk to binary encoding before appending in `read_enum`, since the buffer is only ever used to produce raw bytes for the wire (nothing is lost), since HTTP bodies are transmitted as bytes regardless of Ruby's internal encoding tag:

```ruby
# lib/openai/internal/util.rb

class ReadIOAdapter

  private def read_enum(max_len)
    case max_len
    in nil
      @stream.to_a.map(&:b).join
    in Integer
      @buf << @stream.next.b while @buf.length < max_len
      @buf.slice!(..max_len)
    end
  rescue StopIteration
    @stream = nil
    @buf.slice!(0..)
  end
  
  def initialize(src, &blk)
    ...
    @buf = String.new.b
    @blk = blk
  end
```

Guflly:
## Summary

- normalize multipart enumerator chunks to binary before buffering
- keep bounded and whole-stream reads in binary encoding
- add regression tests for mixed UTF-8 and binary chunks

Fixes #317.

## Validation

- focused adapter tests pass
- RuboCop passes
- the full util test file has one existing Windows CGI filename failure, reproduced on current `main`

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -425,9 +425,9 @@ module OpenAI
         private def read_enum(max_len)
           case max_len
           in nil
-            @stream.to_a.join
+            @stream.to_a.map(&:b).join
           in Integer
-            @buf << @stream.next while @buf.length < max_len
+            @buf << @stream.next.b while @buf.length < max_len
             @buf.slice!(..max_len)
           end
         rescue StopIteration
@@ -476,7 +476,7 @@ module OpenAI
             else
               src
             end
-          @buf = String.new
+          @buf = String.new.b
           @blk = blk
         end
       end
--- a/test/openai/internal/util_test.rb
+++ b/test/openai/internal/util_test.rb
@@ -330,6 +330,37 @@ class OpenAI::Test::UtilIOAdapterTest < Minitest::Test
     end
   end
 
+  def test_read_mixed_encoding_chunks
+    chunks = ["caf\u00E9", "\xFF\xFE".b]
+    expected = chunks.map(&:b).join
+
+    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(chunks.to_enum) do |chunk|
+      chunk
+    end
+    actual = String.new.b
+    loop do
+      chunk = adapter.read(2)
+      break unless chunk
+
+      actual << chunk
+    end
+
+    assert_equal(expected, actual)
+    assert_equal(Encoding::ASCII_8BIT, actual.encoding)
+  end
+
+  def test_read_all_mixed_encoding_chunks
+    chunks = ["caf\u00E9", "\xFF\xFE".b]
+    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(chunks.to_enum) do |chunk|
+      chunk
+    end
+
+    result = adapter.read
+
+    assert_equal(chunks.map(&:b).join, result)
+    assert_equal(Encoding::ASCII_8BIT, result.encoding)
+  end
+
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the adapter and all production call sites. Converting enumerated chunks to ASCII-8BIT before buffering preserves the exact bytes, eliminates UTF-8/BINARY compatibility failures, and leaves IO/StringIO behavior and transport framing unchanged. Bounded reads remain streaming with memory limited to the current source chunk/buffer; the unbounded-read path retains its existing whole-stream semantics. Regression coverage exercises both bounded and whole-stream mixed-encoding reads. Local validation: util suite 38/38 (160 assertions), focused mixed-encoding tests, an end-to-end multipart text+binary byte-integrity probe, RuboCop on both changed files, and git diff --check all pass.

jbeckwith-oai:
Thank you very much for tracking this down and putting together the focused mixed-encoding fix, @Guflly. Normalizing every multipart Enumerator chunk to binary was the essential missing piece for #317, and your reproduction made the UTF-8-followed-by-binary failure concrete and straightforward to verify.

We carried that exact insight into #354 and combined it with the complete `ReadIOAdapter` contract repair: byte-limited reads, buffered and unbounded reads, EOF, negative lengths, output buffers, and native `IO` delegation. The canonical version is rebased onto current `main` in the SDK-owned adapter file and is now fully green across the Ruby matrix, lint, signatures, packaging, and CodeQL, with a real 1 MiB multipart upload verified byte-for-byte.

I’m closing this PR because #354 supersedes it. Your contribution was crucial to the final fix, and we genuinely appreciate the clear report, targeted implementation, and regression coverage. Thank you! 🙏

PR #332, opened 2026-08-03
ci: prepare checks for merge queue

jbeckwith-oai:
## Summary

- trigger the required CI workflow for GitHub merge-group checks
- allow the aggregate `ci / ci-required` job to run for queued merge commits

## Why

GitHub requires Actions-based required checks to listen for the separate `merge_group` event. Without this change, enabling the queue would leave queued pull requests waiting for a required check that never reports.

## Impact

Once the repository merge queue is enabled, CI will run again when an approved pull request enters the queue, validating it against the latest `main` and any changes ahead of it in the queue. The Stainless staging artifact path remains unchanged.

## Validation

- YAML parsed successfully
- `git diff --check`
- trigger/job-gate alignment check
- thermo-nuclear code-quality review (no findings)

## Follow-up

After this lands, enable the `main` ruleset merge queue with squash-only merging, all-green grouping, one PR per merge, five concurrent builds, and a 60-minute check timeout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ on:
     branches-ignore:
       - 'stl-preview-head/**'
       - 'stl-preview-base/**'
+  merge_group:
+    types:
+      - checks_requested
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -33,6 +36,7 @@ jobs:
         (github.event_name == 'push' || github.event.pull_request.head.repo.fork)) ||
       (!startsWith(github.repository, 'stainless-sdks/') &&
         (github.event_name == 'pull_request' ||
+          github.event_name == 'merge_group' ||
           github.ref == 'refs/heads/main' ||
           github.ref == 'refs/heads/next'))
     permissions:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - '**'
+      - '!gh-readonly-queue/**'
       - '!integrated/**'
       - '!stl-preview-head/**'
       - '!stl-preview-base/**'


HAYDEN-OAI (commented):
One substantive merge-queue correctness issue identified: queue-branch push runs can cancel the only workflow run that reports the required merge-group check. Details and remediation are attached inline. The required-check aggregation, current `main` protections, and existing CodeQL PR/post-merge behavior otherwise match the stated rollout.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,9 @@ on:
     branches-ignore:
       - 'stl-preview-head/**'
       - 'stl-preview-base/**'
+  merge_group:
```
HAYDEN-OAI, line 22: [P1] Prevent queue-branch pushes from canceling merge-group CI

Creating a `gh-readonly-queue/main/...` branch emits both a `push` event and `merge_group.checks_requested` ([GitHub confirms the queue-branch push](https://github.blog/changelog/2023-07-12-pull-request-merge-queue-is-now-generally-available/)). The existing `push.branches` includes `**`, and both events resolve `github.event.pull_request.number || github.ref` to the same queue ref. Because `cancel-in-progress: true`, a `push` run that starts second cancels the `merge_group` run; on `openai/openai-ruby` that push run's `checks` job is skipped because its ref is neither `main` nor `next`, so no `ci / ci-required` check remains and the queue waits until its 60-minute timeout. Exclude `gh-readonly-queue/**` from `push.branches` ([as GitHub's own workflow does](https://github.com/github/docs/blob/32e13701d3aa477c4b56ebd014c8b2aca88f07a5/.github/workflows/moda-ci.yaml#L8-L12)), or include `github.event_name` in the concurrency group.

jbeckwith-oai: Fixed in `c4a1f321`. I excluded `gh-readonly-queue/**` from the existing `push.branches` filter, so GitHub's temporary queue refs now start only the authoritative `merge_group` workflow. That removes the skipped push run and prevents it from canceling `ci / ci-required` through the shared concurrency key. I also revalidated YAML parsing, `git diff --check`, and the queue-trigger/exclusion alignment; the required thermo-nuclear review had no findings.

HAYDEN-OAI (approved):
Approved at `c4a1f3215fe675b6626d69a194fa5c888e8b4f51`. The new `!gh-readonly-queue/**` exclusion follows the positive `**` branch pattern, so queue-ref pushes no longer create a competing run; `merge_group: checks_requested` remains the authoritative queue run and still produces the required `ci / ci-required` context. The existing Code Owner review, strict required check, linear-history/deletion/force-push protections, squash-only repository setting, bypass actors, and CodeQL PR/post-merge behavior are preserved.

The remaining Copilot thread is a false positive: GitHub Actions documents that dereferencing an absent context property evaluates to an empty string, not an error ([contexts reference](https://docs.github.com/en/enterprise-cloud%40latest/actions/reference/workflows-and-actions/contexts#available-contexts)); additionally, the `openai/openai-ruby` repository predicates guard the Stainless-only property accesses. No substantive issues remain.

.github/workflows/ci.yml
```diff
@@ -33,6 +37,7 @@ jobs:
         (github.event_name == 'push' || github.event.pull_request.head.repo.fork)) ||
       (!startsWith(github.repository, 'stainless-sdks/') &&
         (github.event_name == 'pull_request' ||
+          github.event_name == 'merge_group' ||
           github.ref == 'refs/heads/main' ||
```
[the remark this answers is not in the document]
jbeckwith-oai: No code change is needed here. On `merge_group`, an absent `github.event.pull_request` property evaluates to an empty string rather than causing expression evaluation to fail. For this repository, the Stainless-only side of both expressions is also gated by `startsWith(github.repository, 'stainless-sdks/')` / the exact Stainless repository comparison, while the OpenAI job gate explicitly accepts `github.event_name == 'merge_group'`. The approving review independently verified this behavior at `c4a1f321`; the merged CI run was fully green.

PR #333, opened 2026-08-03
ci: speed up RBS validation without Steep internals

jbeckwith-oai:
## Summary

This changes the RBS half of CI from a generic “typecheck” job into an explicit `validate (rbs)` job and replaces the slow per-file Steep invocation with a public-CLI validation pipeline.

The validator now:

1. collects the repository’s 1,209 generated `.rbs` files;
2. runs `rbs parse sig`, which parses every original file independently;
3. rejects parser-boundary hazards and file-scoped semantics before combining independently parseable signatures into one temporary signature file;
4. asks the locked `steep check` CLI to perform semantic validation once over the merged declaration environment;
5. reruns ordinary per-file `steep check` against the original files whenever either fast-path stage fails, preserving exact source paths and GitHub annotations.

The RBS job remains an explicit dependency of `ci-required`, so it cannot be skipped while the aggregate required check passes.

## Root cause

The generated signatures are split across 1,209 files and heavily reopen the same namespaces. Steep validates the merged definition of a reopened namespace in the context of each input signature file. That makes the work grow with both the number of signature files and the size of the merged namespaces, even though there are no Ruby source files configured for Steep to typecheck.

This is the same split-signature performance shape discussed in [steep#661](https://github.com/soutaro/steep/issues/661). Consolidating signatures removes the repeated per-file namespace validation while keeping Steep itself as the semantic validator.

On the current repository:

| Path | Wall time |
| --- | ---: |
| Ordinary per-file `steep check --no-type-check --jobs=8` | 52.38s |
| Independent `rbs parse sig` preflight | 0.40s |
| Complete optimized validator, including preflight | 6.22s |

The reference and optimized commands both pass the full 1,209-file corpus. The optimized path remains roughly 8× faster locally. Before this change, the GitHub job spent roughly 3m31s in RBS validation, so this targets the actual long pole rather than Ruby setup or Sorbet.

### Public CLIs are the only tool boundaries

The implementation does not load Steep or RBS as Ruby libraries and does not reference `Steep::` or `RBS::` implementation constants.

It invokes two documented commands from the locked bundle:

- `rbs parse sig` for independent syntax parsing only;
- `steep check` for the existing semantic validation.

The RBS parser preflight is not a replacement for Steep and does not make semantic acceptance decisions. Steep remains the authority for type names, inheritance, aliases, generic bounds, constants, members, and other semantic validity.

### Original file boundaries remain authoritative for syntax

Concatenation cannot be allowed to decide whether the original files are syntactically valid. Without a preflight, two independently invalid files can repair each other. For example:

- `a.rbs`: `class Example`
- `b.rbs`: `end`

Each original file is invalid, but their concatenation is a valid class declaration. Split return types and orphan annotations can create the same class of false green.

The validator therefore runs `rbs parse sig` before creating the combined buffer. The CLI parses every original file independently. A syntax failure skips consolidation and routes directly to ordinary per-file Steep, which reports the exact original file and line. A regression test uses the split `class`/`end` reproduction and requires the wrapper to fail with an original-file `RBS::SyntaxError`.

Independent parsing alone is not sufficient for every parser-boundary byte. Locked RBS 3.9.5 treats an embedded NUL (`\\x00`) as EOF while still exiting successfully. If such a file were concatenated before another signature, the combined parser could silently ignore every later declaration. The eligibility scan therefore reads signatures as binary and routes the whole job through ordinary per-file Steep if any NUL is present. The regression test places a NUL in the first file and an unknown type in the second; validation must fail against the second original file.

### The semantic fast path changes partitioning, not declarations

After every source file has independently parsed, the script writes each file byte-for-byte, in deterministic glob order, into one temporary `sig/all.rbs`, adding only newline separators. The repository’s existing `Steepfile` is copied into the temporary project, so the same target, libraries, diagnostic configuration, and dependency setup are used.

RBS declarations live in a merged environment. For independently parseable declaration files, combining buffers changes how many parser buffers contain those declarations, but not the classes, modules, aliases, constants, methods, type parameters, bounds, or annotations that Steep validates.

### File-scoped semantics are explicitly excluded

Locked RBS 3.9.5 has two constructs that intentionally alter name resolution on a per-file basis:

- `use ...`
- `# resolve-type-names: ...`

Concatenating files containing either construct could change which declarations the directive affects. The validator matches the actual `use` keyword boundary (`use\\b`), including whitespace-free forms such as `use::Example::Foo`, wildcard imports, and comment adjacency. It applies the directive pattern to each complete binary file, rather than physical lines, because locked RBS 3.9.5's magic-comment grammar allows `\\s*` to span newlines (for example, `#\\nresolve-type-names: false`). It scans for both file-scoped forms, alongside the NUL/EOF boundary described above, and bypasses consolidation entirely if any hazard is present anywhere in `sig/`. In that case it immediately runs ordinary per-file Steep.

The current generated corpus contains neither construct. If generation introduces one later, correctness wins automatically and CI becomes slower rather than weaker.

### Any fast-path rejection is checked against the original files

A nonzero result from either independent parsing or consolidated Steep is never reported directly as the final answer. The script reruns Steep over the untouched repository signatures and uses that reference invocation’s exit status.

That has three important properties:

- syntax and semantic errors are reported against their original file and line;
- CI still emits normal GitHub annotations;
- a problem caused only by the optimization cannot create a false CI failure—if ordinary Steep passes, the validation task passes after noting that the reference path was used.

The failure path intentionally retains the old cost. Slow red builds are preferable to fast but ambiguous diagnostics. The workflow preserves `STEEP_JOBS: 8` specifically so this reference fallback produces diagnostics promptly.

### The validation level is not weakened

The command uses `--no-type-check` to state the job’s actual scope: RBS validation. The current `Steepfile` declares `signature("sig")` and no Ruby `check` paths, so the old job was not typechecking Ruby source code. It was already doing signature validation only.

Steep’s normal signature validation remains enabled. Tests demonstrate that the wrapper rejects:

- independently invalid files that become valid when concatenated;
- embedded NUL bytes that would hide later declarations in a combined buffer;
- whitespace-free or comment-adjacent `use` directives whose imports would leak across file boundaries;
- multiline `resolve-type-names` magic directives whose setting would leak into later files;
- ordinary RBS syntax errors;
- unknown type names;
- unsatisfied generic type-parameter bounds;
- other semantic errors surfaced by Steep.

Sorbet’s RBI check remains a separate `typecheck (rbi)` job.

### Operational failures fail closed

Missing signatures, a missing `Steepfile`, unavailable `rbs` or `steep` executables, temporary-file errors, and unexpected exceptions all produce a nonzero exit. The optimization does not turn infrastructure failures into successful validation.

### The required-check contract is explicit

The reusable workflow has separate job IDs for `typecheck` and `validate-rbs`. The `required` job:

- lists `validate-rbs` in `needs`;
- captures `needs.validate-rbs.result`;
- explicitly requires that result to equal `success`.

Branch protection can therefore continue requiring the single aggregate `ci-required` context without losing the RBS gate.

## Intentional tradeoffs

- Green builds pay approximately 0.4s to preserve independent syntax boundaries before taking the consolidated semantic path.
- Red builds rerun the old per-file path to obtain exact diagnostics, so failures remain comparatively slow.
- Repositories that introduce file-scoped RBS directives or NUL-containing signatures lose the optimization until a semantics-preserving strategy is added.
- This job is deliberately named validation, not typechecking, because no Ruby source paths are configured in Steep.

There is no dependency on private Steep/RBS APIs, no new language toolchain, and no reduction in semantic validation.

## Verification

- `ruby test/scripts/validate_rbs_test.rb`: 12 runs, 60 assertions
- Same tests with ambient `CI=1`: 12 runs, 60 assertions
- Full optimized validation: 1,209 RBS files, no errors
- Full reference per-file Steep validation: no errors
- `rbs parse sig`: 1,209 original files parsed independently
- RuboCop: 2,589 files, no offenses
- Ruby syntax check: passes
- Workflow YAML parse: passes
- `git diff --check`: passes

Tests cover the cross-file syntax-repair regression, the NUL/EOF truncation regression, whitespace-free `use` directives, the multiline magic-comment regression, consolidated success, reference fallback, both file-scoped directive families, reopened namespaces, ordinary syntax errors, semantic generic-bound errors, unknown names, fallback worker tuning, and GitHub annotation formatting.

The full local integration harness requires Ruby 3.3 or newer; this machine has Ruby 3.2. PR CI runs all supported Ruby versions and is the authoritative full-suite validation.

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -76,21 +76,10 @@ jobs:
 
   typecheck:
     timeout-minutes: 10
-    name: typecheck (${{ matrix.format }})
+    name: typecheck (rbi)
     permissions:
       contents: read
     runs-on: ${{ inputs.runner }}
-    env:
-      # Steep defaults to two workers in CI; eight is the measured balance for this runner.
-      STEEP_JOBS: 8
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - format: rbi
-            task: typecheck:sorbet
-          - format: rbs
-            task: typecheck:steep
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
@@ -100,8 +89,26 @@ jobs:
         with:
           ruby-version: '4.0'
           bundler-cache: true
-      - name: Typecheck ${{ matrix.format }} files
-        run: bundle exec rake ${{ matrix.task }}
+      - name: Typecheck rbi files
+        run: bundle exec rake typecheck:sorbet
+
+  validate-rbs:
+    timeout-minutes: 10
+    name: validate (rbs)
+    permissions:
+      contents: read
+    runs-on: ${{ inputs.runner }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@95ef2b042f9d7a56d8268cba8559e2842e2ad01b # v1
+        with:
+          ruby-version: '4.0'
+          bundler-cache: true
+      - name: Validate rbs files
+        run: bundle exec rake validate:rbs
 
   package:
     timeout-minutes: 10
@@ -151,6 +158,7 @@ jobs:
       - stainless-artifact
       - lint
       - typecheck
+      - validate-rbs
       - package
       - test-ruby
     if: ${{ always() }}
@@ -162,11 +170,13 @@ jobs:
           ARTIFACT_RESULT: ${{ needs.stainless-artifact.result }}
           LINT_RESULT: ${{ needs.lint.result }}
           TYPECHECK_RESULT: ${{ needs.typecheck.result }}
+          RBS_VALIDATION_RESULT: ${{ needs.validate-rbs.result }}
           PACKAGE_RESULT: ${{ needs.package.result }}
           TEST_RESULT: ${{ needs.test-ruby.result }}
         run: |
           test "$ARTIFACT_RESULT" = "success" || test "$ARTIFACT_RESULT" = "skipped"
           test "$LINT_RESULT" = "success"
           test "$TYPECHECK_RESULT" = "success"
+          test "$RBS_VALIDATION_RESULT" = "success"
           test "$PACKAGE_RESULT" = "success"
           test "$TEST_RESULT" = "success"
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -98,6 +98,10 @@ jobs:
     permissions:
       contents: read
     runs-on: ${{ inputs.runner }}
+    env:
+      # The consolidated green path always uses one worker; keep the reference
+      # fallback tuned so failing builds produce exact diagnostics promptly.
+      STEEP_JOBS: 8
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
--- a/scripts/validate-rbs
+++ b/scripts/validate-rbs
@@ -25,14 +25,15 @@ module OpenAI
         if signatures.any? { contains_file_scoped_directive?(_1) }
           run_reference_check(root, stdout: stdout, stderr: stderr, env: env)
         else
-          consolidated_status = run_consolidated_check(root, steepfile, signatures, env: env)
-          unless consolidated_status.success?
+          fast_status = run_parse_check(root, env: env)
+          fast_status = run_consolidated_check(root, steepfile, signatures, env: env) if fast_status.success?
+          unless fast_status.success?
             stdout.puts(
-              "Consolidated RBS validation failed; checking original signature files for exact diagnostics."
+              "Fast RBS validation failed; checking original signature files for exact diagnostics."
             )
-            consolidated_status = run_reference_check(root, stdout: stdout, stderr: stderr, env: env)
+            fast_status = run_reference_check(root, stdout: stdout, stderr: stderr, env: env)
           end
-          consolidated_status
+          fast_status
         end
 
       if status.success?
@@ -62,6 +63,12 @@ module OpenAI
       File.foreach(path).any? { _1.match?(FILE_SCOPED_DIRECTIVE) }
     end
     private_class_method :contains_file_scoped_directive?
+
+    def run_parse_check(root, env:)
+      _stdout, _stderr, status = Open3.capture3(env, "rbs", "parse", "sig", chdir: root.to_s)
+      status
+    end
+    private_class_method :run_parse_check
 
     def run_consolidated_check(root, steepfile, signatures, env:)
       Dir.mktmpdir("openai-rbs-validation") do |directory|
--- a/scripts/validate-rbs
+++ b/scripts/validate-rbs
@@ -1,0 +1,111 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "fileutils"
+require "open3"
+require "pathname"
+require "tmpdir"
+
+module OpenAI
+  module RBSValidation
+    # These directives change name resolution on a per-file basis, so signatures
+    # containing them cannot be safely combined and must take the reference path.
+    FILE_SCOPED_DIRECTIVE = /^\s*(?:use(?:\s|$)|#\s*resolve-type-names\s*:)/
+
+    module_function
+
+    def run(root: Pathname.pwd, stdout: $stdout, stderr: $stderr, env: ENV.to_h)
+      steepfile = root.join("Steepfile")
+      raise Errno::ENOENT, steepfile unless steepfile.file?
+
+      signatures = signature_paths(root)
+      raise "No RBS files found under #{root.join('sig')}" if signatures.empty?
+
+      status =
+        if signatures.any? { contains_file_scoped_directive?(_1) }
+          run_reference_check(root, stdout: stdout, stderr: stderr, env: env)
+        else
+          consolidated_status = run_consolidated_check(root, steepfile, signatures, env: env)
+          unless consolidated_status.success?
+            stdout.puts(
+              "Consolidated RBS validation failed; checking original signature files for exact diagnostics."
+            )
+            consolidated_status = run_reference_check(root, stdout: stdout, stderr: stderr, env: env)
+          end
+          consolidated_status
+        end
+
+      if status.success?
+        noun = signatures.one? ? "file" : "files"
+        stdout.puts("Validated #{signatures.size} RBS #{noun} with no errors.")
+        0
+      else
+        1
+      end
+    rescue StandardError => e
+      message =
+        if e.respond_to?(:detailed_message)
+          e.detailed_message(highlight: false)
+        else
+          "#{e.class}: #{e.message}"
+        end
+      stderr.puts(message)
+      1
+    end
+
+    def signature_paths(root)
+      Dir.glob(root.join("sig/**/*.rbs")).map { Pathname(_1) }
+    end
+    private_class_method :signature_paths
+
+    def contains_file_scoped_directive?(path)
+      File.foreach(path).any? { _1.match?(FILE_SCOPED_DIRECTIVE) }
+    end
+    private_class_method :contains_file_scoped_directive?
+
+    def run_consolidated_check(root, steepfile, signatures, env:)
+      Dir.mktmpdir("openai-rbs-validation") do |directory|
+        temporary_root = Pathname(directory)
+        temporary_sig = temporary_root.join("sig")
+        temporary_sig.mkpath
+        temporary_steepfile = temporary_root.join("Steepfile")
+        FileUtils.cp(steepfile, temporary_steepfile)
+        # Steep validates reopened namespaces once per input file. Combining the
+        # generated declarations avoids that repeated work; Steep still performs
+        # all parsing and semantic validation through its public CLI.
+        concatenate(signatures, temporary_sig.join("all.rbs"))
+
+        command = [env.fetch("STEEP_COMMAND", "steep"), "check", "--no-type-check", "--jobs=1"]
+        command << "--steepfile=#{temporary_steepfile}"
+        _stdout, _stderr, status = Open3.capture3(env, *command, chdir: root.to_s)
+        status
+      end
+    end
+    private_class_method :run_consolidated_check
+
+    def concatenate(signatures, destination)
+      destination.open("wb") do |combined|
+        signatures.each do |path|
+          contents = path.binread
+          combined.write(contents)
+          combined.write("\n") unless contents.end_with?("\n")
+          combined.write("\n")
+        end
+      end
+    end
+    private_class_method :concatenate
+
+    def run_reference_check(root, stdout:, stderr:, env:)
+      command = [env.fetch("STEEP_COMMAND", "steep"), "check", "--no-type-check"]
+      command << "--jobs=#{env.fetch('STEEP_JOBS')}" if env.key?("STEEP_JOBS")
+      command << "--format=github" if env.key?("CI")
+      output, errors, status = Open3.capture3(env, *command, chdir: root.to_s)
+      stdout.print(output)
+      stderr.print(errors)
+      status
+    end
+    private_class_method :run_reference_check
+  end
+end
+
+exit(OpenAI::RBSValidation.run) if $PROGRAM_NAME == __FILE__
--- a/scripts/validate-rbs
+++ b/scripts/validate-rbs
@@ -10,7 +10,7 @@ module OpenAI
   module RBSValidation
     # These directives change name resolution on a per-file basis, so signatures
     # containing them cannot be safely combined and must take the reference path.
-    FILE_SCOPED_DIRECTIVE = /^\s*(?:use(?:\s|$)|#\s*resolve-type-names\s*:)/
+    FILE_SCOPED_DIRECTIVE = /^\s*(?:use\b|#\s*resolve-type-names\s*:)/
 
     module_function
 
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-checks.yml
```diff
@@ -76,21 +76,28 @@ jobs:
...
-      STEEP_JOBS: 8
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - format: rbi
-            task: typecheck:sorbet
-          - format: rbs
-            task: typecheck:steep
     steps:
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 95b56084. I restored `STEEP_JOBS: 8` on the dedicated `validate-rbs` job. The consolidated green path still explicitly uses `--jobs=1`; only the ordinary per-file reference fallback receives the tuned worker count. I also updated the fallback test to assert `--jobs=8` is forwarded alongside the GitHub formatter.

.github/workflows/ci-checks.yml
```diff
@@ -76,21 +76,28 @@ jobs:
...
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@95ef2b042f9d7a56d8268cba8559e2842e2ad01b # v1
+        with:
+          ruby-version: '4.0'
+          bundler-cache: true
+      - name: Typecheck rbi files
+        run: bundle exec rake typecheck:sorbet
+
+  validate-rbs:
+    timeout-minutes: 10
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 95b56084. The `validate-rbs` job now sets `STEEP_JOBS: 8`, preserving the previous tuning for slow reference checks while the consolidated fast path remains pinned to one worker. The test now verifies the fallback command includes both `--jobs=8` and `--format=github`.

scripts/validate-rbs
```diff
@@ -0,0 +1,111 @@
...
+      Dir.mktmpdir("openai-rbs-validation") do |directory|
+        temporary_root = Pathname(directory)
+        temporary_sig = temporary_root.join("sig")
+        temporary_sig.mkpath
+        temporary_steepfile = temporary_root.join("Steepfile")
+        FileUtils.cp(steepfile, temporary_steepfile)
+        # Steep validates reopened namespaces once per input file. Combining the
+        # generated declarations avoids that repeated work; Steep still performs
+        # all parsing and semantic validation through its public CLI.
+        concatenate(signatures, temporary_sig.join("all.rbs"))
```
HAYDEN-OAI, line 88: **[P1] Validate each original file before accepting consolidated success**

With locked RBS 3.9.5 / Steep 1.10.0, create `sig/a.rbs` containing `class Example\n` and `sig/b.rbs` containing `end\n`. Ordinary `steep check --no-type-check --jobs=1 --format=github` exits 1 and annotates syntax errors in both originals, but this concatenation produces valid `class Example\n\nend\n\n`, so `scripts/validate-rbs` exits 0 and prints `Validated 2 RBS files with no errors.` Since fallback only happens when the combined check fails, malformed generated signatures can silently satisfy `ci-required`. Split return types and orphan `%a{...}` annotations produce the same false negative. Parse the original files independently before merging (the documented `rbs parse sig` processes all 1,209 current files in ~0.64s) and route parse failures through the existing ordinary-Steep fallback.

HAYDEN-OAI (commented):
I found one P1 correctness issue in the consolidation safety boundary: independently invalid RBS files can form a valid combined parser buffer, allowing the wrapper to report success without invoking its exact-diagnostics fallback. The inline comment includes an exact reproduction on locked RBS 3.9.5 / Steep 1.10.0 and a measured public-CLI mitigation. I also verified the file-scoped directive grammar, restored fallback worker count, required-check dependency, and actual CI speedup.

scripts/validate-rbs
jbeckwith-oai: Fixed in b339aae2. The fast path now runs documented `rbs parse sig` before creating the consolidated buffer, so every original file must parse independently. Any parse failure skips consolidation and routes through the existing ordinary-Steep fallback for exact source diagnostics. I added your split `class Example` / `end` reproduction as a regression test; it now fails with an original-file `RBS::SyntaxError`. Verification: 9 tests / 45 assertions in local and simulated-CI modes, all 1,209 current signatures pass, full RuboCop (2,589 files) is clean, and the preflight measured 0.40s (6.22s complete optimized path). I also expanded the PR safety section to document this boundary and tradeoff.

scripts/validate-rbs
```diff
@@ -0,0 +1,118 @@
...
+        status
+      end
+    end
+    private_class_method :run_consolidated_check
+
+    def concatenate(signatures, destination)
+      destination.open("wb") do |combined|
+        signatures.each do |path|
+          contents = path.binread
+          combined.write(contents)
```
HAYDEN-OAI, line 102: **[P1] Fall back before concatenating NUL-containing signatures**

Locked RBS 3.9.5 treats `\x00` as EOF, so this byte can silently terminate the consolidated parser buffer even though every original file passes the new independent preflight. With `sig/a.rbs` containing `module Good\nend\n\x00` and `sig/b.rbs` containing `module Invalid\n  type bad = MissingType\nend\n`, I reproduced `rbs parse sig` exiting 0 and exact-head `ruby scripts/validate-rbs` exiting 0 (`Validated 2 RBS files with no errors.`). Ordinary `steep check --no-type-check --jobs=8 --format=github` instead exits 1 and annotates `sig/b.rbs:2` with `RBS::UnknownTypeName`; the combined buffer never reaches that declaration because `combined.write(contents)` embeds the earlier EOF. Thus a NUL-containing signature can disable semantic validation of every subsequent file while `ci-required` remains green. Detect NUL bytes before merging and route those signatures through the ordinary per-file fallback, or otherwise preserve original parser-buffer boundaries.

HAYDEN-OAI (commented):
Re-reviewed exact head `b339aae226a9100db4c25e60046cbeb12d8e2463`, all three commits/four changed files, prior review #4848718240 and its resolution, and the locked RBS 3.9.5 / Steep 1.10.0 execution paths. The split `class`/`end` issue is fixed: independent parsing now fails before consolidation and ordinary Steep preserves original-file diagnostics. File-scoped directives, fallback workers/GitHub annotations, semantic errors, and the required-check dependency were also verified. However, the inline P1 demonstrates a distinct, independently reproduced consolidation-boundary false green: an embedded NUL terminates RBS lexing and can hide every later invalid signature. Submitting COMMENT because this still bypasses the required validation gate.

scripts/validate-rbs
jbeckwith-oai: Fixed in ea5d75b0. The consolidation eligibility check now reads each signature as binary and routes the entire validation through ordinary Steep whenever any file contains a NUL byte, before a combined buffer can be created. I added your exact two-file reproduction as a regression test; it now exits nonzero and reports `RBS::UnknownTypeName` against `b.rbs`. Verification: 10 tests / 50 assertions pass in both local and simulated-CI modes, all 1,209 repository signatures validate, focused RuboCop is clean, and the thermo-nuclear maintainability review found no structural blockers.

scripts/validate-rbs
```diff
@@ -0,0 +1,121 @@
...
+require "fileutils"
+require "open3"
+require "pathname"
+require "tmpdir"
+
+module OpenAI
+  module RBSValidation
+    # These directives change name resolution on a per-file basis, so signatures
+    # containing them cannot be safely combined and must take the reference path.
+    FILE_SCOPED_DIRECTIVE = /^\s*(?:use(?:\s|$)|#\s*resolve-type-names\s*:)/
```
HAYDEN-OAI: **[P1] Detect whitespace-free RBS `use` directives before consolidation**

Locked RBS 3.9.5 accepts `use::Example::Foo` (and `use::Example::*` or comment-adjacent `use# ...`) as a file-scoped `RBS::AST::Directives::Use`, but `use(?:\s|$)` misses these valid keyword boundaries. I independently reproduced the resulting false green with:

`sig/a.rbs`:
```rbs
use::Example::Foo
module Example
  class Foo
  end
end
```

`sig/b.rbs`:
```rbs
module Invalid
  type bad = Foo
end
```

`rbs parse sig` exits 0, and ordinary `steep check --no-type-check --jobs=8 --format=github` exits 1 with `RBS::UnknownTypeName` annotated at `sig/b.rbs:2`; exact-head `ruby scripts/validate-rbs` instead exits 0 and prints `Validated 2 RBS files with no errors.` Consolidation incorrectly applies the first file's import to the second, so invalid signatures can again satisfy `ci-required`. Match the actual `use` keyword boundary (for example `use\b`) or conservatively bypass consolidation for every lexer-valid `use` directive.

HAYDEN-OAI (commented):
Re-reviewed exact head `ea5d75b0f7ce15b5cced4835da0e652569576476`, all four commits/changed files, all prior conversations, and prior P1 review #4849204675. On locked RBS 3.9.5 / Steep 1.10.0, independently verified that NUL-containing signatures now route directly to ordinary Steep with the correct original-file diagnostic, the earlier split-file syntax regression stays fixed, and fallback workers/GitHub annotations, semantic validation, and `ci-required` wiring remain intact. However, the inline P1 independently reproduces another consolidation-boundary false green: valid whitespace-free/comment-adjacent `use` directives bypass the guard, leak file-scoped imports into later signatures, and let the required validation gate succeed when ordinary Steep fails. Submitting COMMENT because this still weakens required RBS validation.

scripts/validate-rbs
jbeckwith-oai: Fixed in 8b6a690c. The file-scoped directive guard now uses the actual Ruby/RBS keyword boundary (`use\b`), so whitespace, `::`, wildcard imports, and comment adjacency all bypass consolidation and run ordinary per-file Steep. I added your exact `use::Example::Foo` two-file reproduction as a regression test; it now exits nonzero and reports `RBS::UnknownTypeName` against `b.rbs`. Verification: 11 tests / 55 assertions pass in both local and simulated-CI modes, all 1,209 repository signatures validate, focused RuboCop and Ruby syntax checks are clean, and the thermo-nuclear review found no structural blockers.

HAYDEN-OAI (commented):
Re-reviewed exact head `8b6a690c81b78f5a7e6248c061eb4e099276ef36`, all five commits/four changed files, every prior review conversation, and the locked RBS 3.9.5 / Steep 1.10.0 parser and validator behavior. Prior P1 #4849278617 is fixed: exhaustive valid `use` keyword-boundary probes now take ordinary Steep, and the earlier NUL/split-file protections, original-file GitHub annotations, fallback workers, semantic validation, and `ci / ci-required` dependency remain intact. However, the inline P1 independently reproduces another required-validation false green: a valid multiline `resolve-type-names: false` magic directive evades the physical-line guard and suppresses a later original-file semantic error. Submitting COMMENT because the required RBS gate can still pass invalid signatures.

scripts/validate-rbs
```diff
@@ -0,0 +1,121 @@
...
+    def signature_paths(root)
+      Dir.glob(root.join("sig/**/*.rbs")).map { Pathname(_1) }
+    end
+    private_class_method :signature_paths
+
+    def requires_reference_check?(path)
+      contents = path.binread
+      # RBS 3.9 treats NUL as EOF. In a combined buffer, that would hide every
+      # declaration from later files even though the originals parse separately.
+      contents.include?("\0") || contents.each_line.any? { _1.match?(FILE_SCOPED_DIRECTIVE) }
```
HAYDEN-OAI: **[P1] Detect multiline RBS magic directives before consolidation**

Locked RBS 3.9.5 accepts multiline magic directives because `RBS::Parser.magic_comment` matches `\s*` across newlines, but `contents.each_line.any? { _1.match?(FILE_SCOPED_DIRECTIVE) }` never sees `#` and `resolve-type-names:` in the same physical line. I independently reproduced a required-check false green with:

`sig/a.rbs`:
```rbs
#
resolve-type-names: false
module Good
end
```

`sig/b.rbs`:
```rbs
class Broken
  def call: () -> Missing
end
```

On locked RBS 3.9.5 / Steep 1.10.0, `rbs parse sig` exits 0; ordinary `steep check --no-type-check --jobs=2 --format=github` exits 1 and emits `::error file=sig/b.rbs,line=2,...::[RBS::UnknownTypeName] Cannot find type Missing`; the exact-head wrapper instead exits 0 and prints `Validated 2 RBS files with no errors.` The first file's `ResolveTypeNames(false)` leaks across the consolidated buffer and suppresses validation of the later file, allowing invalid signatures to satisfy `ci / ci-required`. Match this magic directive against complete file contents using the locked parser grammar, including multiline whitespace, and route the entire job through ordinary per-file Steep.

jbeckwith-oai: Fixed in `1a8c4498`. The eligibility scan now applies `FILE_SCOPED_DIRECTIVE` to the complete binary file contents instead of each physical line, matching locked RBS 3.9.5's newline-spanning `\s*` behavior. This routes multiline magic directives such as `#\nresolve-type-names: false` directly through ordinary per-file Steep, so their semantics cannot leak into later signatures in the consolidated buffer.

I added the exact two-file regression from this thread. It failed before the fix because the wrapper exited 0, and now fails closed with `RBS::UnknownTypeName` against `b.rbs`. Verification: 12 tests / 60 assertions in normal mode, the same 12 / 60 with `CI=1`, all 1,209 repository RBS files through the real validator, focused RuboCop, Ruby syntax, `git diff --check`, and a full thermo-nuclear maintainability review all pass.

HAYDEN-OAI (approved):
Re-reviewed exact head `1a8c4498facc6c5ede04c662e44bfcf5fc41c5f6`, all six commits and four changed files, every prior review conversation, and locked RBS 3.9.5 / Steep 1.10.0. The P1 from review #4849370345 is fixed: complete-binary-file matching covers the actual parser magic-comment grammar, including newline-spanning whitespace. An executable exhaustive audit checked 1,242,528 directive spellings (776,580 parser-accepted; zero missed) and 7,500 `use` inputs (1,014 parser-accepted; zero missed). Independently reproduced the earlier split-file, embedded-NUL, whitespace-free `use`, multiline-directive, and additional vertical-tab/form-feed cases: each now takes ordinary per-file Steep, rejects invalid signatures, and preserves the exact original-file GitHub annotation. Also verified parser boundaries, comments, line endings, encodings, BOM, filesystem enumeration, fallback workers/formatter, semantic validation, temporary Steepfile resolution, and explicit `validate-rbs` enforcement by `ci / ci-required`. No substantive findings; approving.

PR #339, opened 2026-08-06
fix(internal): encode nested multipart fields with bracket notation

### LINKED ISSUE #212: Files - expires_after was unexpected
Hello, 

According to the [documentation](https://platform.openai.com/docs/api-reference/files/create), I could pass an "expires_after" parameter to automatically delete file. 
However, when I pass it, I got this error back:

```
$openai = OpenAI::Client.new(
  api_key: Rails.application.credentials.openai,
  max_retries: 1
)

$openai.files.create(
  file: download,
  purpose: "user_data", 
   expires_after: {anchor: 'created_at', seconds: 3600}
 )

private 

  def download
    file = Tempfile.new([ "upload", ".pdf" ])
    file.binmode
    file.write(@file.download)
    file.rewind
    Pathname(file.path)
  end

```

Response

 {:url=>"https://api.openai.com/v1/files", :status=>400, :body=>{:error=>{:message=>"Additional properties are not allowed ('expires_after' was unexpected)", :type=>"invalid_request_error", :param=>nil, :code=>nil}}} (OpenAI::Errors::BadRequestError)

leoplct:
The OpenAI API expects nested fields in multipart/form-data bodies to be flattened deepObject-style (e.g. `expires_after[anchor]=created_at`), but nested hashes were serialized as a single JSON part, which the API rejects with "Additional properties are not allowed ('expires_after' was unexpected)".

Flatten nested hashes and arrays inside multipart bodies using bracket notation, matching openai-python's `_serialize_multipartform` (qs "brackets" array format). Top-level primitive arrays keep the existing repeated-key encoding.

Fixes #212

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -616,6 +616,32 @@ module OpenAI
 
         # @api private
         #
+        # Flattens nested hashes into `deepObject` style bracket notation
+        # (`expires_after[anchor]`), matching how the OpenAI API expects nested
+        # multipart fields to be encoded.
+        #
+        # @param y [Enumerator::Yielder]
+        # @param boundary [String]
+        # @param key [Symbol, String]
+        # @param val [Object]
+        # @param closing [Array<Proc>]
+        private def write_multipart_value(y, boundary:, key:, val:, closing:)
+          case val
+          in Hash
+            val.each do |k, v|
+              write_multipart_value(y, boundary: boundary, key: "#{key}[#{k}]", val: v, closing: closing)
+            end
+          in Array
+            val.each do |v|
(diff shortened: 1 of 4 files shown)


jbeckwith-oai:
Thanks for the clear reproduction and for grounding the approach in openai-python’s global brackets format. The recursive nested-object work here is an important part of the complete fix.

We’ve incorporated that behavior into #348 and combined it with the missing bracket names for top-level primitive and file arrays, plus scalar compatibility and broader wire-format coverage. I’m closing this PR in favor of #348 so the consolidated implementation can be reviewed and landed as one change. Appreciate the thoughtful work here.

PR #340, opened 2026-08-07
chore(api): Update SDK generation and API-reference metadata

apcha-oai:
## Summary

chore(api): Update SDK generation and API-reference metadata


jbeckwith-oai (changes requested):
Two blockers in the generated output:

1. The new header is emitted immediately after Sorbet’s `# typed:` magic comment in the RBI files (for example `rbi/openai/client.rbi:2`) without an intervening blank line. RuboCop rejects these with `Layout/EmptyLineAfterMagicComment`, so `ci / lint (rubocop)` and the required CI aggregate are failing. Please update the generator/header placement so magic comments remain a separate leading block and regenerate.

2. This nominally header-only regeneration also removes `minLength: 1` from `NamespaceToolParam.description` and `BetaNamespaceToolParam.description` in `api_reference/openapi.transformed.yml`, weakening those schema contracts. Please regenerate against the same input/transform behavior or restore the constraints so the PR is limited to the intended header addition (plus generation metadata).

PR #341, opened 2026-08-07
feat: support custom HTTP transports

jbeckwith-oai:
## Summary

feat: support custom HTTP transports


lib/openai/internal/util.rb
```diff
@@ -780,7 +786,7 @@ def fused_enum(enum, external: false, &close)
         #
         # @param enum [Enumerable<Object>, nil]
         def close_fused!(enum)
-          return unless enum.is_a?(Enumerator)
+          return unless enum.is_a?(FusedEnumerator)
```
HAYDEN-OAI, line 789: P2 — Preserve cleanup for request enumerators

`ReadIOAdapter#close` still routes request-side `Enumerator` bodies here. For raw or JSONL requests, the body is a plain/lazy enumerator rather than a `FusedEnumerator`; if Net::HTTP aborts mid-upload, this guard returns without unwinding the enumerator, so its `ensure` cleanup (for a file, socket, or similar resource) never runs. Keep the marker restriction for completed response bodies, but wrap request-side enumerators in a closeable fused wrapper or otherwise close them on upload abort.

HAYDEN-OAI (approved):
One low-priority cleanup issue noted inline. Otherwise the exact promoted head looks good: the public tree matches internal source f18c575e2, and I found no further substantive issues in the transport API/runtime, retry and response cleanup, dependency and type surfaces, docs/example, packaging, or Castiron boundaries.

lib/openai/net_http_client.rb
```diff
@@ -0,0 +1,277 @@
...
+      expected_ssl = %w[https wss].include?(url.scheme)
+      unless connection.use_ssl? == expected_ssl
+        raise ArgumentError, "connection configuration must preserve TLS for the requested URL"
+      end
+
+      connection.max_retries = 0
+      connection
+    rescue StandardError
+      begin
+        connection.finish if connection&.started?
```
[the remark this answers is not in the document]
jbeckwith-oai: Addressed upstream in openai-ruby-internal commit 9f9818760 and re-promoted as #342. configured_connection now explicitly initializes connection to nil before connect, keeping the existing cleanup behavior while satisfying definite-assignment analysis.

lib/openai/internal/util.rb
jbeckwith-oai: Addressed upstream in openai-ruby-internal commit 9f9818760 and re-promoted as #342. ReadIOAdapter now uses a private Fiber cancellation boundary: closing a partially read request unwinds the suspended source and runs its ensure without draining another chunk, while closing before the first read does not start user code. The replacement includes regressions for both cases.

jbeckwith-oai:
Superseded by #342, which is a fresh Castiron promotion bound to the updated and fully validated internal source head 9f9818760.

PR #342, opened 2026-08-07
feat: support custom HTTP transports

jbeckwith-oai:
## Summary

feat: support custom HTTP transports


lib/openai/net_http_client.rb
```diff
@@ -0,0 +1,279 @@
...
+    NETWORK_ERRORS = [
+      EOFError,
+      IOError,
+      SocketError,
+      SystemCallError,
+      OpenSSL::SSL::SSLError,
+      Net::HTTPBadResponse,
+      Net::HTTPHeaderSyntaxError,
+      Net::ProtocolError,
+      Zlib::Error,
```
HAYDEN-OAI, line 24: **P2 — Guard the optional zlib constant**

`net/http` rescues `LoadError` when the optional zlib extension is unavailable and continues with decompression disabled, so `Zlib` is not defined on those otherwise usable Ruby installations. Because this class body now resolves `Zlib::Error` during `require "openai"`, the SDK raises `NameError` before any client can be created. Build `NETWORK_ERRORS` conditionally when `defined?(Zlib::Error)`, or make zlib an explicit runtime dependency.

HAYDEN-OAI (approved):
Found one low-priority portability issue, left inline. The custom transport boundary, per-attempt request re-encoding and replayability handling, response cleanup, configurator isolation, and request-abort unwind paths otherwise look sound at this head. Approving because the remaining issue is limited to Ruby builds without the optional zlib extension.

PR #352, opened 2026-08-10
feat: expose request IDs

jbeckwith-oai:
## Summary

- expose the server-generated `x-request-id` as `_request_id` on successful typed model and paginated responses
- expose `request_id` on API errors
- add RBI and RBS declarations plus user-facing documentation
- cover model responses, page responses, error responses, and serialization behavior

## Why

The Ruby SDK retained response headers at the transport boundary but discarded them while coercing successful responses into typed objects. Errors retained their headers but required callers to inspect the header hash manually. This left Ruby behind the other official SDKs for request-ID convenience and made production troubleshooting harder.

Request IDs are response metadata, so they are attached only to the top-level returned object and are excluded from `to_h`, JSON, and YAML serialization.

## Validation

- thermo-nuclear code-quality review
- `./scripts/test` on Ruby 4.0.6 — 518 runs, 1702 assertions
- `./scripts/test` on Ruby 3.3.12 — 518 runs, 1702 assertions
- `bundle exec rake lint:rubocop` — 2596 files, no offenses
- `bundle exec rake validate:rbs` — 1211 files

Surface gap: https://openai-sdk-surface-atlas.openai.chatgpt.site/?capability=request-id#matrix

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,11 +1,5 @@
 # Contributor instructions
 
-## Pull requests
-
-- Before pushing code, run the `thermo-nuclear-code-quality-review` skill and address its findings.
-- After submitting a pull request, monitor CI, fix failures, and push the fixes. Once CI is green, ask for review in `#sdk-reviews`, create a thread, and keep follow-up asks in that thread.
-- When addressing pull request feedback, push the fix, reply to each comment with what changed, and then resolve the thread.
-
 ## Ruby implementation guidelines
 
 - Prefer direct method calls over Ruby reflection (`send`, `__send__`, or `public_send`) for internal SDK plumbing. When an internal method must be callable across components without becoming supported public API, keep the method public for direct dispatch and mark it `@api private`. Keep its RBI and RBS declarations at the same visibility.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,11 @@
+# Contributor instructions
+
+## Pull requests
+
+- Before pushing code, run the `thermo-nuclear-code-quality-review` skill and address its findings.
+- After submitting a pull request, monitor CI, fix failures, and push the fixes. Once CI is green, ask for review in `#sdk-reviews`, create a thread, and keep follow-up asks in that thread.
+- When addressing pull request feedback, push the fix, reply to each comment with what changed, and then resolve the thread.
+
+## Ruby implementation guidelines
+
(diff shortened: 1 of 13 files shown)


lib/openai/internal/transport/base_client.rb
```diff
@@ -639,7 +639,11 @@ def request(req)
             page.new(client: self, req: req, headers: response.headers, page_data: decoded)
           else
             unwrapped = OpenAI::Internal::Util.dig(decoded, unwrap)
-            OpenAI::Internal::Type::Converter.coerce(model, unwrapped)
+            OpenAI::Internal::Type::Converter.coerce(model, unwrapped).tap do |result|
+              if result.is_a?(OpenAI::Internal::Type::BaseModel)
+                result.__send__(:_set_request_id, response.headers["x-request-id"])
```
apcha-oai: can we avoid send here preferably (ruby reflection)

Think the way to do that would be `_set_request_id` public but mark it @api private? Then we can call it directly and avoid __send__.

jbeckwith-oai: Fixed in 15f525ee: `_set_request_id` is now public with `@api private`, and the transport calls it directly. RBI and RBS visibility were updated to match. I also added an `AGENTS.md` guardrail to prefer direct internal calls over reflection. Tests, RuboCop, Sorbet, and RBS validation all pass.

AGENTS.md
```diff
@@ -0,0 +1,11 @@
+# Contributor instructions
+
+## Pull requests
+
+- Before pushing code, run the `thermo-nuclear-code-quality-review` skill and address its findings.
+- After submitting a pull request, monitor CI, fix failures, and push the fixes. Once CI is green, ask for review in `#sdk-reviews`, create a thread, and keep follow-up asks in that thread.
```
apcha-oai: nit: feel like this (and the direct below) should probably not be defaults / we can make internal skill or something for this

apcha-oai (approved):

AGENTS.md
jbeckwith-oai: Addressed in 56fa2f26: removed the entire generic pull-request workflow section, including the CI, Slack, and feedback-handling defaults. `AGENTS.md` now retains only the focused Ruby reflection guideline; process automation can live in an internal skill instead.

PR #345, opened 2026-08-07
fix: harden custom HTTP transports

jbeckwith-oai:
## Summary

- harden custom HTTP transport retries, redirects, timeout mapping, decompression errors, and connection lifecycle
- preserve one-shot request semantics for IO, Enumerator, and arbitrary Enumerable bodies
- move request-body adaptation into an SDK-owned internal component, keeping Castiron-owned root/client files unchanged
- add focused coverage for cleanup, replayability, malformed responses, configurator failures, and transport loading

## Regeneration boundary

This promotion contains only SDK-owned runtime, signature, and test files. `lib/openai.rb` and `lib/openai/client.rb` are unchanged from public `main`, so the hardening is additive to the generated API and will not be overwritten by Castiron regeneration.

Source: openai/openai-ruby-internal#15 at `89927dd58c9742c6a67823fefdfe2eabbde9ec45`.

## Validation

- internal CI: Ruby 3.3, 3.4, and 4.0
- RuboCop, Sorbet/RBI, RBS, and package validation
- CodeQL
- local full suite: 514 tests / 1,691 assertions

PR #354, opened 2026-08-10
fix: preserve binary multipart stream reads

### LINKED ISSUE #317: Encoding::CompatibilityError during multipart file upload is swallowed and reported as a generic APIConnectionError ("Connection error.")

jbeckwith-oai:
## Summary

- preserve arbitrary binary bytes when multipart request bodies yield a mix of UTF-8 and binary chunks
- honor `IO#read(max_len, out_string)` byte limits, buffering, EOF, negative-length, and output-buffer semantics for Enumerator-backed bodies
- retain native `IO` and `StringIO` delegation, including implicit `to_int` length conversion
- update RBI/RBS contracts and add focused regression coverage

This fixes #317 and consolidates the complete behavior needed from #275, #294, and #324 on the current SDK-owned `ReadIOAdapter` implementation.

## Ownership

This is handwritten SDK runtime behavior introduced as an SDK-owned component in #345. It does not change OpenAPI inputs, Castiron facts, renderer rules, templates, or generator configuration.

## Validation

- `mise exec ruby@4.0.6 -- ./scripts/test` — 531 runs, 1,736 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — RuboCop, Sorbet, and 1,211 RBS files clean
- `mise exec ruby@4.0.6 -- bundle exec rake build:gem`
- installed-gem adapter probe
- Ruby 3.3.12 mixed-encoding compatibility probe
- 1,000 randomized read sequences compared against `StringIO`
- real local HTTP multipart upload with a 1 MiB binary file and UTF-8 prompt, verified byte-for-byte by SHA-256

### DIFF
--- a/test/openai/internal/util_test.rb
+++ b/test/openai/internal/util_test.rb
@@ -445,17 +445,15 @@ class OpenAI::Test::UtilIOAdapterTest < Minitest::Test
   def test_io_read_preserves_native_length_conversion
     length = Object.new
     def length.to_int = 2
-    # rubocop:disable Lint/EmptyBlock
-    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(StringIO.new("abc")) {}
-    # rubocop:enable Lint/EmptyBlock
+    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(
+      StringIO.new("abc")
+    ) { |_chunk| nil }
 
     assert_equal("ab", adapter.read(length))
   end
 
   def test_enum_read_respects_max_len
-    # rubocop:disable Lint/EmptyBlock
-    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(%w[abc def].to_enum) {}
-    # rubocop:enable Lint/EmptyBlock
+    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(%w[abc def].to_enum) { |_chunk| nil }
 
     assert_equal("", adapter.read(0))
     assert_equal("a", adapter.read(1))
@@ -466,9 +464,7 @@ class OpenAI::Test::UtilIOAdapterTest < Minitest::Test
 
   def test_enum_read_preserves_mixed_encoding_chunks_by_byte_length
     chunks = ["caf\u00E9", "\xFF\xFE".b]
-    # rubocop:disable Lint/EmptyBlock
-    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(chunks.to_enum) {}
-    # rubocop:enable Lint/EmptyBlock
+    adapter = OpenAI::Internal::Util::ReadIOAdapter.new(chunks.to_enum) { |_chunk| nil }
     actual = String.new.b
 
     while (chunk = adapter.read(2))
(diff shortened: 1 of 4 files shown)


test/openai/internal/util_test.rb
```diff
@@ -442,6 +442,89 @@ def test_read_all_from_enumerator
     assert_equal("hello world", adapter.read)
   end
 
+  def test_io_read_preserves_native_length_conversion
+    length = Object.new
+    def length.to_int = 2
```
apcha-oai, line 447: non-blocking nit:

avoid disabling lints + prefer using mocks:
```
def test_io_read_preserves_native_length_conversion
  length = Minitest::Mock.new
  length.expect(:to_int, 2)

  adapter = OpenAI::Internal::Util::ReadIOAdapter.new(
    StringIO.new("abc")
  ) { |_chunk| nil }

  assert_equal("ab", adapter.read(length))
  assert_mock(length)
end
```

apcha-oai (approved):

test/openai/internal/util_test.rb
jbeckwith-oai: Addressed in `f711f478`. I removed all seven new `Lint/EmptyBlock` suppressions and replaced the empty callbacks with explicit `{ |_chunk| nil }` blocks.

I kept the minimal concrete `to_int` protocol object rather than the exact bare `Minitest::Mock` shown here: `StringIO#read` performs this coercion in C and rejects a method-missing-backed mock with `TypeError` even though the mock reports `respond_to?(:to_int)`. Adding forwarding/mock plumbing would make the test more coupled and less clear. The return assertion still proves native implicit conversion is preserved.

Validated with the focused adapter suite (53 runs / 204 assertions), the full suite (531 / 1,736), RuboCop, Sorbet, and RBS.

PR #357, opened 2026-08-10
docs: document fiber scheduler concurrency

jbeckwith-oai:
## Summary

- document that the default `Net::HTTP` transport cooperates with an active Ruby fiber scheduler
- add an `async` usage example without introducing a runtime dependency
- add a localhost integration test proving two requests can make progress concurrently on one scheduler thread

## Why

The [SDK Surface Atlas](https://openai-sdk-surface-atlas.openai.chatgpt.site/?capability=async&sdk=ruby#matrix) classifies Ruby async support as a gap because the official SDK does not document a native path. The existing transport already uses scheduler-aware Ruby I/O, so this makes that behavior an explicit, tested SDK contract instead of adding a duplicate async client hierarchy.

Without an active scheduler, requests retain their existing blocking behavior.

## Validation

- `mise exec ruby@4.0.6 -- ./scripts/test` — 525 runs, 1,712 assertions
- `mise exec ruby@4.0.6 -- bundle exec rake lint:rubocop` — 2,597 files, no offenses
- focused scheduler integration test on Ruby 3.3.12

### DIFF
--- a/README.md
+++ b/README.md
@@ -671,11 +671,12 @@ response = client.request(
 #### Asynchronous requests with a fiber scheduler
 
 The default HTTP client also cooperates with Ruby's fiber scheduler interface
-(`Fiber.scheduler`). When a scheduler is active, network operations, streaming
-reads, retry delays, and waits for a pooled connection yield to other fibers
-instead of blocking the thread. Resource methods keep their normal return types;
-schedule each complete request or stream operation as a task using the scheduler
-implementation your application already uses.
+(`Fiber.scheduler`). When a request runs inside a non-blocking fiber managed by
+an active scheduler, network operations, streaming reads, retry delays, and
+waits for a pooled connection yield to other fibers instead of blocking the
+thread. Resource methods keep their normal return types; schedule each complete
+request or stream operation as a task using the scheduler implementation your
+application already uses.
 
 For example, with the [`async`](https://github.com/socketry/async) gem:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -668,6 +668,36 @@ response = client.request(
 
 `OpenAI::Client` instances using the default `OpenAI::NetHTTPClient` are threadsafe, but are only fork-safe when there are no in-flight HTTP requests. Injected HTTP clients are responsible for documenting and enforcing their own concurrency guarantees.
 
+#### Asynchronous requests with a fiber scheduler
+
+The default HTTP client also cooperates with Ruby's fiber scheduler interface
+(`Fiber.scheduler`). When a scheduler is active, network operations, streaming
```
HAYDEN-OAI: Low priority: could we qualify this as “when a request runs inside a non-blocking fiber managed by an active scheduler”? `Fiber.scheduler` being non-nil alone does not make I/O, sleeps, or pool waits cooperative; calls from the root/blocking fiber can still block the thread. The Async example below does the right thing, but this sentence currently promises more than Ruby provides.

HAYDEN-OAI (approved):
Overall looks good. The Async example and transport behavior line up; I left one low-priority wording clarification inline.

README.md
jbeckwith-oai: Updated the README to scope cooperative behavior to requests running inside non-blocking fibers managed by an active scheduler, and clarified that calls outside that context remain blocking. I also aligned the `OpenAI::NetHTTPClient` class documentation with the same constraint. Pushed in 5b90f02f.

PR #360, opened 2026-08-11
fix(api): Add new model identifiers and remove audit log source

jbeckwith-oai:
## Summary

feat(api): Add new model identifiers and remove audit log source


apcha-oai (approved):

HAYDEN-OAI (approved):
Reviewed the generated Responses-only model identifier additions and the intentional audit-log `source` schema removal. No substantive issues found.

PR #364, opened 2026-08-11
feat: add Tapioca typing for structured outputs

### LINKED ISSUE #309: Provide sound Sorbet typing for application-owned structured-output models
Carried over from #126 and originally requested by @agarcher.

Applications need a sound, idiomatic Sorbet typing experience for structured-output responses, including typed access to nested application-owned models after response parsing. The existing SDK keeps response `parsed` fields broad because callers can supply different structured-output models.

The earlier `OpenAI::BaseModel` Tapioca compiler from #364 treated the proposed implementation as the requested outcome. Its generated reader signatures claimed nested model values even though directly constructed or assigned SDK transport models deliberately preserve caller-provided hashes. #385 attempted to reconcile that mismatch with additional runtime coercion, but architecture review established that retrofitting SDK transport models into an application typing framework is the wrong abstraction. The unreleased compiler is being removed separately while retaining the independent load-order and nullable-array fixes.

## Desired outcome

- Give applications a supported, idiomatic way to use their own strongly typed structured-output models and obtain correctly typed nested values from parsed API responses.
- Evaluate purpose-built application modeling tools, such as an optional Sorbet `T::Struct` integration that derives JSON Schema and hydrates responses at the structured-output boundary; do not treat this example as an approved public API.
- Preserve SDK transport-model construction and assignment semantics, caller-provided raw request identity, and existing generated request-model behavior. Do not introduce global nested coercion or reader signatures that contradict runtime values.
- Keep `sorbet-runtime` optional and define dependency loading, supported schema/model features, validation or hydration errors, and compatibility guarantees before selecting a public API.
- Cover representative nested models, arrays, nilability, enums, and unions where the selected design supports them.
- Add public-boundary regression coverage for both Chat Completions and Responses structured-output flows, together with construction/assignment and generated-signature checks that prove static types match runtime behavior.

The underlying customer need remains open; the compiler implementation from #364 is not the desired architectural direction.

jbeckwith-oai:
## Summary

- ship a Tapioca DSL compiler for application-defined `OpenAI::BaseModel` subclasses
- generate typed readers for nested models, arrays, enums, unions, and nilable fields
- preserve the SDK's no-`sorbet-runtime` runtime dependency and document the required `T.cast` for broad `parsed` values
- fix Sorbet conversion for nilable `OpenAI::ArrayOf` elements
- make real `tapioca dsl` loading work after Thor removes the subcommand from `ARGV`

Fixes #309.

## Validation

- `mise x ruby@4.0.6 -- ./scripts/test` — 571 runs, 1,994 assertions
- `mise x ruby@3.3.12 -- ./scripts/test` — 571 runs, 1,994 assertions
- `mise x ruby@4.0.6 -- bundle exec rake lint` — Sorbet, 1,211 RBS files, and 2,603 RuboCop files pass
- real `bundle exec tapioca dsl --list-compilers --only OpenAIBaseModel` discovery
- built and installed the gem in isolation; verified runtime loading and packaged compiler generation

## Review

Ran the repository-required thermo-nuclear maintainability review plus a two-pass independent correctness review. The second pass caught and drove the real CLI load-order regression test; no actionable findings remain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -821,6 +821,26 @@ params = OpenAI::Chat::CompletionCreateParams.new(
 openai.chat.completions.create(**params)
 ```
 
+### Structured output models
+
+The SDK includes a Tapioca DSL compiler for application-defined subclasses of
+`OpenAI::BaseModel`. When Tapioca loads your application, running
+`bundle exec tapioca dsl` generates typed readers for fields declared with
+`required`, including nested models, arrays, enums, unions, and fields declared
+with `nil?: true`.
+
+Response `parsed` fields can contain different application-defined models, so
+their generated SDK type remains broad. Cast a parsed value to the structured
+output model supplied with the request before accessing its generated readers:
+
+```ruby
+event = T.cast(content.parsed, CalendarEvent)
+puts(event.name)
+```
+
+The compiler is only loaded by Tapioca; using the SDK normally still does not
+require `sorbet-runtime`.
+
 ### Enums
 
 Since this library does not depend on `sorbet-runtime`, it cannot provide [`T::Enum`](https://sorbet.org/docs/tenum) instances. Instead, we provide "tagged symbols" instead, which is always a primitive at runtime:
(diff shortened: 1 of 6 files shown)


HAYDEN-OAI (approved):
Reviewed the Tapioca compiler discovery/load-order path, generated field typing, nilable array handling, and packaging boundary. No substantive issues found.

PR #374, opened 2026-08-12
test: run examples end to end against the live API

jbeckwith-oai:
## Summary

- add a `workflow_dispatch`-only suite that runs 22 Ruby examples against the live OpenAI API
- classify every example as covered or explicitly excluded, validate expected output, and upload JSON/Markdown result reports
- update image examples and README usage to `gpt-image-2`

## Fixes uncovered by the live suite

- use `response_id` when resuming response streams
- require the chat streaming examples to receive non-whitespace model content before reporting success

## Impact

The live suite is opt-in and only runs through manual workflow dispatch. A dispatched run fails if an example fails or the inventory becomes incomplete. Inventory validation also runs in required CI without making live API requests.

## Validation

- 22/22 runnable examples passed against the live API on Ruby 4.0
- both strengthened chat streaming checks rerun successfully against the live API
- 723 tests, 3,350 assertions
- 2,622 files passed RuboCop
- Sorbet and 1,212 RBS files validated successfully


lib/openai/models/completion_choice.rb
```diff
@@ -9,8 +9,8 @@ class CompletionChoice < OpenAI::Internal::Type::BaseModel
       #   number of tokens specified in the request was reached, or `content_filter` if
       #   content was omitted due to a flag from our content filters.
       #
-      #   @return [Symbol, OpenAI::Models::CompletionChoice::FinishReason]
-      required :finish_reason, enum: -> { OpenAI::CompletionChoice::FinishReason }
+      #   @return [Symbol, OpenAI::Models::CompletionChoice::FinishReason, nil]
+      required :finish_reason, enum: -> { OpenAI::CompletionChoice::FinishReason }, nil?: true
```
apcha-oai: (blocking) This is generated code / the same across all SDKs, we shouldn't manually add this and needs to be fixed in openapi or backend

rbi/openai/models/completion_choice.rbi
```diff
@@ -12,7 +12,7 @@ module OpenAI
       # hit a natural stop point or a provided stop sequence, `length` if the maximum
       # number of tokens specified in the request was reached, or `content_filter` if
       # content was omitted due to a flag from our content filters.
-      sig { returns(OpenAI::CompletionChoice::FinishReason::TaggedSymbol) }
+      sig { returns(T.nilable(OpenAI::CompletionChoice::FinishReason::TaggedSymbol)) }
```
apcha-oai: similar

sig/openai/models/completion_choice.rbs
```diff
@@ -18,14 +18,14 @@ module OpenAI
...
-        finish_reason: OpenAI::Models::CompletionChoice::finish_reason,
+        finish_reason: OpenAI::Models::CompletionChoice::finish_reason?,
         index: Integer,
         logprobs: OpenAI::CompletionChoice::Logprobs?,
         text: String
       ) -> void
 
       def to_hash: -> {
-        finish_reason: OpenAI::Models::CompletionChoice::finish_reason,
+        finish_reason: OpenAI::Models::CompletionChoice::finish_reason?,
```
apcha-oai: similar

test/scripts/examples_e2e_test.rb
```diff
@@ -0,0 +1,237 @@
...
+
+require "fileutils"
+require "pathname"
+require "stringio"
+require "tmpdir"
+
+require_relative "../../scripts/examples-e2e"
+require_relative "../openai/test_helper"
+
+class OpenAI::Test::ExamplesE2ETest < Minitest::Test
```
apcha-oai: nit: tests for tests; imo we can drop / simplify

apcha-oai (changes requested):

.simplecov
```diff
@@ -0,0 +1,20 @@
...
+SimpleCov.configure do
+  root project_root.to_s
+  coverage_path coverage_dir.to_s
+  cover "lib/**/*.rb"
+  merge_timeout 7_200
+  formatters [SimpleCov::Formatter::HTMLFormatter, SimpleCov::Formatter::JSONFormatter]
+
+  coverage :line do
+    # Raise this baseline whenever example E2E coverage increases. Never lower it.
+    minimum 89.96
```
apcha-oai: (nonblocking) IMO this creates some pretty odd incentives that are probably bad for codex.

- vendoring in code, adjusting internals, etc. can increase line count without expanding public surface area
- that code should be covered by unit tests

If we want line coverage, that should come from the unit test suite.
For examples, would want something like percentage of the public API exercised, although that’s harder to define and compute.

lib/openai/models/completion_choice.rb
apcha-oai: poked around looks like a very old openapi bug we can fix upstream

test/scripts/examples_e2e_test.rb
jbeckwith-oai: Removed the dedicated tests-for-tests suite entirely and kept the example inventory validation plus the live example run as the checks for this workflow.

.simplecov
jbeckwith-oai: Removed SDK line coverage completely: the SimpleCov dependency/configuration, subprocess instrumentation, coverage collation and threshold, coverage-specific test, and related documentation/workflow references are gone. The suite now reports only which examples are exercised or explicitly excluded.

.github/workflows/examples-e2e.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@95ef2b042f9d7a56d8268cba8559e2842e2ad01b # v1
+        with:
+          ruby-version: "4.0"
+          bundler-cache: true
+      - name: Run examples against the live API
+        env:
+          EXAMPLES_E2E_REPORT_DIR: ${{ runner.temp }}/examples-e2e
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: bundle exec rake test:examples:e2e
```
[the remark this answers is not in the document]
jbeckwith-oai: The SDK line-coverage floor was intentionally removed in response to the maintainer’s review feedback. I also updated the PR description, contributor documentation, dependencies, workflow artifact label, and runner so nothing advertises or expects SimpleCov reports anymore; example-inventory coverage remains.

.github/workflows/examples-e2e.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  examples-e2e:
+    name: live Ruby examples
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
```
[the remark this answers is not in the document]
jbeckwith-oai: Raised the workflow timeout to 90 minutes, which covers the current 22 × 180-second worst-case example budget and still leaves time for setup, report generation, and artifact upload.

scripts/examples-e2e.rb
```diff
@@ -0,0 +1,367 @@
...
+
+      examples.each do |example|
+        unless VALID_STATUSES.include?(example.status)
+          errors << "#{example.path}: status must be one of #{VALID_STATUSES.join(', ')}"
+          next
+        end
+
+        if example.status == "covered"
+          has_expected_output = !example.expected_output.to_s.empty?
+          has_minimum_output = example.minimum_output_bytes.to_i.positive?
```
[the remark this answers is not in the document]
jbeckwith-oai: Inventory validation now requires every configured expected_output to be a non-empty String and rejects values such as YAML false even when minimum_output_bytes is also present. I verified the malformed-manifest case directly.

Rakefile
```diff
@@ -153,6 +154,16 @@ end
 desc("Typecheck and validate everything")
 multitask(typecheck: [:"typecheck:sorbet", :"validate:rbs"])
 
+desc("Validate the Ruby example E2E inventory without making live requests")
+task("test:examples:inventory") do
+  ruby(*%w[scripts/examples-e2e.rb --inventory-only])
```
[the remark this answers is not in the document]
jbeckwith-oai: Made test:examples:inventory a prerequisite of the existing test task, so every required Ruby-version CI test job now validates that every example is classified before running the unit suite.

examples/e2e.yml
```diff
@@ -0,0 +1,78 @@
...
+    expected_output: streams can be lazy
+  examples/azure_openai.rb:
+    status: excluded
+    reason: Requires Azure endpoint, deployment, and API-key credentials.
+  examples/bedrock.rb:
+    status: excluded
+    reason: Requires AWS credentials, region configuration, and a Bedrock model.
+  examples/chat/streaming_basic.rb:
+    status: covered
+    minimum_output_bytes: 1
```
HAYDEN-OAI: [P2] Assert actual streamed content here

`examples/chat/streaming_text.rb` unconditionally calls `puts` after iterating, and `streaming_basic.rb` can emit the same newline on a content-done event even when no text delta was produced. As a result, a text-empty stream still writes the one byte required here (and by the identical setting on line 30), so both examples can be reported as passing without any model output. Please measure non-whitespace output or emit/assert a sentinel only after receiving content.

HAYDEN-OAI (commented):
Found one output-validation gap in the new live example suite.

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

examples/responses/streaming_basic.rb
```diff
@@ -11,13 +11,20 @@
   model: "gpt-4o-2024-08-06"
 )
 
+text_received = T.let(false, T::Boolean)
```
HAYDEN-OAI: [P1] Remove runtime Sorbet calls from standalone examples

The SDK intentionally has no sorbet-runtime dependency, and neither lib/openai.rb nor these examples require it, so this new T.let raises NameError: uninitialized constant T in a normal SDK installation. The same newly introduced T.let/T.cast calls occur in seven covered examples; the new runner also launches each example in a fresh Ruby process, so constants from its parent cannot make them available. As a result, all seven standalone examples fail even when the API succeeds, sometimes after a billable request. Please use ordinary Ruby local assignments and an explicit is_a? check instead of requiring Sorbet at runtime.

HAYDEN-OAI (commented):
Found one substantive runtime compatibility regression in the updated standalone examples; details are in the inline comment.

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

examples/streaming.rb
```diff
@@ -27,6 +29,7 @@
       break
     end
   end
+  abort("The stream completed without yielding data") if stream_data_count.zero?
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `759cb2b3`. `examples/streaming.rb` now requires at least one terminal choice before the example can pass. It reads the existing raw `choice[:finish_reason]` field to avoid the known generated `CompletionChoice` nil-conversion issue, and aborts if no terminal chunk arrives. I applied the same terminal-stream invariant to the other covered legacy/raw stream examples (`advanced_streaming.rb` and `demo.rb`). All five changed examples pass against the live API; full lint, Sorbet, and RBS validation are clean.

examples/e2e.yml
```diff
@@ -0,0 +1,78 @@
...
+    expected_output: parsed outputs from final response
+  examples/streaming.rb:
+    status: covered
+    expected_output: manual closing of stream
+  examples/structured_outputs_chat_completions.rb:
+    status: covered
+    expected_output: CalendarEvent
+  examples/structured_outputs_chat_completions_function_calling.rb:
+    status: covered
+    expected_output: GetWeather
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `759cb2b3`. Both structured function-calling examples now force the named function: Chat Completions uses `tool_choice: {type: :function, function: {name: "GetWeather"}}`, and Responses uses `tool_choice: {type: :function, name: "GetWeather"}`. Both examples pass against the live API and deterministically produce a parsed `GetWeather` call.

HAYDEN-OAI (approved):

apcha-oai (approved):

PR #377, opened 2026-08-13
chore: lint Ruby files repository-wide

jbeckwith-oai:
## Summary

Replace the hand-maintained RuboCop path allowlist with repository-root discovery, while retaining an explicit RBI glob because `.rbi` is not a default RuboCop file type.

This makes new first-party Ruby files and Ruby-bearing directories linted automatically. The owned target set grows from 1,387 to 1,388 files by adding the existing `Steepfile`. `--force-exclusion` keeps dependency/build output out of scope, RuboCop's default exclusions are merged with repository exclusions, and the disposable Tapioca trees are explicitly excluded.

Castiron impact: none. `Rakefile` and `.rubocop.yml` are excluded from Castiron generation, and the existing Castiron validation invokes this SDK-owned `lint:rubocop` task, so it inherits the stronger coverage without a generator change.

## Test plan

- `bundle exec rake lint` — RuboCop inspected 1,388 first-party files with zero offenses; Sorbet and all 1,212 RBS validations passed.
- Controlled first-party probe: added an invalid Ruby file under previously unlisted `api_reference/`; `bundle exec rake lint:rubocop` discovered it and failed.
- Controlled dependency probe: added an invalid RBI under `vendor/bundle`; the lint task ignored it and passed.
- Controlled generated-RBI probe: added an invalid RBI under `sorbet/rbi/gems`; the lint task ignored it and passed.
- All probe files were removed before commit.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,10 @@
 # yaml-language-server: $schema=https://www.rubyschema.org/rubocop.json
 ---
+# Keep RuboCop's dependency and build-output exclusions alongside ours.
+inherit_mode:
+  merge:
+    - Exclude
+
 # Explicitly disable pending cops for now. This is the default behaviour but
 # this avoids a large warning every time we run it.
 # Stop RuboCop nagging about rubocop-rake.
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,7 @@ AllCops:
     - "bin/*"
     - "sorbet/rbi/annotations/**/*"
     - "sorbet/rbi/gems/**/*"
+    - "sorbet/tapioca/**/*"
   NewCops: enable
   SuggestExtensions: false
   TargetRubyVersion: 3.3
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,6 +12,8 @@ inherit_mode:
 AllCops:
   Exclude:
     - "bin/*"
+    - "sorbet/rbi/annotations/**/*"
+    - "sorbet/rbi/gems/**/*"
   NewCops: enable
   SuggestExtensions: false
   TargetRubyVersion: 3.3
--- a/Rakefile
+++ b/Rakefile
@@ -60,7 +60,7 @@ RuboCop::RakeTask.new(:"lint:rubocop") do |task|
   task.formatters = %w[github] if ENV.key?("CI")
 
   # some lines cannot be shortened
-  task.options = %w[--parallel --except Layout/LineLength]
+  task.options = %w[--parallel --force-exclusion --except Layout/LineLength]
 end
 
 norm_lines = %w[tr -- \n \0].shelljoin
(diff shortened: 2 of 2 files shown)


Rakefile
```diff
@@ -56,14 +56,7 @@ end
...
 RuboCop::RakeTask.new(:"lint:rubocop") do |task|
-  task.patterns = FileList[
-    "./{lib,test,rbi,examples}/**/*.rb",
-    "./{lib,test,rbi,examples}/**/*.rbi",
-    "./scripts/validate-rbs",
-    "./Gemfile",
-    "./Rakefile",
-    "./openai.gemspec"
-  ]
+  task.patterns = %w[. ./**/*.rbi]
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `72b9cc7`. The repository-wide task now passes `--force-exclusion`, and `AllCops/Exclude` explicitly lists both disposable Tapioca trees (`sorbet/rbi/annotations/**/*` and `sorbet/rbi/gems/**/*`). I verified this with a deliberately invalid RBI under `sorbet/rbi/gems` (ignored) and reran the full lint/typecheck suite: 1,388 first-party Ruby/RBI files inspected with zero offenses, Sorbet clean, and 1,212 RBS files validated.

.rubocop.yml
```diff
@@ -1,12 +1,19 @@
...
+
 # Explicitly disable pending cops for now. This is the default behaviour but
 # this avoids a large warning every time we run it.
 # Stop RuboCop nagging about rubocop-rake.
 # Ensure that RuboCop validates according to the lowest version of Ruby that we support.
 AllCops:
   Exclude:
     - "bin/*"
+    - "sorbet/rbi/annotations/**/*"
+    - "sorbet/rbi/gems/**/*"
```
HAYDEN-OAI, line 16: One more disposable Tapioca path appears to need the same treatment: `.gitignore` ignores `sorbet/tapioca/*`, and the Rakefile adds the entire `sorbet/tapioca` directory to `CLOBBER`. `tapioca init` can create `sorbet/tapioca/require.rb`, which the new `.` target will now lint even though it is an ignored/generated local artifact. `--force-exclusion` only helps for paths listed here, so please add `sorbet/tapioca/**/*` as well (or otherwise document that this generated Ruby file is intentionally in scope).

HAYDEN-OAI (approved):
Reviewed `72b9cc722f1602d94bf829ec440c9c581f014c99`. Repository-root discovery plus the explicit RBI glob and `--force-exclusion` are a sound replacement for the owned-path allowlist, and the earlier generated-RBI issue is addressed. I left one low-priority inline note for another disposable Tapioca path; it does not block approval.

.rubocop.yml
jbeckwith-oai: Fixed in `4c56750`. I added `sorbet/tapioca/**/*` to the forced RuboCop exclusions. I also created a deliberately invalid `sorbet/tapioca/require.rb` probe and confirmed `lint:rubocop` ignored it; after removing the probe, the full `rake lint` suite passed with 1,388 first-party files clean, Sorbet clean, and all 1,212 RBS files validated.

PR #385, opened 2026-08-13
fix: align structured output readers with Sorbet types

jbeckwith-oai:
## Summary

Fix the contract mismatch between the Tapioca structured-output compiler introduced in #364 and the raw-value compatibility rollback in #375.

- Materialize typed readers only for application-defined `OpenAI::BaseModel` structured-output classes.
- Keep caller-owned hashes and arrays untouched in `[]` and `to_h`, including on directly constructed and reassigned structured-output models.
- Preserve already-materialized parsed-response values, including nullable array identity, and retain precise nested model, array, union, enum, boolean, and nullable Sorbet signatures.
- Fix the root cause of recursive `Symbol` failures once in the SDK-owned primitive converter; `String → Symbol` now works naturally through arbitrary arrays, unions, and nested models without structured-output-specific normalization branches.
- Discard stale errors from viable structured-output union branches, honor explicitly assigned nullable values before inherited converter-error checks, and reject missing/non-nullable `nil`, non-viable boolean, enum, and mutated-container values with `OpenAI::Errors::ConversionError` instead of violating generated reader types.
- Verify the original customer-facing `message.parsed` workflow from #126/#309 and the Responses `content.parsed` workflow through real public client calls.
- Add a cross-boundary contract matrix that checks generated RBIs against actual construction, assignment, and parsed-response behavior.

## Why existing tests did not catch this

1. #295 had already enabled global nested coercion when #364 introduced the Tapioca compiler, so concrete nested signatures matched the then-current behavior.
2. The compiler regression generated RBI text and ran Sorbet against a synthetic parsed value, but never instantiated the models or compared runtime reader values to those signatures.
3. #375 reverted global coercion and added strong raw-value compatibility tests for generated/internal request models and parsed API models, but did not exercise application-defined structured-output models or the Tapioca compiler contract.
4. Existing structured-output streaming tests used scalar-only models, so neither nested readers nor nested array elements exposed the mismatch.
5. The primitive converter matrix tested literal Symbol constants but never tested `Symbol` as a class target, even though structured-output schemas explicitly advertise `Symbol`; successive container-specific fixes therefore masked the missing canonical scalar behavior.

The new compiler test closes the gap by checking every advertised field shape against runtime readers for direct construction, explicit assignment, and parsed responses in the same subprocess that generates and typechecks the RBI. It additionally checks raw identity through both accessors, nullable array elements and parsed array identity, direct/nested Symbol arrays and unions, including Symbol-bearing models inside arrays and unions, enum conversion, both union branches, nested model inheritance, caller-owned mutations, and conversion errors for invalid booleans, enums, and nested container values.

A 108-assertion scalar/container matrix systematically combines every supported scalar (`String`, `Symbol`, `Integer`, `Float`, boolean, enum) with nested model, array, and union boundaries across construction, assignment, and parsed conversion. Public-client integration tests exercise both `chat.completions.create(response_format: Model)` and `responses.create(text: Model)` using Minitest's stub on the SDK's actual HTTP-client protocol. Primitive-converter regressions separately verify valid Symbols, JSON strings, invalid input, and exactness/error accounting. A separate nine-shape nullable matrix verifies explicitly constructed, assigned, and parsed `nil` across String, Symbol, Integer, Float, Boolean, model, array, enum, and union readers; the generated-RBI subprocess independently checks nullable constructor and assignment behavior.

## Design tradeoff

Globally restoring nested coercion would regress the published generated-request model contract. Widening generated readers to model/hash unions would make signatures truthful but would also require narrowing every nested field in normal parsed-response code. Documenting the signatures as parsed-only would leave direct construction and assignment unsound.

Instead, the structured-output subclass prepends one small reader module per application-defined model. Each wrapper delegates to the original generated reader for existing validation, returns already-materialized values unchanged, and uses the existing canonical converter only when the raw value does not already match its declared target. Raw storage remains unchanged, and generated SDK request models never enter this path.

`Symbol` conversion belongs alongside the converter's existing `String`, `Integer`, and `Float` primitive branches, not in separate model/array/union pre-processing hooks. Fixing that canonical scalar branch deletes 41 lines of workaround logic and automatically handles every recursive composition. An audit of all 2,308 generated/internal model classes found zero direct `Symbol` declarations, and the generated-request raw-value compatibility suite remains unchanged and green. This changes scalar parsing only; it does not restore global nested-model coercion.

Setters are intentionally not added to generated RBIs: the documented compiler contract promises readers.

## Castiron ownership

No upstream Castiron compiler or renderer change is required.

The SDK contributor guide explicitly says the generator never modifies `lib/openai/helpers/`. Castiron's authoritative `generated-file-excludes.yaml` independently excludes `README.md` globally, both `lib/openai/helpers/` and `lib/openai/internal/` for Ruby, and the handwritten `test/` tree except generated resource tests and namespace scaffolding. I evaluated that policy against every changed PR path:

```text
EXCLUDED README.md
EXCLUDED lib/openai/helpers/structured_output/base_model.rb
EXCLUDED lib/openai/helpers/structured_output/array_of.rb
EXCLUDED lib/openai/helpers/structured_output/union_of.rb
EXCLUDED lib/openai/internal/type/converter.rb
EXCLUDED test/openai/helpers/structured_output_test.rb
EXCLUDED test/openai/internal/type/base_model_test.rb
EXCLUDED test/openai/tapioca/base_model_compiler_test.rb
```

Castiron's Ruby renderer generates package/client/model/resource/signature files and resource tests; it does not emit the handwritten structured-output helper or these tests. The existing Tapioca compiler file is unchanged.

## Verification

- `mise exec ruby@4.0.6 -- env TEST_API_BASE_URL=http://127.0.0.1:4520 ./scripts/test` — 648 runs, 2,813 assertions.
- `mise exec ruby@3.4.10 -- env TEST_API_BASE_URL=http://127.0.0.1:4520 ./scripts/test` — 648 runs, 2,813 assertions.
- `mise exec ruby@3.3.12 -- env TEST_API_BASE_URL=http://127.0.0.1:4520 ./scripts/test` — 648 runs, 2,813 assertions.
- `mise exec ruby@4.0.6 -- bundle exec rake lint` on current main — 1,389 RuboCop files clean, Sorbet clean, and 1,212 RBS files valid.
- Expanded Tapioca compiler/runtime contract regression passes on Ruby 3.3, 3.4, and 4.0.
- Existing raw-value compatibility suite remains green.
- `mise exec ruby@4.0.6 -- bundle exec rake build:gem`; package contains the handwritten structured-output runtime and compiler with no `sorbet-runtime` runtime dependency.
- Required thermo-nuclear code-quality review repeated with no findings.

### DIFF
--- a/lib/openai/helpers/structured_output/array_of.rb
+++ b/lib/openai/helpers/structured_output/array_of.rb
@@ -23,19 +23,6 @@ module OpenAI
 
         # @api private
         #
-        # @param value [Object]
-        # @param state [Hash{Symbol=>Object}]
-        # @return [Object]
-        def coerce(value, state:)
-          if item_type.equal?(Symbol) && value.is_a?(Array)
-            value = value.map { _1.is_a?(String) ? _1.to_sym : _1 }
-          end
-
-          super
-        end
-
-        # @api private
-        #
         # @param state [Hash{Symbol=>Object}]
         #
         #   @option state [Hash{Object=>String}] :defs
--- a/lib/openai/helpers/structured_output/array_of.rb
+++ b/lib/openai/helpers/structured_output/array_of.rb
@@ -13,6 +13,14 @@ module OpenAI
       class ArrayOf < OpenAI::Internal::Type::ArrayOf
         include OpenAI::Helpers::StructuredOutput::JsonSchemaConverter
 
+        # @api public
+        #
+        # @param other [Object]
+        # @return [Boolean]
+        def ===(other)
+          super || (nilable? && other.is_a?(Array) && other.compact.all?(item_type))
+        end
+
         # @api private
         #
         # @param state [Hash{Symbol=>Object}]
--- a/lib/openai/helpers/structured_output/array_of.rb
+++ b/lib/openai/helpers/structured_output/array_of.rb
@@ -23,6 +23,19 @@ module OpenAI
 
         # @api private
         #
+        # @param value [Object]
+        # @param state [Hash{Symbol=>Object}]
+        # @return [Object]
+        def coerce(value, state:)
+          if item_type.equal?(Symbol) && value.is_a?(Array)
+            value = value.map { _1.is_a?(String) ? _1.to_sym : _1 }
+          end
+
+          super
+        end
+
+        # @api private
+        #
         # @param state [Hash{Symbol=>Object}]
         #
         #   @option state [Hash{Object=>String}] :defs
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -68,12 +68,12 @@ module OpenAI
             # Preserve the original reader's validation without replacing raw field storage.
             readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
             readers.define_method(name_sym) do
+              return nil if nilable && self[name_sym].nil?
+
               value = super()
               target = type.call
 
               case value
-              when nil
-                return nil if nilable
               when target
                 return value
               end
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -76,6 +76,8 @@ module OpenAI
                 return nil if nilable
               when target
                 return value
+              when String
+                return value.to_sym if target == Symbol
               end
 
               state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -59,28 +59,6 @@ module OpenAI
 
         class << self
           # @api private
-          #
-          # @param value [Object]
-          # @param state [Hash{Symbol=>Object}]
-          # @return [Object]
-          def coerce(value, state:)
-            if value.is_a?(Hash)
-              original = value
-              known_fields.each do |name, field|
-                next unless field.fetch(:type_fn).call.equal?(Symbol)
-
-                key = state.fetch(:translate_names) ? field.fetch(:api_name) : name
-                next unless (item = value[key]).is_a?(String)
-
-                value = value.dup if value.equal?(original)
-                value[key] = item.to_sym
-              end
-            end
-
-            super
-          end
-
-          # @api private
           def required(name_sym, type_info, spec = {})
             super
 
@@ -98,8 +76,6 @@ module OpenAI
                 return nil if nilable
               when target
                 return value
-              when String
-                return value.to_sym if target == Symbol
               end
 
               state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -62,7 +62,9 @@ module OpenAI
           def required(name_sym, type_info, spec = {})
             super
 
-            type = known_fields.fetch(name_sym).fetch(:type_fn)
+            field = known_fields.fetch(name_sym)
+            type = field.fetch(:type_fn)
+            nilable = field.fetch(:nilable)
             # Preserve the original reader's validation without replacing raw field storage.
             readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
             readers.define_method(name_sym) do
@@ -70,25 +72,27 @@ module OpenAI
               target = type.call
 
               case value
-              when nil, target
-                value
-              else
-                state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
-                converted = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
-
-                case converted
-                when target
-                  return converted if state.fetch(:error).nil? && state.fetch(:exactness).fetch(:no).zero?
-                end
-
-                raise OpenAI::Errors::ConversionError.new(
-                  on: self.class,
-                  method: name_sym,
-                  target: target,
-                  value: value,
-                  cause: state.fetch(:error)
-                )
+              when nil
+                return nil if nilable
+              when target
+                return value
               end
+
+              state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+              converted = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
+
+              case converted
+              when target
+                return converted if state.fetch(:error).nil? && state.fetch(:exactness).fetch(:no).zero?
+              end
+
+              raise OpenAI::Errors::ConversionError.new(
+                on: self.class,
+                method: name_sym,
+                target: target,
+                value: value,
+                cause: state.fetch(:error)
+              )
             end
           end
 
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -59,6 +59,28 @@ module OpenAI
 
         class << self
           # @api private
+          #
+          # @param value [Object]
+          # @param state [Hash{Symbol=>Object}]
+          # @return [Object]
+          def coerce(value, state:)
+            if value.is_a?(Hash)
+              original = value
+              known_fields.each do |name, field|
+                next unless field.fetch(:type_fn).call.equal?(Symbol)
+
+                key = state.fetch(:translate_names) ? field.fetch(:api_name) : name
+                next unless (item = value[key]).is_a?(String)
+
+                value = value.dup if value.equal?(original)
+                value[key] = item.to_sym
+              end
+            end
+
+            super
+          end
+
+          # @api private
           def required(name_sym, type_info, spec = {})
             super
 
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -58,6 +58,7 @@ module OpenAI
         end
 
         class << self
+          # @api private
           def required(name_sym, type_info, spec = {})
             super
 
@@ -72,10 +73,20 @@ module OpenAI
               when nil, target
                 value
               else
-                OpenAI::Internal::Type::Converter.coerce(
-                  target,
-                  value,
-                  state: OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+                state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+                converted = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
+
+                case converted
+                when target
+                  return converted if state.fetch(:error).nil? && state.fetch(:exactness).fetch(:no).zero?
+                end
+
+                raise OpenAI::Errors::ConversionError.new(
+                  on: self.class,
+                  method: name_sym,
+                  target: target,
+                  value: value,
+                  cause: state.fetch(:error)
                 )
               end
             end
--- a/lib/openai/helpers/structured_output/base_model.rb
+++ b/lib/openai/helpers/structured_output/base_model.rb
@@ -58,6 +58,29 @@ module OpenAI
         end
 
         class << self
+          def required(name_sym, type_info, spec = {})
+            super
+
+            type = known_fields.fetch(name_sym).fetch(:type_fn)
+            # Preserve the original reader's validation without replacing raw field storage.
+            readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
+            readers.define_method(name_sym) do
+              value = super()
+              target = type.call
+
+              case value
+              when nil, target
+                value
+              else
+                OpenAI::Internal::Type::Converter.coerce(
+                  target,
+                  value,
+                  state: OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+                )
+              end
+            end
+          end
+
           def optional(...)
             message = "`optional` is not supported for structured output APIs, use `#required` with `nil?: true` instead"
             raise RuntimeError.new(message)
--- a/README.md
+++ b/README.md
@@ -934,7 +934,9 @@ The SDK includes a Tapioca DSL compiler for application-defined subclasses of
 `OpenAI::BaseModel`. When Tapioca loads your application, running
 `bundle exec tapioca dsl` generates typed readers for fields declared with
 `required`, including nested models, arrays, enums, unions, and fields declared
-with `nil?: true`.
+with `nil?: true`. Structured-output model readers materialize these values
+even when a model is constructed or assigned with hashes. Accessing a field
+with `[]` or `to_h` still returns the original caller-provided value.
 
 Response `parsed` fields can contain different application-defined models, so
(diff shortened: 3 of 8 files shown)


lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,29 @@ def to_json_schema_inner(state:)
...
+            readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
+            readers.define_method(name_sym) do
+              value = super()
+              target = type.call
+
+              case value
+              when nil, target
+                value
+              else
+                OpenAI::Internal::Type::Converter.coerce(
```
HAYDEN-OAI: The result of this best-effort conversion still is not guaranteed to match the generated reader type. `Converter.coerce` deliberately returns the original input when it cannot convert it, and some converters only update `state[:exactness]` rather than setting `state[:error]`. For example, for `required :active, OpenAI::Boolean`, `Event.new(active: "yes").active` still returns the String `"yes"`; likewise an unknown string passed to an enum reader remains a String even though its RBI returns `Symbol`. That leaves direct construction/assignment unsound, which is the contract this PR is intended to repair. Please retain the coercion state and turn a non-viable conversion into `ConversionError` instead of returning the unchanged value; the same validation needs to occur when a retained caller-owned container is mutated after assignment.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,29 @@ def to_json_schema_inner(state:)
...
+
+            type = known_fields.fetch(name_sym).fetch(:type_fn)
+            # Preserve the original reader's validation without replacing raw field storage.
+            readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
+            readers.define_method(name_sym) do
+              value = super()
+              target = type.call
+
+              case value
+              when nil, target
```
HAYDEN-OAI: `target === value` does not reliably recognize every already-materialized value. In particular, `ArrayOf#===` checks every element against `item_type` without honoring its `nilable?` flag. A valid parsed `OpenAI::ArrayOf[Participant, nil?: true]` containing `nil` therefore misses this branch and is rebuilt on every read. `parsed.participants` is no longer the same materialized array as `parsed[:participants]`, and mutations to the reader-returned array are lost, contrary to the stated parsed-value preservation. Please make the matcher honor nilable elements (and add the missing identity assertion) or use stored coercion state rather than `===` to decide whether to rematerialize.

HAYDEN-OAI (commented):
Reviewed `e01f430b4572b8eeebf5d64ac3625dd5f0a0b9c0`. I found two remaining reader-contract gaps in the new wrapper; details are inline. I also agree with the existing unresolved `@api private` documentation note.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,29 @@ def to_json_schema_inner(state:)
         end
 
         class << self
+          def required(name_sym, type_info, spec = {})
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in f6583664. Added the `@api private` annotation directly to the structured-output `required` override while keeping the method publicly callable and retaining its inherited public RBI/RBS visibility.

lib/openai/helpers/structured_output/base_model.rb
jbeckwith-oai: Fixed in f6583664. The reader now retains the coercion state, rejects explicit errors and non-viable exactness, and verifies the converted value matches the declared target before returning it. Invalid values raise `OpenAI::Errors::ConversionError`; new constructor, assignment, and caller-owned-container mutation regressions cover invalid booleans, unknown enum strings, and invalid nested elements.

lib/openai/helpers/structured_output/base_model.rb
jbeckwith-oai: Fixed in f6583664. The structured-output-specific `ArrayOf#===` now honors `nil?: true` while delegating to the existing matcher for ordinary arrays. Parsed nullable arrays therefore keep the same object through the reader and `[]`; the regression asserts identity before and after mutating the reader-returned array.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,40 @@ def to_json_schema_inner(state:)
...
+
+            type = known_fields.fetch(name_sym).fetch(:type_fn)
+            # Preserve the original reader's validation without replacing raw field storage.
+            readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
+            readers.define_method(name_sym) do
+              value = super()
+              target = type.call
+
+              case value
+              when nil, target
```
HAYDEN-OAI: The unconditional `nil` fast path still bypasses the new viability checks for required, non-nilable fields. For example, with `required :active, OpenAI::Boolean`, both `Event.new(active: nil).active` and `Event.new.active` return `nil`: the underlying Boolean converter records a non-viable exactness but yields `nil`, and this branch accepts it even though the generated RBI returns `T::Boolean`. The same issue applies to missing required nested-model/enum fields. Please capture this field's `nilable` metadata and only return `nil` when the declaration actually permits it; otherwise raise `ConversionError` so direct construction remains consistent with the advertised reader type.

HAYDEN-OAI (commented):
Re-reviewed the new head `2bebdfda62cdece88e476bb831f1028bd402b264`. The previous conversion-viability, nullable-array identity, and private-API findings are addressed. One remaining non-nil reader-contract gap is inline.

lib/openai/helpers/structured_output/base_model.rb
jbeckwith-oai: Fixed in ff1511fc. The reader now captures the field's `nilable` metadata and returns nil only for declarations that explicitly permit it. Missing or explicitly nil non-nullable booleans, enums, nested models, and arrays now raise `OpenAI::Errors::ConversionError`; new regressions cover explicit nil, missing required fields across those shapes, and valid nullable fields. I also flattened the reader control flow while preserving raw storage and parsed identity.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,40 @@ def to_json_schema_inner(state:)
...
+                value
+              else
+                state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+                converted = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
+
+                case converted
+                when target
+                  return converted if state.fetch(:error).nil? && state.fetch(:exactness).fetch(:no).zero?
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in bc0c6a16. `OpenAI::Helpers::StructuredOutput::UnionOf#coerce` now clears a stale error only after the selected value matches the union and the conversion has zero non-viable branches. A focused regression covers `UnionOf[String, EmptyUnionMember]` through direct construction, reassignment, and parsed responses while preserving raw request identity.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,40 @@ def to_json_schema_inner(state:)
...
+            readers.define_method(name_sym) do
+              value = super()
+              target = type.call
+
+              case value
+              when nil, target
+                value
+              else
+                state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+                converted = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in bc0c6a16. Structured-output readers now materialize JSON strings as symbols when the declared field type is `Symbol`, while preserving the original caller-owned string in `[]`/`to_h`. Added direct-construction, assignment, parsed-response, generated RBI, and Sorbet typecheck coverage for Symbol fields.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,46 @@ def to_json_schema_inner(state:)
...
+              value = super()
+              target = type.call
+
+              case value
+              when nil
+                return nil if nilable
+              when target
+                return value
+              when String
+                return value.to_sym if target == Symbol
```
HAYDEN-OAI: The new Symbol fast path only applies when the field target itself is Symbol. Symbol is also an advertised nested structured-output type: required :kinds, OpenAI::ArrayOf[Symbol] generates an array-of-string schema and a T::Array[Symbol] reader, but JSON ["ready"] is still delegated to the canonical item converter, which leaves the String unchanged and increments exactness[:no]; this reader then raises ConversionError. UnionOf[Symbol, Integer] has the same gap. Please normalize Symbol recursively within structured-output arrays and unions, and add those regressions, rather than only handling it at the top-level field reader.

HAYDEN-OAI (commented):
Re-reviewed current head bc0c6a1647ae5d7c868e3bbcca3e2eafae2c7ca1. The required-nil and stale-union-error fixes address their reported cases. One remaining recursive Symbol-normalization gap is inline.

lib/openai/helpers/structured_output/base_model.rb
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 9ad3490c. Structured-output `ArrayOf` and `UnionOf` now normalize JSON strings to `Symbol` recursively while leaving the shared generated-request converter unchanged. Regression coverage exercises direct construction, assignment, parsed responses, nullable and nested arrays, Symbol unions, arrays of unions, retained raw-value identity, generated RBI output, and Sorbet checking on Ruby 3.3/3.4/4.0. The full suite passes with 643 tests and 2,649 assertions.

jbeckwith-oai: Fixed in 9ad3490c. The structured-output `ArrayOf#coerce` now normalizes String elements when its item type is Symbol, and structured-output `UnionOf#coerce` normalizes Symbol branches without disturbing unions that explicitly accept String. The recursive converter consequently also handles nullable arrays, nested arrays, and arrays of Symbol unions. New regression tests cover direct construction, assignment, parsed API responses, raw object identity, the generated `T::Array[Symbol]`/`T.any(Integer, Symbol)` RBIs, and Sorbet checking on Ruby 3.3, 3.4, and 4.0. Shared generated-request coercion remains untouched; the full suite passes 643 tests and 2,649 assertions.

HAYDEN-OAI (commented):
Re-reviewed current head 9ad3490c5bd4c7e8f1b363a8a796f09acc55a52a. Direct Symbol arrays and unions are fixed, but one remaining model-boundary recursion gap is inline.

lib/openai/helpers/structured_output/array_of.rb
```diff
@@ -13,6 +13,27 @@ module StructuredOutput
...
+          super || (nilable? && other.is_a?(Array) && other.compact.all?(item_type))
+        end
+
+        # @api private
+        #
+        # @param value [Object]
+        # @param state [Hash{Symbol=>Object}]
+        # @return [Object]
+        def coerce(value, state:)
+          if item_type.equal?(Symbol) && value.is_a?(Array)
```
HAYDEN-OAI: This recursion still stops at a BaseModel boundary. For example, if Item has required :kind, Symbol, then Event.new(items: [{kind: "ready"}]) with items declared as OpenAI::ArrayOf[Item] still raises ConversionError: Item coercion sends the nested String through the canonical Symbol class converter, records exactness[:no], and this outer array reader rejects the otherwise materialized Item array. OpenAI::UnionOf[Integer, Item] is worse because the same non-viable count makes the union discard the valid model branch for parsed API responses too. Please make structured Symbol normalization participate in BaseModel field coercion, and cover models containing Symbol fields inside arrays and unions; walking only ArrayOf and UnionOf nodes is not fully recursive.

lib/openai/helpers/structured_output/array_of.rb
```diff
@@ -13,6 +13,27 @@ module StructuredOutput
...
+            value = value.map { _1.is_a?(String) ? _1.to_sym : _1 }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in ec0f2bef. Structured-output `BaseModel.coerce` now normalizes declared Symbol fields before delegating to the existing model converter, copying an input hash only when a Symbol actually needs conversion. That makes recursive model branches viable inside arrays and unions while preserving caller-owned raw hashes/arrays and leaving generated-request conversion unchanged. Added failing-then-passing construction, assignment, parsed-response, raw-identity, generated-RBI, and Sorbet regressions for models with Symbol fields nested in both containers. Ruby 3.3/3.4 compiler checks, full Ruby 4.0 suite (644 tests / 2,662 assertions), lint, Sorbet, RBS, and gem packaging pass.

lib/openai/helpers/structured_output/array_of.rb
jbeckwith-oai: Fixed in ec0f2bef. Structured-output `BaseModel.coerce` now normalizes declared Symbol fields before delegating to the existing model converter, copying an input hash only when a Symbol actually needs conversion. That makes recursive model branches viable inside arrays and unions while preserving caller-owned raw hashes/arrays and leaving generated-request conversion unchanged. Added failing-then-passing construction, assignment, parsed-response, raw-identity, generated-RBI, and Sorbet regressions for models with Symbol fields nested in both containers. Ruby 3.3/3.4 compiler checks, full Ruby 4.0 suite (644 tests / 2,662 assertions), lint, Sorbet, RBS, and gem packaging pass.

lib/openai/helpers/structured_output/base_model.rb
```diff
@@ -58,6 +58,44 @@ def to_json_schema_inner(state:)
...
+          def required(name_sym, type_info, spec = {})
+            super
+
+            field = known_fields.fetch(name_sym)
+            type = field.fetch(:type_fn)
+            nilable = field.fetch(:nilable)
+            # Preserve the original reader's validation without replacing raw field storage.
+            readers = @structured_output_readers ||= Module.new.tap { prepend(_1) }
+            readers.define_method(name_sym) do
+              value = super()
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d5277328. The structured-output reader now checks its declared nilability and raw stored value before calling the inherited reader, so a cached converter error cannot reject a valid explicit nil. Added a nine-shape construction/assignment/parsed-response matrix covering String, Symbol, Integer, Float, Boolean, nested models, arrays, enums, and unions, plus failing-then-passing constructor/assignment checks inside the real generated-RBI/Sorbet subprocess. Non-nullable nil still raises; raw storage stays unchanged. Ruby 3.3/3.4/4.0 focused coverage, full Ruby 4.0 suite (648 tests / 2,813 assertions), RuboCop, Sorbet, RBS, and packaging all pass.

jbeckwith-oai:
Closing after architecture review. OpenAI::BaseModel is an SDK transport model with permissive construction and raw-storage compatibility; retrofitting it into a strongly typed modeling framework creates conflicting reader/storage semantics and unnecessary coercion complexity. The customer need is useful, but the proposed abstraction is not. A future design should evaluate purpose-built Ruby typed models such as Sorbet T::Struct and derive schemas/hydration at the integration boundary rather than extending this PR.

PR #392, opened 2026-08-13
fix: validate and bound retry delays

jbeckwith-oai:
## Summary

- reject negative, expired, malformed, and non-finite retry-after hints; preserve valid retry-after-ms precedence and explicit zero-delay hints
- cap server-requested delays at the configured maximum and use jittered exponential backoff starting at the configured initial delay
- add focused direct-policy and end-to-end HTTP 429 regression coverage, including custom retry settings, real SDK HTTP transport mocks, and retry callback events

## Castiron ownership

No upstream Castiron, generator, schema, template, or regeneration change is required. The canonical Ruby generation ownership policy excludes the entire `lib/openai/internal/` runtime and non-resource `test/` trees; its `ruby_policy_keeps_generated_api_surface_and_preserves_sdk_documentation` regression test explicitly asserts that `lib/openai/internal/transport/base_client.rb` is excluded. The Ruby renderer does not emit this transport file, and the Castiron SDK preparation overlay restores handwritten paths from the integrated SDK tree. Both modified runtime logic and new internal tests are therefore SDK-owned and preserved across regeneration.

## Validation

- Ruby 3.3.12: 639 tests, 2,565 assertions, no failures
- Ruby 3.4.10: 639 tests, 2,565 assertions, no failures
- Ruby 4.0.6: 639 tests, 2,565 assertions, no failures
- full RuboCop: 1,390 files, no offenses
- Sorbet: no errors
- RBS validation: 1,212 files, no errors
- independently reproduced `429 retry-after: -1` raising `ArgumentError` before the fix and confirmed successful retry afterward
- required thermo-nuclear code-quality review completed with no structural findings

### DIFF
--- a/lib/openai/internal/transport/base_client.rb
+++ b/lib/openai/internal/transport/base_client.rb
@@ -432,23 +432,22 @@ module OpenAI
         #
         # @return [Float]
         private def retry_delay(headers, retry_count:)
-          # Non-standard extension
-          span = Float(headers["retry-after-ms"], exception: false)&.then { _1 / 1000 }
-          return span if span
-
           retry_header = headers["retry-after"]
-          return span if (span = Float(retry_header, exception: false))
+          delays = [
+            Float(headers["retry-after-ms"], exception: false)&.then { _1 / 1000 },
+            Float(retry_header, exception: false),
+            retry_header&.then do
+              Time.httpdate(_1) - Time.now
+            rescue ArgumentError
+              nil
+            end
+          ]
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Reviewed the retry-delay parsing, precedence, finite/non-negative validation, server-delay cap, and exponential backoff/jitter behavior. The implementation safely falls back for malformed, expired, negative, and non-finite hints while preserving valid zero-delay and millisecond precedence.

PR #393, opened 2026-08-13
chore: enforce missing super lint

jbeckwith-oai:
## Summary
- Remove the RBI exclusion for Lint/MissingSuper.
- Enforce the rule across every repository-owned Ruby and RBI file.

## Validation
- Targeted RuboCop: 1,389 files, zero offenses.

## Stack
1 of 5. Base: main.

## Generator impact
No Castiron change is needed. The RuboCop config and RBI paths involved here are generator-excluded.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -82,10 +82,6 @@ Lint/EmptyInPattern:
   Exclude:
     - "test/**/*"
 
-Lint/MissingSuper:
-  Exclude:
-    - "**/*.rbi"
-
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Looks good. This narrowly removes the RBI exception while preserving the repository’s broader generated-RBI exclusions, so `Lint/MissingSuper` now covers the intended repository-owned Ruby/RBI surface without changing runtime behavior. The standalone base is coherent and does not depend on later patches in the lint stack.

PR #394, opened 2026-08-13
chore: enforce useless assignment lint

jbeckwith-oai:
## Summary
- Remove the examples exclusion for Lint/UselessAssignment while keeping unsafe autocorrection disabled.
- Mark the intentionally unused Sorbet demonstration value explicitly and keep its commented example consistent.

## Validation
- Targeted RuboCop: 1,389 files, zero offenses.
- Sorbet typecheck for the example passes.

## Stack
2 of 5. Depends on the missing-super lint PR.

## Generator impact
No Castiron change is needed. The RuboCop config and examples directory are generator-excluded.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -82,10 +82,6 @@ Lint/EmptyInPattern:
   Exclude:
     - "test/**/*"
 
-Lint/MissingSuper:
-  Exclude:
-    - "**/*.rbi"
-
 # Disabled for safety reasons, this option changes code semantics.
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Looks good. The atomic delta removes only the examples exemption while keeping unsafe autocorrection disabled, and the sole intentional assignment is made explicitly unused with its opt-in demonstration kept consistent. This composes cleanly with the inherited missing-super ratchet and does not change runtime behavior.

PR #395, opened 2026-08-13
chore: enforce boolean symbol lint

jbeckwith-oai:
## Summary
- Enable Lint/BooleanSymbol repository-wide.
- Preserve the intentional boolean-like symbol coercion edge case without a forbidden symbol literal.

## Validation
- Targeted RuboCop passes for BooleanSymbol, SymbolConversion, and LiteralInInterpolation.
- The affected test passes: 17 runs, 261 assertions.

## Stack
3 of 5. Depends on the useless-assignment lint PR.

## Generator impact
No Castiron change is needed. The RuboCop config and internal type test are generator-excluded.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -82,10 +82,6 @@ Lint/EmptyInPattern:
   Exclude:
     - "test/**/*"
 
-Lint/MissingSuper:
-  Exclude:
-    - "**/*.rbi"
-
 # Disabled for safety reasons, this option changes code semantics.
 Lint/UnusedMethodArgument:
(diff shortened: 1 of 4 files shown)


HAYDEN-OAI (approved):
The atomic delta cleanly enables `Lint/BooleanSymbol` without weakening the coercion regression: `true.to_s.to_sym` still supplies the exact `:true` value under test while avoiding a forbidden literal. The inherited useless-assignment base is already approved and introduces no conflict with this change. I found no substantive issues.

PR #398, opened 2026-08-13
fix: honor workload identity environment defaults

jbeckwith-oai:
## Summary
- Honor the documented IDENTITY_PROVIDER_ID and SERVICE_ACCOUNT_ID environment defaults while keeping explicit identifier precedence.
- Reject missing, nil, empty, and whitespace-only identifiers with actionable setup errors.
- Keep the Sorbet constructor signature aligned and cover isolated environment state, symbols, provider, client_id, and refresh-buffer compatibility.

## Generator ownership
Castiron explicitly excludes lib/openai/auth/, rbi/openai/auth.rbi, and non-resource tests from generated ownership in api/castiron/docs/stainless-sdk-json-fixtures/release-primary/generated-file-excludes/ruby.yaml. These are SDK-authored files preserved across generation; https://github.com/openai/openai/pull/1290530 adds companion regression coverage for the ownership boundary and three-way overlay preservation. No renderer, template, schema, or production generator change is required.

## Verification
- Added regression tests first and observed four failures plus one error on unmodified production code.
- Focused workload-identity suite: 20 tests, 85 assertions.
- Full SDK suite: 632 tests, 2,582 assertions.
- Full RuboCop: 1,389 files, no offenses.
- Sorbet typecheck: clean.
- RBS validation: 1,212 files, no errors.
- Completed the requested thermo-nuclear code-quality review and aligned RBI environment defaults with the existing client signature convention.

### DIFF
--- a/lib/openai/auth/workload_identity.rb
+++ b/lib/openai/auth/workload_identity.rb
@@ -6,12 +6,22 @@ module OpenAI
       attr_reader :client_id, :identity_provider_id, :service_account_id, :provider, :refresh_buffer_seconds
 
       def initialize(
-        identity_provider_id:,
-        service_account_id:,
         provider:,
+        identity_provider_id: ENV["IDENTITY_PROVIDER_ID"],
+        service_account_id: ENV["SERVICE_ACCOUNT_ID"],
         client_id: nil,
         refresh_buffer_seconds: 1200
       )
+        if identity_provider_id.to_s.strip.empty?
+          raise ArgumentError,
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Looks good. Environment defaults are evaluated at construction time, explicit identifiers take precedence (including explicit blank/nil values being rejected instead of silently falling back), symbols retain the existing string coercion, and validation prevents invalid identifiers from reaching token exchange. Provider/client-id/refresh-buffer behavior remains unchanged. The RBI correctly models the newly optional keyword values as nilable String/Symbol while retaining a required typed provider, and the test environment is isolated and restored.

PR #399, opened 2026-08-13
fix: reject dot-only identifiers in API paths

jbeckwith-oai:
## Summary
- Reject standalone `.` and `..` identifiers in the shared path encoder before URI joining can normalize an authenticated request onto another endpoint.
- Add red-to-green coverage for every path-argument position, nested group/user admin role deletes, dotted and numeric identifiers, percent-looking strings, and escaped slashes.
- Ownership: active Castiron excludes `lib/openai/internal/**` from generated output and preserves this SDK-owned runtime through its bootstrap overlay. A companion Castiron PR will add explicit generator ownership/preservation regression coverage and will be linked here.

## Security reproduction
Before this change, `client.admin.organization.projects.groups.roles.delete("role_123", project_id: "proj_123", group_id: "..")` sends `DELETE /v1/projects/proj_123/roles/role_123`, the distinct project-role delete endpoint. Both standalone dot identifiers now raise `ArgumentError` before transport execution.

## Verification
- Ruby 3.3, 3.4, and 4.0 focused path-security regressions: 5 tests / 41 assertions each.
- Full Ruby 4.0 suite: 631 tests / 2,576 assertions.
- RuboCop: 1,390 files, no offenses.
- Sorbet: no errors.
- RBS validation: 1,212 files, no errors.
- Gem packaging: succeeds.
- Thermo-nuclear code-quality review completed before pushing; no findings.

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -261,9 +261,13 @@ module OpenAI
         #
         # @param path [String, Integer]
         #
+        # @raise [ArgumentError]
         # @return [String]
         def encode_path(path)
-          path.to_s.gsub(OpenAI::Internal::Util::RFC_3986_NOT_PCHARS) { ERB::Util.url_encode(_1) }
+          path = path.to_s
+          raise ArgumentError, "path segment cannot be '.' or '..'" if path == "." || path == ".."
+
+          path.gsub(OpenAI::Internal::Util::RFC_3986_NOT_PCHARS) { ERB::Util.url_encode(_1) }
         end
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed the shared path encoding and URL-join boundary, including dot segments in every interpolation position, slash escaping, percent/double-encoded inputs, dotted identifiers, and both nested admin group/user role delete routes. Rejecting exact dot segments before encoding closes the URI normalization crossover while preserving legitimate segment values.

PR #400, opened 2026-08-13
fix: preserve existing BaseModel instances during coercion

jbeckwith-oai:
## Summary\n\n- preserve existing BaseModel and subclass instances during response coercion by checking the actual target model class instead of Class\n- preserve response metadata and request IDs, correctly reject Class objects, and retain exactness/union/nested-model behavior\n- add five regressions that fail before the one-token runtime fix; preserve the separately pinned raw nested-constructor contract\n\n## Generator ownership\n\nNo Castiron template, schema, generated model, or regeneration change is required for the runtime fix. The active generator at openai/openai:api/castiron explicitly excludes lib/openai/internal/, its internal RBI/RBS signatures, and handwritten non-resource tests; generated lib/openai/internal.rb only requires this SDK-owned BaseModel implementation. A companion active-monorepo PR is being prepared to lock these Ruby ownership boundaries with generator regression coverage and will be linked here.\n\n## Verification\n\n- Ruby 3.3.12: 631 runs, 2,560 assertions, zero failures/errors\n- Ruby 3.4.10: 631 runs, 2,560 assertions, zero failures/errors\n- Ruby 4.0.6: 631 runs, 2,560 assertions, zero failures/errors\n- focused BaseModel suite: 22 runs, 286 assertions\n- raw nested-value compatibility suite: 9 runs, 39 assertions\n- bundle exec rake lint: 1,389 RuboCop files clean, Sorbet clean, 1,212 RBS files valid\n- mandatory thermo-nuclear code-quality review: no findings

### DIFF
--- a/lib/openai/internal/type/base_model.rb
+++ b/lib/openai/internal/type/base_model.rb
@@ -273,7 +273,7 @@ module OpenAI
           def coerce(value, state:)
             exactness = state.fetch(:exactness)
 
-            if value.is_a?(self.class)
+            if value.is_a?(self)
               exactness[:yes] += 1
               return value
             end
--- a/test/openai/internal/type/base_model_test.rb
+++ b/test/openai/internal/type/base_model_test.rb
@@ -388,6 +388,54 @@ class OpenAI::Test::BaseModelTest < Minitest::Test
     end
   end
 
+  def test_coerce_preserves_model_identity_and_response_metadata
+    response = OpenAI::ResponseMetadata.new(status: 200, headers: {"x-request-id" => "req_123"})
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (approved):
The target of this converter is the BaseModel subclass itself, so `is_a?(self)` correctly implements the existing-instance fast path (including subclasses); `self.class` was testing against `Class` and admitted class objects instead. Returning the original instance preserves response metadata/request IDs, while non-model class objects now take the normal failed hash-coercion path. The nested array/model and discriminated-union paths retain identity and exactness as expected. I found no substantive issues.

PR #401, opened 2026-08-13
fix(webhooks): support Rack and case-insensitive HTTP headers

jbeckwith-oai:
## Summary

- Accept Rack/Sinatra HTTP_WEBHOOK_* environment keys, conventional title-case HTTP headers, and existing lowercase or symbol webhook headers.
- Reject conflicting aliases for the same signed header while accepting equivalent aliases with matching values.
- Correct both Sinatra README examples to explicitly fetch their API key and webhook secret, preserving the existing requirement for API credentials.
- Add genuine signed Rack request-environment coverage, title-case and symbol compatibility checks, duplicate-header security regressions, and README example assertions.

## Generator ownership

The webhook resource and original README examples were generated by Stainless. The required upstream Castiron Ruby-renderer companion is https://github.com/openai/openai/pull/1290513. It emits matching normalization and signed-request regression coverage only for the production OpenAI codegen profile while preserving existing Stainless-parity fixtures.

## Validation

- Full Ruby 4.0 SDK suite after rebasing onto current main: 646 tests, 2,590 assertions, zero failures.
- Complete RuboCop, Sorbet, and RBS validation: 2,615 Ruby/RBI files and 1,212 RBS files passed.
- Focused webhook regressions: 14 tests and 32 assertions passed on Ruby 3.3, 3.4, and 4.0.
- Gem packaging completed successfully.
- Companion Castiron Ruby renderer suite: 573 tests passed.
- Regression tests were observed failing before each implementation change and passing afterward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,8 +423,11 @@ Note that the `body` parameter must be the raw JSON string sent from the server 
 require 'sinatra'
 require 'openai'
 
-# Set up the client with webhook secret from environment variable
-client = OpenAI::Client.new(webhook_secret: ENV['OPENAI_WEBHOOK_SECRET'])
+# Set up the client with API credentials and the webhook secret
+client = OpenAI::Client.new(
+  api_key: ENV.fetch('OPENAI_API_KEY'),
+  webhook_secret: ENV.fetch('OPENAI_WEBHOOK_SECRET')
+)
 
 post '/webhook' do
   request_body = request.body.read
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Reviewed Rack/Sinatra environment normalization, title/lower/snake-case string and symbol aliases, conflicting versus equivalent aliases, signed payload construction, and the README credential setup. The companion Castiron change generates the same normalization for the OpenAI profile while preserving Stainless-parity output.

PR #402, opened 2026-08-13
fix: keep required path parameters out of resource query strings

jbeckwith-oai:
## Summary

- remove required path parameters before encoding queries in all eight affected generated Ruby resource methods
- preserve valid filters, array query parameters, request-level extra query options, and pagination follow-up requests
- add ten real HTTP/WebMock regression cases spanning conversations, evals, vector stores, beta thread run steps, and project groups/group roles/user roles

### Generator ownership

These resources are generated by the active Castiron Ruby renderer in `openai/openai` under `api/castiron`, not by the archived standalone `openai/castiron` repository. A companion generator PR fixes the source for `CodegenProfile::OpenAI` while preserving required Stainless-parity fixtures: https://github.com/openai/openai/pull/1290567

## Validation

- Command: `bundle exec ruby -Itest test/openai/path_parameter_query_test.rb` — 10 regression cases passed; the original eight endpoint cases were independently observed failing before the fix.
- Command: `./scripts/test` — 636 tests and 2,545 assertions passed against the Steady OpenAPI mock server.
- Command: `bundle exec rake lint:rubocop` — 1,390 files inspected with no offenses.
- Command: `bundle exec rake typecheck` — Sorbet passed and 1,212 RBS files validated.
- Command: `bundle exec rake build:gem` — the Ruby gem built successfully.

### DIFF
--- a/lib/openai/resources/admin/organization/projects/groups.rb
+++ b/lib/openai/resources/admin/organization/projects/groups.rb
@@ -53,11 +53,11 @@ module OpenAI
             # @see OpenAI::Models::Admin::Organization::Projects::GroupRetrieveParams
             def retrieve(group_id, params)
               parsed, options = OpenAI::Admin::Organization::Projects::GroupRetrieveParams.dump_request(params)
-              query = OpenAI::Internal::Util.encode_query_params(parsed)
               project_id =
                 parsed.delete(:project_id) do
                   raise ArgumentError.new("missing required path argument #{_1}")
                 end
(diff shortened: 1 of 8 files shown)


HAYDEN-OAI (approved):
Reviewed all eight generated method fixes and the companion renderer change. Required dumped path parameters are now deleted before sliced or unsliced query serialization, while legitimate filters, array values, request-level extra query options, empty-query behavior, and pagination follow-ups remain intact. The generator change is limited to the OpenAI profile and explicitly preserves Stainless-parity ordering.

PR #410, opened 2026-08-13
test: cover issued credentials in fail-closed debug logs

jbeckwith-oai:
## Summary

- Add end-to-end regression coverage proving SDK DEBUG logs never expose newly issued credentials returned under generic JSON `value` fields.
- Cover organization admin API keys, project service-account API keys, nested service-account creation keys, Realtime client secrets, and translation client secrets.
- Exercise percent-encoded issuer paths, optional trailing slashes, and body-preserving HTTP 307/308 redirects.
- Verify credentials and ordinary response values remain available to SDK callers while logs contain only safe structural JSON summaries.

## Why this PR is test-only

`main` now includes #411, which fixes this vulnerability more broadly by making DEBUG body logging fail closed: JSON and form contents are never emitted. The previous version of this PR implemented endpoint-specific redaction, but retaining that narrower mechanism would add unnecessary production complexity and conflict with the stronger upstream architecture.

The rebased PR therefore preserves the newly merged runtime, README, RBI, and RBS behavior unchanged and adds only focused regression tests for the original credential-issuance vulnerability and prior review findings.

## Verification

- Full mock-backed suites on Ruby 3.3.12, 3.4.10, and 4.0.6: **723 tests / 3,350 assertions per version**, zero failures, errors, or skips.
- Focused logging-security suite on all three Ruby versions: **14 tests / 451 assertions per version**.
- RuboCop: **2,621 files**, no offenses.
- Sorbet typecheck: clean.
- RBS validation: **1,212 files**, no errors.
- Gem packaging succeeds as part of each full test run.

## Castiron impact

None. This change adds SDK-owned regression tests only; no generator, generated resource, transport, schema, runtime, or type-signature changes are required.

### DIFF
--- a/test/openai/logging_security_test.rb
+++ b/test/openai/logging_security_test.rb
@@ -62,6 +62,75 @@ class LoggingSecurityTest < Minitest::Test
     end
   end
 
+  def test_debug_logs_never_expose_newly_issued_credential_values
+    issuers = {
+      "organization/admin_api_keys" => "sk-admin-demo-release-audit",
+      "organization/projects/proj_123/service_accounts/acct_123/api_keys" => "service-account-credential",
+      "realtime/client_secrets" => "ek_demo-release-audit",
+      "realtime/translations/client_secrets" => "translation-client-credential",
+      "organization/admin%5fapi%5Fkeys/" => "percent-encoded-admin-credential",
+      "realtime/%63lient%5Fsecrets/" => "percent-encoded-realtime-credential"
+    }
+
+    issuers.each do |path, value|
+      response_body = {"value" => value, "details" => {"value" => "ordinary-nested-value"}}
+      output, response = logged_response(path: path, response_body: response_body)
+
+      assert_equal(value, response[:value])
+      assert_equal("ordinary-nested-value", response[:details][:value])
+      assert_includes(
+        output,
+        "[JSON BODY] bytes=#{JSON.generate(response_body).bytesize} type=object fields=2"
+      )
+      refute_includes(output, value)
+      refute_includes(output, "ordinary-nested-value")
+      refute_includes(output, "ordinary-request-value")
+    end
+  end
+
+  def test_debug_logs_never_expose_credentials_after_body_preserving_redirects
+    redirects = {
+      307 => "/v1/realtime/client%5Fsecrets",
+      308 => "/v1/organization/projects/proj_123/service_accounts/acct_123/api%5Fkeys"
+    }
+
+    redirects.each do |status, location|
+      value = "redirected-credential-#{status}"
+      output, response = logged_response(
+        path: "legacy/issue",
+        response_body: {"value" => value},
+        redirect: {status: status, location: location}
+      )
+
+      assert_equal(value, response[:value])
+      assert_includes(output, "redirect=1 method=POST url=https://example.com#{location}")
+      assert_includes(output, "type=object fields=1")
+      refute_includes(output, value)
+    end
+  end
+
+  def test_debug_logs_never_expose_nested_service_account_api_key_values
+    value = "service-account-creation-credential"
+    output, response = logged_response(
+      path: "organization/projects/proj_123/service_accounts",
+      response_body: {"id" => "acct_123", "api_key" => {"value" => value}}
+    )
+
+    assert_equal(value, response[:api_key][:value])
+    assert_includes(output, "type=object fields=2")
+    refute_includes(output, value)
+    refute_includes(output, "api_key")
+  end
+
+  def test_debug_logs_summarize_ordinary_response_values_without_changing_payloads
+    value = "ordinary-visible-value"
+    output, response = logged_response(path: "responses", response_body: {"value" => value})
+
+    assert_equal(value, response[:value])
+    assert_includes(output, "type=object fields=1")
+    refute_includes(output, value)
+  end
+
   def test_debug_logs_summarize_form_urlencoded_request_bodies_without_fields_or_values
     body =
       "client_secret=client-form-secret&access_token=access-form-secret&" \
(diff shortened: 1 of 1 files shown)


lib/openai/internal/logging.rb
```diff
@@ -152,6 +153,10 @@ def response_body(body, headers:, complete:, total_bytes:, attempt:)
 
         private def safe_field(value) = OpenAI::Internal::Logging.safe_field(value)
 
+        private def credential_issuing_request?
+          @method == "POST" && @url.path.match?(%r{/(?:admin_api_keys|api_keys|client_secrets)\z})
```
HAYDEN-OAI: **[P1] Classify the effective request after redirects**

`@method` and `@url` belong to the original logical request, but `BaseClient#send_request` follows 307/308 redirects while reusing this `Logging::Context`. A POST to a custom or legacy issuer path that redirects to `/realtime/client_secrets` (or an `/api_keys` endpoint) therefore leaves this predicate false and logs the newly issued top-level `value` in plaintext. Classify the actual method and URL recorded by `request_started`, or thread the effective request into response redaction.

HAYDEN-OAI (commented):
Found one credential-redaction gap in the supported redirect flow; see the inline comment for the affected path and fix.

lib/openai/internal/logging.rb
jbeckwith-oai: Fixed in 309190f9. `Logging::Context#request_started` now records the actual prepared request for each attempt, and credential redaction classifies that effective request's method and URL instead of the original logical request. The original method/URL remain unchanged for request-completion summaries.

Added end-to-end regression coverage for 307 and 308 redirects into Realtime/service-account credential issuers, redirects away from an issuer, and 303 redirects that convert POST to GET. The focused security suite passes 15 tests / 377 assertions on Ruby 3.3, 3.4, and 4.0; all three complete suites pass 704 tests / 3,214 assertions.

lib/openai/internal/logging.rb
```diff
@@ -152,6 +154,11 @@ def response_body(body, headers:, complete:, total_bytes:, attempt:)
 
         private def safe_field(value) = OpenAI::Internal::Logging.safe_field(value)
 
+        private def credential_issuing_request?
+          @request.method.to_s.upcase == "POST" &&
+            @request.url.path.match?(%r{/(?:admin_api_keys|api_keys|client_secrets)\z})
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 775815cd. The issuer-path matcher now accepts one optional trailing slash, and the focused end-to-end matrix covers trailing-slash organization admin keys, project service-account keys, Realtime client secrets, translation client secrets, and a 307 redirect into `/v1/realtime/client_secrets/`. The original returned credentials remain unchanged while their logged top-level `value` is redacted.

lib/openai/internal/logging.rb
```diff
@@ -38,6 +38,7 @@ def initialize(logger:, log_level:, on_retry:, method:, url:)
         end
 
         def request_started(request, redirect_count:)
+          @request = request
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 775815cd. `request_started` now computes `credential_issuing_request?(request)` immediately and retains only the resulting Boolean; the complete prepared request, its headers, URL, and encoded body are no longer kept by the long-lived logging context. Added an `ObjectSpace.reachable_objects_from` regression proving neither a large prepared request nor its body is retained. Redirect and POST-to-GET behavior remain covered.

lib/openai/internal/logging.rb
```diff
@@ -152,6 +154,11 @@ def response_body(body, headers:, complete:, total_bytes:, attempt:)
 
         private def safe_field(value) = OpenAI::Internal::Logging.safe_field(value)
 
+        private def credential_issuing_request?(request)
+          request.method.to_s.upcase == "POST" &&
+            request.url.path.match?(%r{/(?:admin_api_keys|api_keys|client_secrets)/?\z})
```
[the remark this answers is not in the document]
jbeckwith-oai: Verified against the actual service-account creation response shape: this path was already safe because `SENSITIVE_BODY_KEY` matches the parent `api_key` field and `scrub_value` replaces the entire nested object with `[REDACTED]` before descending to `value`.

Added an explicit end-to-end regression in 0406c47d for POST `/organization/projects/proj_123/service_accounts` returning `{"api_key":{"value":"service-account-creation-credential"}}`. It proves the SDK still returns the original key, the log contains `"api_key":"[REDACTED]"`, no credential appears anywhere in the log, and an unrelated nested `details.value` remains visible. No redundant endpoint-specific production special case is needed.

HAYDEN-OAI: **[P2] Normalize percent-encoded issuer paths before classification**

A supported custom POST or 307/308 redirect to `/v1/realtime/client%5Fsecrets` does not match this raw-path regex. `%5F` is the RFC-equivalent encoding of `_`, and ordinary HTTP servers/proxies decode it before routing, so the request still reaches `/client_secrets` and receives a live top-level `value`; because `redact_value` remains false, that credential is written verbatim to DEBUG logs. Normalize percent-encoded **unreserved** path characters before comparing issuer suffixes; do not indiscriminately decode reserved delimiters such as `%2F`.

HAYDEN-OAI (commented):
Reviewed commit: 0406c47d385c00c0a9ea49061f824b63b77cd482

Found one remaining issuer-path normalization bypass that can expose a newly issued credential in DEBUG logs; see the inline comment.

lib/openai/internal/logging.rb
jbeckwith-oai: Fixed in a2200be0. Credential-issuer classification now normalizes only RFC 3986 unreserved percent-encoded path characters before matching the effective POST request. Added end-to-end coverage for encoded admin, service-account, Realtime, and translation issuer paths; both 307 and 308 redirects to encoded issuer paths; and an encoded `%2F` delimiter that must remain encoded so ordinary response values stay visible. The full suites pass on Ruby 3.3, 3.4, and 4.0 (711 tests / 3,326 assertions each), with RuboCop, Sorbet, and RBS validation clean.

HAYDEN-OAI (approved):
Reviewed 92063ec5 against merged #411: DEBUG logging fails closed for request and response bodies, including encoded and trailing-slash credential issuers, nested service-account keys, and body-preserving redirects. Issued credentials and ordinary response values remain intact for SDK callers. No security or compatibility issues found.

PR #413, opened 2026-08-13
fix: enable streaming when retrieving responses

jbeckwith-oai:
## Summary

- Fix stable and beta `retrieve_streaming` methods that encoded GET query parameters before adding their implicit `stream=true` discriminator, causing normal JSON responses to be exposed as empty event streams.
- Preserve existing `include`, `starting_after`, `include_obfuscation`, `beta=true`, streaming transport headers, and beta feature headers.
- Add four WebMock transport regressions that assert the actual request query and decode real server-sent events without passing `stream:` explicitly.

## Verification

- `bundle exec rake test`: 699 tests and 3,142 assertions pass on Ruby 3.3.12, 3.4.10, and 4.0.6.
- `bundle exec rake lint:rubocop`: 2,618 files pass.
- `bundle exec rake typecheck:sorbet`: passes.
- `bundle exec rake validate:rbs`: 1,212 RBS files pass.
- `bundle exec rake build:gem`: passes.

Upstream Castiron generator fix: https://github.com/openai/openai/pull/1292380.

### DIFF
--- a/lib/openai/resources/beta/responses.rb
+++ b/lib/openai/resources/beta/responses.rb
@@ -301,12 +301,12 @@ module OpenAI
         def retrieve_streaming(response_id, params = {})
           query_params = [:include, :include_obfuscation, :starting_after, :stream]
           parsed, options = OpenAI::Beta::ResponseRetrieveParams.dump_request(params)
-          query = OpenAI::Internal::Util.encode_query_params(parsed.slice(*query_params))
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Correctly serializes the implicit `stream=true` discriminator before building both stable and beta response-retrieval queries while preserving existing query parameters and beta headers.

PR #414, opened 2026-08-14
feat(api): Ultrafast tier, structured MCP and websocket errors, separate websocket events

apcha-oai:
## Summary

- Add GPT-5.5 dated and Pro model constants, the `ultrafast` service tier, optional model shutdown dates, and cache-only web-search controls.
- Expose reusable stable and beta MCP tool-call error unions with protocol, tool-execution, and HTTP variants; remove the inaccurate string case because the API neither accepts nor emits it.
- Model Responses WebSocket failures independently so their nested error payload, transport details, and stream metadata reflect the events applications actually receive.
- Add typed output-text annotations and text/image token usage, preserve recent streaming fixes, and explain safe reuse of completed reasoning items.


jbeckwith-oai (changes requested):
The shared output-text annotation schema is emitted as incompatible nominal Ruby types. The annotation-added event defines event-local annotation classes in runtime/RBI/RBS, so an annotation obtained from ResponseOutputText#annotations is not accepted by Sorbet/RBS as the event annotation even though both represent the identical schema. This is the Ruby analogue of the downstream incompatibility exposed by the Python Agents check. Please make the canonical/codegen representation reuse one named annotation union across output text and annotation-added events, then regenerate runtime, RBI, and RBS together. The remaining generated additions and parity otherwise look coherent.

jbeckwith-oai (approved):

PR #417, opened 2026-08-14
chore: enforce line length on constants

jbeckwith-oai:
Remove the broad `Layout/LineLength` exemption for constant assignments and split the two generated assignments that it was hiding.

This is the third atomic layer in the line-length ratchet stack. Baseline after removing the exemption: 2 offenses. Final: 0 offenses.

Generator impact is covered by openai/openai#1296847, which teaches Castiron to account for physical nesting indentation when rendering enum constants.

Validation:
- `bundle exec rubocop --only Layout/LineLength .` (2,621 files, 0 offenses)
- `bundle exec rake lint` (RuboCop, Sorbet, directive guard, and 1,212 RBS files)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,7 +46,6 @@ Layout/FirstMethodParameterLineBreak:
 Layout/LineLength:
   AllowedPatterns:
     - "^\\s*#.*$"
-    - "^\\s*[A-Z0-9_]+ = :"
     - "OpenAI::(Models|Resources|Test)::"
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Reviewed against the actual stacked base and corresponding Castiron generation change. Both audit-log enum constants preserve their existing symbols, request/response serialization, and public API; no PR-introduced correctness or compatibility concerns found.

PR #421, opened 2026-08-14
feat(api): add file and vector store polling helpers

jbeckwith-oai:
## Summary

- add `files.wait_for_processing` with server-directed polling, configurable intervals, and an overall deadline
- add vector-store file `create_and_poll`, `poll`, `upload`, and `upload_and_poll` helpers
- add vector-store file-batch `create_and_poll`, `poll`, and bounded-concurrency `upload_and_poll` helpers
- add shared polling errors and private polling/upload infrastructure
- document the helpers and add matching RBI and RBS declarations

## Why

The Ruby SDK exposes the underlying file and vector-store endpoints but is missing the convenience workflows available across the Python, TypeScript, and Go SDKs. Users currently have to compose uploads, attachments, status loops, server polling hints, and failure handling themselves.

## Cross-SDK feature matrix

This matrix reflects the current SDK repository surfaces reviewed on 2026-08-15. A dash means the SDK has the raw endpoint but no equivalent first-party convenience helper.

| Convenience workflow | Python | TypeScript | Go | Java | .NET | Ruby with this PR |
| --- | --- | --- | --- | --- | --- | --- |
| Wait for an uploaded file to finish processing | ✅ sync + async | ✅ async | — | — | — | ✅ `wait_for_processing` |
| Attach a file to a vector store and poll | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `create_and_poll` |
| Poll an existing vector-store file | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `poll` |
| Upload a file and attach it to a vector store | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `upload` |
| Upload, attach, and poll a vector-store file | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `upload_and_poll` |
| Create and poll a vector-store file batch | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `create_and_poll` |
| Poll an existing vector-store file batch | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `poll` |
| Upload and poll a vector-store file batch | ✅ sync + async | ✅ async | ✅ | — | — | ✅ `upload_and_poll` |

## Design

The API follows Ruby conventions while preserving the convenience of the other SDKs:

- polling durations are keyword arguments in seconds; `timeout: nil` waits indefinitely
- server `openai-poll-after-ms` hints are honored unless the caller supplies `poll_interval`
- invalid polling, concurrency, stream, and 2,000-file capacity inputs are rejected before any API side effects
- finite timeouts use one monotonic deadline across authentication, connection/configuration, request replay, and polling; per-attempt transport retries are disabled and `PollingTimeoutError` retains the last resource when available
- workload-identity refresh cooperates with that deadline without allowing an asynchronous timeout to strand the shared refresh state
- `timeout: nil` retains configured transport retries
- failed and cancelled vector resources are returned for caller inspection
- batch uploads accept any `Enumerable`, inspect at most the remaining capacity plus one item, and safely spool stream-backed bodies during caller-thread enumeration so block-scoped IO remains valid without retaining all body bytes in memory
- a fixed worker set uses a start gate and atomic claim cursor, enables normal interrupt delivery inside each worker, captures fatal worker failures, and joins every worker before returning or raising
- request options and required endpoint headers propagate across every stage of composite operations without mutating caller hashes
- partial-upload behavior is documented because remote uploads are not transactional

## Castiron generator impact

No Castiron change is required to ship these convenience helpers. They are Ruby-specific custom SDK methods and shared internals, and the generator preserves manual resource patches.

This PR also corrects header merging in the two generated vector-store `create` methods. That exposed a pre-existing generic Ruby renderer behavior where caller `extra_headers` could replace endpoint defaults. The targeted fix here is safe and persistent, while a separate Castiron follow-up should make default-plus-caller header merging generic so the issue cannot recur on other endpoints. That renderer improvement is not helper-specific and is intentionally outside this PR's scope.

## Validation

- Ruby 3.3, 3.4, and 4.0 full suites: 798 runs, 3,977 assertions, 0 failures, 0 errors on each version (1 existing skip)
- `bundle exec rake lint`: 2,637 Ruby files with no RuboCop offenses; all 1,217 RBS files valid; Sorbet clean
- executable lines changed by this PR: 355/355 covered (100%); changed branches: 137/144 covered (95.14%)
- cancellation/concurrency, memory, workload-identity, and connection-deadline regressions passed; the focused stress suite passed 20 consecutive runs
- the minimum supported `connection_pool` 2.2.3 passes the connection deadline/configuration regression suite
- exact-head GitHub CI and CodeQL green on `b555d951`
- live OpenAI API lifecycle completed for file wait, vector-file, and vector-batch upload/poll workflows; temporary files and vector stores were cleaned up


lib/openai/internal/poller.rb
```diff
@@ -0,0 +1,121 @@
...
+        headers = {
+          **options[:extra_headers].to_h,
+          "X-Stainless-Poll-Helper" => "true"
+        }
+        unless @poll_interval.nil?
+          milliseconds = [(@poll_interval * 1000).round, 1].max
+          headers["X-Stainless-Custom-Poll-Interval"] = milliseconds.to_s
+        end
+
+        {**options, extra_headers: headers}
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 696a388e. Poller now case-normalizes and merges endpoint, caller, and polling headers; both vector-store poll paths seed the required assistants=v2 header. Regression assertions cover file and batch polling.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,90 @@
...
+      def upload(files)
+        queue = SizedQueue.new(@max_concurrency)
+        finished = Object.new
+        lock = Mutex.new
+        uploaded = []
+        state = {error: nil}
+        workers = start_workers(queue, finished, lock, uploaded, state)
+
+        begin
+          files.each.with_index do |file, index|
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 696a388e by iterating with each_with_index, which always supplies a block to Enumerable#each. A regression uses a custom Enumerable whose each implementation yields directly, proving the documented input contract beyond arrays and enumerators.

lib/openai/resources/files.rb
```diff
@@ -164,6 +164,56 @@ def content(file_id, params = {})
...
+      )
+        poller = OpenAI::Internal::Poller.new(
+          operation: "file #{file_id}",
+          poll_interval: poll_interval,
+          timeout: timeout
+        )
+        options = poller.request_options(request_options)
+
+        loop do
+          file = retrieve(file_id, request_options: options)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 696a388e. Poller recomputes the remaining monotonic deadline before every retrieval, caps the request timeout to the shorter of the caller timeout and deadline remainder, and converts deadline exhaustion to PollingTimeoutError with the last resource when available. Tests cover an expired deadline, a stalled first retrieval, shorter caller timeouts across all pollers, and unbounded polling.

lib/openai/resources/files.rb
```diff
@@ -164,6 +164,60 @@ def content(file_id, params = {})
...
+               :deleted,
+               "deleted"
+            return file
+          else
+            raise OpenAI::Errors::PollingError,
+                  "Unexpected status while waiting for file #{file_id}: #{file.status.inspect}"
+          end
+        end
+      rescue OpenAI::Errors::APITimeoutError
+        poller.check_deadline!(file)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 3320b240. I narrowed the APITimeoutError rescue to an explicit begin block entered only after the poller and last-resource local are initialized. This makes the invariant statically obvious while avoiding a nil poller and safe-navigation in the normal control flow. The focused suite remains green at 30 tests / 135 assertions.

lib/openai/resources/vector_stores/files.rb
```diff
@@ -169,6 +219,152 @@ def delete(file_id, params)
...
+                 OpenAI::VectorStores::VectorStoreFile::Status::FAILED,
+                 OpenAI::VectorStores::VectorStoreFile::Status::CANCELLED
+              return file
+            else
+              raise OpenAI::Errors::PollingError,
+                    "Unexpected status while waiting for vector store file #{file_id}: #{file.status.inspect}"
+            end
+          end
+        rescue OpenAI::Errors::APITimeoutError
+          poller.check_deadline!(file)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 3320b240. I narrowed the APITimeoutError rescue to an explicit begin block entered only after the poller and last-resource local are initialized. This makes the invariant statically obvious while avoiding a nil poller and safe-navigation in the normal control flow. The focused suite remains green at 30 tests / 135 assertions.

lib/openai/resources/vector_stores/file_batches.rb
```diff
@@ -139,6 +194,132 @@ def list_files(batch_id, params)
...
+                 OpenAI::VectorStores::VectorStoreFileBatch::Status::CANCELLED
+              return batch
+            else
+              raise OpenAI::Errors::PollingError,
+                    "Unexpected status while waiting for vector store file batch " \
+                    "#{batch_id}: #{batch.status.inspect}"
+            end
+          end
+        rescue OpenAI::Errors::APITimeoutError
+          poller.check_deadline!(batch)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 3320b240. I narrowed the APITimeoutError rescue to an explicit begin block entered only after the poller and last-resource local are initialized. This makes the invariant statically obvious while avoiding a nil poller and safe-navigation in the normal control flow. The focused suite remains green at 30 tests / 135 assertions.

lib/openai/internal/poller.rb
```diff
@@ -0,0 +1,145 @@
...
+        unless @poll_interval.nil?
+          milliseconds = [(@poll_interval * 1000).round, 1].max
+          headers["x-stainless-custom-poll-interval"] = milliseconds.to_s
+        end
+
+        bounded = {**options, extra_headers: headers}
+        unless remaining.nil?
+          request_timeout = options[:timeout]
+          request_timeout = request_timeout.to_f.clamp(0..) unless request_timeout.nil?
+          bounded[:timeout] = [request_timeout, remaining].compact.min
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in f41452aa. Finite-deadline polling requests now force transport max_retries to zero because the transport timeout is per attempt; this makes the remaining deadline bound the complete retrieval. Indefinite polling preserves the client or request retry configuration. Regression tests prove both behaviors, and the full suite is green at 754 tests / 3,490 assertions.

lib/openai/internal/poller.rb
```diff
@@ -0,0 +1,149 @@
...
+
+          # The transport applies `timeout` to every retry attempt independently.
+          # Disable those retries so the polling deadline bounds the whole request.
+          bounded[:max_retries] = 0
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 3aa6fde1. Poller#request now atomically builds bounded request options and runs the full retrieval inside the remaining monotonic deadline, so workload-identity token acquisition and the automatic 401 replay are covered as well as transport I/O. A postcondition also rejects a response that arrives as the deadline expires, while inner timeout errors that occur before the polling deadline remain unchanged. Added focused tests for slow initial token refresh, slow 401 refresh/replay, and inner timeout preservation; Poller coverage is 66/66 lines and 34/34 branches. Full suite: 757 runs, 3504 assertions, 0 failures.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,90 @@
...
+        begin
+          files.each_with_index do |file, index|
+            break unless lock.synchronize { state[:error].nil? }
+
+            queue << [index, file]
+          end
+        rescue StandardError => e
+          lock.synchronize { state[:error] ||= e }
+        ensure
+          workers.length.times { queue << finished }
```
HAYDEN-OAI: **[P1] Make upload-worker shutdown safe under cancellation.** With `max_concurrency: 1`, block the first upload and enqueue a second file so this `ensure` blocks while pushing the `finished` sentinel into the full `SizedQueue`. Wrapping the helper in `Timeout.timeout(0.2)` interrupts that `ensure`: no sentinel is delivered and `workers.each(&:join)` is skipped. Releasing the first upload after the caller has received `Timeout::Error` then starts the second remote upload, after which the worker remains blocked forever in `queue.pop`. I reproduced both the post-return upload and permanently live worker against this exact commit. Protect teardown against asynchronous interruption and stop/join workers before allowing cancellation to escape.

lib/openai/resources/vector_stores/files.rb
```diff
@@ -169,6 +221,156 @@ def delete(file_id, params)
...
+        # @return [OpenAI::Models::VectorStores::VectorStoreFile]
+        def upload(
+          vector_store_id,
+          file:,
+          attributes: nil,
+          chunking_strategy: nil,
+          request_options: {}
+        )
+          uploaded = @client.files.create(file: file, purpose: :assistants, request_options: request_options)
+          params = {file_id: uploaded.id, request_options: request_options}
```
HAYDEN-OAI: **[P2] Preserve the required beta header on vector-store create/attach requests.** Repro: `upload_and_poll("vs_123", file: Pathname("one.pdf"), request_options: {extra_headers: {"X-Trace" => "1"}})`. The `/files` upload succeeds, but these unchanged options reach the generated `create`, whose `{extra_headers: {"OpenAI-Beta" => "assistants=v2"}, **options}` replaces the required beta-header hash with the caller's headers. The attachment POST therefore lacks `OpenAI-Beta: assistants=v2`, fails on deployments that require it, and leaves the successfully uploaded file orphaned. `files.create_and_poll` and both file-batch composite helpers have the same create-stage omission; the earlier fix only merges headers for polling GETs. Merge endpoint and caller headers for every vector-store POST without mutating the original request options.

lib/openai/resources/vector_stores/file_batches.rb
```diff
@@ -139,6 +196,135 @@ def list_files(batch_id, params)
...
+          poll_interval: nil,
+          timeout: OpenAI::Internal::Poller::DEFAULT_TIMEOUT,
+          request_options: {}
+        )
+          OpenAI::Internal::Poller.validate!(poll_interval: poll_interval, timeout: timeout)
+          uploaded = OpenAI::Internal::VectorStoreFileUploader.new(
+            client: @client,
+            max_concurrency: max_concurrency,
+            request_options: request_options
+          ).upload(files)
```
HAYDEN-OAI: **[P2] Enforce the 2,000-file batch limit before uploading.** `FileBatchCreateParams` documents a hard maximum of 2,000 total file IDs, but this call uploads the entire enumerable before combining it with `file_ids`. Repro: pass 2,000 existing IDs and `files: [Pathname("extra.pdf")]`: `/files` succeeds, then batch creation submits 2,001 IDs and necessarily fails, leaving the new remote file unattached. Passing 2,001 new inputs can orphan every upload; an infinite enumerable can continuously upload files and grow `uploaded` without ever reaching batch creation. Validate existing-plus-new capacity before side effects and bound enumeration/uploads to the remaining batch capacity.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,90 @@
...
+        lock = Mutex.new
+        uploaded = []
+        state = {error: nil}
+        workers = start_workers(queue, finished, lock, uploaded, state)
+
+        begin
+          files.each_with_index do |file, index|
+            break unless lock.synchronize { state[:error].nil? }
+
+            queue << [index, file]
```
HAYDEN-OAI: **[P2] Keep block-scoped IO alive until its upload has consumed it.** A documented lazy `Enumerable<IO>` can be written as `Enumerator.new { |y| File.open("one.pdf", "rb") { |io| y << io } }`. This queue push returns immediately, allowing enumeration to resume and the `File.open` block to close `io` before a worker executes `@client.files.create`; multipart encoding subsequently calls `IO.copy_stream` on the closed descriptor and raises `IOError: closed stream`. The race exists even with `max_concurrency: 1` whenever the worker is delayed. Preserve yielded-stream lifetime through worker consumption, or explicitly narrow the advertised arbitrary-Enumerable/IO contract.

HAYDEN-OAI (commented):
Reviewed commit `3aa6fde141369183831c030b58eea4c7a0361d23`. I found four new substantive correctness issues in the added upload workflows, with precise reproductions in the inline comments. The previously resolved automated findings were also checked.

lib/openai/resources/vector_stores/file_batches.rb
jbeckwith-oai: Fixed in cb2eebca. upload_and_poll now computes the remaining capacity from existing file_ids and the uploader enumerates at most capacity + 1 inputs before starting any worker or request. Oversized existing IDs fail without enumerating files; 2,001 new inputs and an unbounded enumerable both fail with zero HTTP requests. The unbounded test with 1,999 existing IDs confirms enumeration stops after exactly two yields.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in cb2eebca. The uploader now preflights inputs, uses a closed Queue, and masks asynchronous exceptions throughout worker startup and teardown. Cancellation is enabled only while waiting; teardown marks queued work stopped and joins every worker before the exception can escape. The regression test blocks upload 1, triggers Timeout, verifies the caller cannot return early, releases upload 1, and confirms upload 2 never starts. The concurrency suite also passed 20 consecutive stress runs.

lib/openai/resources/vector_stores/files.rb
jbeckwith-oai: Fixed in cb2eebca. Both vector-store create methods now merge normalized caller headers with the endpoint beta header after copying the request options, so custom headers no longer replace OpenAI-Beta and the caller hash is not mutated. Coverage exercises create_and_poll and upload_and_poll for both Files and FileBatches with custom headers.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in cb2eebca by explicitly narrowing and enforcing the stream-lifetime contract. Inputs are fully enumerated before workers start, then raw IO and FilePart-backed IO are checked for closure; a block-scoped File.open enumerable now raises ArgumentError before any request. README and API docs state that yielded streams must remain open until upload_and_poll returns and recommend paths for block-scoped lazy enumeration.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,88 @@
...
+                lock.synchronize { uploaded[index] = result }
+              rescue StandardError => e
+                lock.synchronize { state[:error] ||= e }
+              end
+            end
+          end.tap { _1.report_on_exception = false }
+        end
+        workers
+      rescue ThreadError
+        workers.length.times { queue << finished }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in f86ecdb4. The worker array is initialized before the explicit begin/rescue region, so the ThreadError cleanup path can always enumerate and join the workers that were successfully started. The uploader was subsequently hardened further in cb2eebca with cancellation-safe startup/teardown and complete worker joins; the full suite and 20-run concurrency stress test are green.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,88 @@
...
+        workers.each(&:join)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in f86ecdb4. The worker array is initialized before the explicit begin/rescue region, so the ThreadError cleanup path can always enumerate and join the workers that were successfully started. The uploader was subsequently hardened further in cb2eebca with cancellation-safe startup/teardown and complete worker joins; the full suite and 20-run concurrency stress test are green.

lib/openai/internal/poller.rb
```diff
@@ -0,0 +1,172 @@
...
+      # @return [Object]
+      def request(request_options, extra_headers: {}, resource: nil, &block)
+        remaining = check_deadline!(resource)
+        options = bounded_request_options(request_options, extra_headers: extra_headers, remaining: remaining)
+
+        result =
+          if remaining.nil?
+            block.call(options)
+          else
+            Timeout.timeout(remaining) { block.call(options) }
```
HAYDEN-OAI, line 64: **[P1] Do not asynchronously interrupt workload-identity refresh before cleanup is installed.** This whole-request `Timeout.timeout` can fire after `WorkloadIdentityAuth#get_token` sets `@refreshing = true` (`auth/workload_identity_auth.rb:47`) but before `perform_refresh` installs the ensure that clears it. I reproduced the real SDK path by pausing before line 50: `wait_for_processing(..., timeout: 0.02)` raised `PollingTimeoutError`, left `@refreshing == true`, and a subsequent ordinary `client.files.retrieve` blocked indefinitely. One timed-out poll can therefore permanently wedge a shared workload-identity client and every concurrent/future request. Make refresh-state setup and cleanup atomic against asynchronous exceptions, or propagate a cooperative absolute deadline through authentication instead of interrupting arbitrary instructions.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,125 @@
...
+        return unless (content.is_a?(IO) || content.is_a?(StringIO)) && content.closed?
+
+        raise ArgumentError, "IO inputs yielded by `files` must remain open until `upload_and_poll` returns"
+      end
+
+      private def start_workers(queue, worker_count, lock, uploaded, state)
+        workers = []
+        begin
+          worker_count.times do
+            workers << Thread.new do
```
HAYDEN-OAI: **[P1] Restore asynchronous interrupt delivery inside upload workers.** These workers are created under the caller's `Thread.handle_interrupt(Exception => :never)` block, and Ruby child threads inherit that interrupt mask. With an injected `OpenAI::HTTPClient` that implements its documented request timeout using `Timeout.timeout`, direct `files.create(..., request_options: {timeout: 0.03})` correctly raised `APITimeoutError` after 33 ms, while the same upload through public `file_batches.upload_and_poll(..., max_concurrency: 1)` incorrectly succeeded after 152 ms. Ruby's `Net::HTTP` also implements workload-identity connection timeouts this way, so stalled authentication/uploads can hang indefinitely and defeat cancellation cleanup. Restore interrupt delivery inside each worker while keeping caller startup and teardown cancellation-safe.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,125 @@
...
+          workers
+        rescue ThreadError
+          stop_workers(workers, lock, state)
+          raise
+        end
+      end
+
+      private def stop_workers(workers, lock, state)
+        lock.synchronize { state[:stopping] = true }
+        workers.each(&:join)
```
HAYDEN-OAI, line 214: **[P1] Join every upload worker even when another dies with a non-StandardError.** Worker execution only rescues `StandardError`, but the supported `OpenAI::HTTPClient#execute` abstract implementation raises `NotImplementedError < ScriptError`. With two concurrent uploads, worker A raises that exception while worker B remains blocked; the initial join propagates A's failure, then this cleanup immediately rejoins A, raises again, and never joins B. A public-SDK reproduction returned `NotImplementedError` to the caller while B was still alive, and B subsequently completed a `/v1/files` upload after the helper had returned. Ensure cleanup joins every started worker before reraising any fatal worker failure.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,125 @@
...
+
+        uploaded
+      end
+
+      private def stage(files, max_files:)
+        unless max_files.is_a?(Integer) && max_files >= 0
+          raise ArgumentError, "`max_files` must be a non-negative integer"
+        end
+
+        staged = files.each_with_object([]) do |file, result|
```
HAYDEN-OAI: **[P2] Bound staged bytes before consuming lazy upload bodies.** `files.each_with_object([])` exhausts and retains every input before the first worker starts, so `max_concurrency` bounds request count but does not bound memory for documented lazy `String`/`StringIO` inputs. On this exact head, a one-shot generator yielding 100 x 1 MiB strings with `max_concurrency: 1` had `yielded_before_first_upload=100` and retained 104,861,540 additional string bytes before its first upload; valid batches can approach 2,000 x 512 MB and exhaust process memory before any request. Keep staged payload bytes bounded, safely spool/clean up large inputs, or reject excessive in-memory batches.

lib/openai/internal/poller.rb
HAYDEN-OAI, line 64: **[P2] Enforce polling deadlines while pooled connections are configured.** `ConnectionPool#with` runs connection checkout/construction under `Thread.handle_interrupt(Exception => :never)`, including the documented `OpenAI::NetHTTPClient.new { |connection| ... }` configurator. The asynchronous timeout here cannot interrupt that region: with a configurator sleeping 250 ms, real `client.files.wait_for_processing("file_123", timeout: 0.02)` took 251 ms before raising, while a blocked configurator can hang forever. All three polling helpers therefore violate their strict finite-deadline contract during connection setup. Propagate/check the remaining deadline through connection construction or avoid relying on asynchronous interrupt delivery inside a masked region.

HAYDEN-OAI (commented):
Reviewed the complete file/vector-store helper change, multipart/header propagation, bounded ingestion, cancellation, worker lifecycle, workload-identity authentication, and finite polling deadlines. Focused reproductions against the exact PR head and public Ruby SDK identified three high-priority concurrency/authentication failures and two additional resource/deadline defects; the inline comments include concrete triggering scenarios and observed behavior.

lib/openai/internal/poller.rb
jbeckwith-oai: Fixed in eef6afcb. WorkloadIdentityAuth#get_token now makes the @refreshing transition and installs its cleanup while asynchronous exceptions are masked, then restores immediate delivery only around the refresh itself. One absolute monotonic deadline is carried through initial authentication, token exchange, the 401 refresh, replay, and transport. The regression test pauses exactly at the perform_refresh boundary, injects Timeout::Error, verifies @refreshing is cleared, and proves the next token request succeeds; waiter deadline and failed-refresh cases are also covered. Full suites pass on Ruby 3.3, 3.4, and 4.0.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in eef6afcb. Every upload worker explicitly enters Thread.handle_interrupt(Exception => :immediate) before it passes the start gate or performs a request, so it does not retain the parent teardown mask. test_upload_workers_honor_async_timeouts proves a 20 ms Timeout.timeout interrupts a 200 ms simulated request in under 150 ms. The complete focused lifecycle suite also passed 20 stress repetitions.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in eef6afcb. A single catch-all at the worker boundary records every Exception, including ScriptError subclasses such as NotImplementedError, and stops new claims. Parent teardown remains interruption-safe and joins every started worker before propagating the recorded failure. The two-worker regression forces worker A to raise NotImplementedError after worker B has claimed work and asserts B has finished before the helper reraises; a separate test proves caller cancellation still takes precedence while cleanup joins complete.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in eef6afcb. Caller-thread preflight now streams String, StringIO, IO, and FilePart-backed bodies into closed Tempfiles and retains only lightweight Pathname descriptors; success, upload failure, partial staging failure, and cleanup-error paths are covered. Workers claim staged entries by an atomic cursor, so at most max_concurrency bodies are in flight. The ObjectSpace regression blocks uploads for 16 one-MiB lazy strings, observes less than four MiB of retained string growth, and verifies every spool is removed. Block-scoped IO and multipart filename/content-type preservation are also tested.

lib/openai/internal/poller.rb
jbeckwith-oai: Fixed in eef6afcb. The ConnectionPool factory now creates only an empty PooledConnection slot. Actual Net::HTTP construction and the user configurator run inside the pool interruptible yield, with cooperative deadline checks before and after configuration and discard/close cleanup if setup is interrupted. Separate regressions block connection construction and configuration for two seconds under a 20 ms poll deadline, assert prompt PollingTimeoutError with no request, and then confirm the pool remains reusable.

lib/openai/auth/workload_identity_auth.rb
```diff
@@ -29,31 +31,51 @@ def initialize(
...
-        @mutex.synchronize do
-          @cond_var.wait(@mutex) while @refreshing && token_unusable?
-
-          unless token_unusable? || needs_refresh?
-            return @cached_token
+      # @api private
+      #
+      # @param deadline [Float, nil] absolute monotonic deadline for this request
+      # @return [String]
+      def get_token(deadline: nil)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8edb5fbd. `WorkloadIdentityAuth#get_token(deadline:)` and `invalidate_token` now have public declarations in both `rbi/openai/auth.rbi` and `sig/openai/auth.rbs`, matching the runtime methods' public `@api private` visibility and preserving direct dispatch from `Client`. Validation: all 1,217 RBS files pass, Sorbet reports no errors, RuboCop reports 2,637 files with no offenses, and the workload-identity suite passes 23 runs / 101 assertions.

lib/openai/net_http_client.rb
```diff
@@ -137,17 +141,34 @@ def initialize(original)
...
+        if pooled_connection.connection.nil?
+          connection = nil
+          begin
+            connection = connect(url: url)
+            configure_connection(connection, url, deadline: deadline)
+            pooled_connection.connection = connection
+          ensure
+            if pooled_connection.connection.nil?
+              close_connection(connection) unless connection.nil?
+              pool.discard_current_connection
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in b555d951. The 3.x-only `discard_current_connection` call is gone. A failed setup leaves the checked-out `PooledConnection` slot empty after closing any provisional connection; both connection_pool 2.2.3 and 3.x check that lightweight slot back in, and the next checkout naturally retries construction. This preserves the declared `>= 2.2.3` dependency contract without a major-version floor or compatibility branch. I reproduced the original three failures under 2.2.3, then verified the focused deadline/configuration suite passes there (5 runs / 25 assertions); the supported full suite passes 798 / 3,977 and lint/type validation is clean.

lib/openai/client.rb
```diff
@@ -177,9 +177,13 @@ class Client < OpenAI::Internal::Transport::BaseClient
       workload_identity_auth_header = "Bearer #{WORKLOAD_IDENTITY_API_KEY_PLACEHOLDER}"
       return super unless request[:headers]["authorization"] == workload_identity_auth_header
 
-      token = @workload_identity_auth.get_token
+      deadline = request[:timeout]&.then { OpenAI::Internal::Util.monotonic_secs + _1 }
+      token = @workload_identity_auth.get_token(deadline: deadline)
```
HAYDEN-OAI: **[P2] Preserve the SDK timeout contract during workload-identity refresh.** The new authentication deadline is evaluated before the transport's timeout/error handling, so when a workload-identity provider takes longer than `request_options[:timeout]`, `get_token` / `request_with_remaining_timeout` leak a raw `Timeout::Error`. With a 45 ms provider, `request_options: {timeout: 0.01}`, and `max_retries: 2`, both ordinary `files.retrieve` and all three polling helpers raised `Timeout::Error` after one attempt and zero HTTP requests instead of the documented `OpenAI::Errors::APITimeoutError`; callers' SDK-specific rescue paths and configured timeout retries are bypassed. Normalize authentication-deadline exhaustion to the SDK timeout type and send it through the normal retry/error path while preserving the overall polling deadline.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,200 @@
...
+        in String
+          path = spool(file, temporary_files)
+          OpenAI::FilePart.new(path, filename: "upload", content_type: "text/plain")
+        in StringIO
+          path = spool(file, temporary_files)
+          OpenAI::FilePart.new(path, filename: "upload", content_type: "application/octet-stream")
+        in IO
+          filename = file.to_path.nil? ? "upload" : ::File.basename(file.to_path)
+          path = spool(file, temporary_files)
+          OpenAI::FilePart.new(path, filename: filename, content_type: "application/octet-stream")
```
HAYDEN-OAI: **[P2] Preserve raw-IO non-retry semantics after spooling.** Wrapping an incoming `IO` as `FilePart(Pathname)` hides the original stream from `FileInput.dump`, which normally sets `can_retry = false` for `IO` and IO-backed `FilePart`; this new batch path therefore silently enables retries despite the SDK's documented raw-IO no-retry contract. With `max_retries: 1` and a server that persists `/files` before returning 500, direct `files.create(file: io)` made one request, while `file_batches.upload_and_poll(files: [io])` made two unkeyed uploads, attached only `file_uploaded_2`, and orphaned `file_uploaded_1`. Preserve original stream replayability metadata or disable retries for originally streamed inputs unless safe per-file idempotency is guaranteed.

lib/openai/internal/vector_store_file_uploader.rb
```diff
@@ -0,0 +1,200 @@
...
+          end.tap { _1.report_on_exception = false }
+        end
+      end
+
+      private def upload_one(work, lock, uploaded)
+        index, file = work
+        result = @client.files.create(
+          file: file,
+          purpose: :assistants,
+          request_options: @request_options
```
HAYDEN-OAI: **[P2] Do not reuse one idempotency key for distinct file uploads.** Every worker receives the same `@request_options`, so `request_options: {extra_headers: {"Idempotency-Key" => "operation-key"}}` is sent unchanged on each distinct `POST /files` and subsequently on batch creation. In a public-SDK reproduction against an idempotency-aware transport, uploading `["ALPHA", "BRAVO"]` replayed the first file response for the second upload, produced duplicate file IDs, and failed batch creation with `OpenAI::Errors::BadRequestError`. Derive a stable, distinct idempotency key for each file/batch operation while retaining the same key across retries of that specific operation.

HAYDEN-OAI (commented):
Independently re-reviewed commit `b555d951e4836b512012f948b554bb3b3baad98d` and the previously resolved feedback, including minimum supported connection-pool behavior. Direct public-SDK reproductions identified three new correctness issues involving workload-identity timeout classification, per-upload idempotency isolation, and unintended retries of raw IO; concrete triggers and observed outcomes are inline.

lib/openai/client.rb
jbeckwith-oai: Fixed in eb32c589. Workload-identity preparation now runs inside the BaseClient APIConnectionError retry boundary, and deadline exhaustion is normalized to OpenAI::Errors::APITimeoutError with the original Timeout::Error as its cause. Normal request max_retries therefore applies, while finite pollers still force max_retries: 0 and translate an exhausted overall deadline to PollingTimeoutError. The public-SDK regression test uses a slow subject-token provider and proves 3 auth attempts with max_retries: 2, zero API requests, and the SDK timeout class; the bounded polling and 401-replay tests remain green. Verified by the 37-run/189-assertion polling suite and full 807-run/4,012-assertion suites on Ruby 3.3, 3.4, and 4.0.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in eb32c589. Staging now retains whether an input originated as raw IO. Stream-backed inputs are still spooled before workers start, but each worker reopens the spool as an IO only for files.create; the canonical FileInput dumper therefore preserves max_retries: 0 instead of treating the staged Pathname as retryable. Direct files.create behavior is untouched. A public fail-after-persistence 500 regression proves both direct raw IO and batch raw IO make exactly one request, and the focused uploader suite is green at 19 runs/83 assertions plus 20 repeated concurrency/lifecycle iterations.

lib/openai/internal/vector_store_file_uploader.rb
jbeckwith-oai: Fixed in eb32c589. The new private RequestOptionsScope snapshots caller options and deterministically derives a fixed-size SHA-256 child key from the caller key plus a logical operation label. Uploads use file-upload-<index>, batch creation uses file-batch, and the single-file composite helper separately scopes upload and attach. Retries of one operation keep the same key, while distinct uploads and the batch never collide. The fail-after-persistence regression observes two attempts under each of two distinct upload keys and two attempts under a third batch key; the scope unit suite is green at 5 runs/14 assertions and changed line/branch coverage is 100%.

HAYDEN-OAI (commented):
Independently re-reviewed exact commit `eb32c58974f84b86bf4a6f3e71fa800ef9eff701` and verified the prior workload-identity timeout, raw-IO retry, and extra-header idempotency fixes. Focused public-SDK reproductions identified one remaining correctness issue: structured `request_options[:idempotency_key]` is derived but never sent, so retryable uploads can persist duplicate files and leave originals orphaned. The concrete reproduction and fix direction are inline.

lib/openai/internal/request_options_scope.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      # @api private
+      #
+      # @param operation [String]
+      # @return [Hash{Symbol=>Object}]
+      def child(operation)
+        options = @options.dup
+        key = options[:idempotency_key]
+        options[:idempotency_key] = derive(key, operation) unless key.nil?
```
HAYDEN-OAI: **[P2] Materialize structured idempotency keys on composite requests.** Passing `request_options: {idempotency_key: "operation-key"}` derives child values here but sends none: `OpenAI::Client` never configures `BaseClient#idempotency_header`, so `build_request` discards this documented option. Against a transport that persists uploads before a retryable 500, a public two-file batch with `max_retries: 1` completed after creating four unkeyed files, attaching only files 2 and 4 and orphaning files 1 and 3; its batch POST was unkeyed too. Emit each derived structured key in the actual `Idempotency-Key` header while preserving explicit caller-header precedence.

jbeckwith-oai: Fixed in 4e96bbad3c9f1a5e4273123837b509bf6e0e136d. RequestOptionsScope now materializes every derived structured child key into the actual Idempotency-Key header at the canonical composite-operation boundary. A case-insensitive explicit caller header remains authoritative, including explicit nil suppression, and direct files.create behavior is unchanged. The public regression simulates persistence before a retryable 500 for two uploads and the final batch POST, then proves each upload reuses its own distinct deterministic key across both attempts and the batch reuses a separate deterministic key; explicit-header precedence is also covered through the public helper and the scope unit suite. Validation passed: focused request-options/uploader/polling/resource/retry/client suites; full Ruby 3.3, 3.4, and 4.0 suites at 808 runs / 4,012 assertions each; lint, Sorbet, RBS, gem packaging, and docs generation.

HAYDEN-OAI (approved):
Independently re-reviewed exact commit 4e96bbad3c9f1a5e4273123837b509bf6e0e136d. Confirmed the addressed structured idempotency fix reaches the actual Idempotency-Key HTTP header: each upload and batch write receives its own deterministic key, retries reuse that key, and persisted-before-failure retries do not create orphan files. Also verified explicit-header precedence, provider/auth middleware, beta headers, caller-option preservation, raw-IO ownership/replayability, bounded enumerable staging and worker lifecycle, partial failures, polling deadlines, and workload-identity timeout handling. The remaining automated proactive-refresh warning reproduces on the base branch and is not introduced by this change. No new substantive issues found.

lib/openai/auth/workload_identity_auth.rb
```diff
@@ -29,31 +31,51 @@ def initialize(
...
+        check_deadline!(deadline)
+        action = nil
+        token = nil
+
+        # Installing refresh cleanup is part of the state transition. No async
+        # exception may observe @refreshing after it changes but before the ensure.
+        Thread.handle_interrupt(Exception => :never) do
+          @mutex.synchronize do
+            if @refreshing
+              action = :wait
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `0ddaeada`. The workload-identity cache transition now distinguishes a usable cached token from an unusable/expired one while refresh is in flight: concurrent callers return the still-valid token during proactive refresh, while expiry or explicit invalidation continues to wait for the single refresher under the caller deadline. Added deterministic regressions for short-deadline parallel callers, expired-token waiters, refresh failure, and invalidation. On the rebased tree, the focused auth suite is green on Ruby 3.3/3.4/4.0; the complete suite is 826 tests / 4,127 assertions / 0 failures on each runtime; `get_token` is 25/25 lines and 12/12 branches covered; lint, RBS, Sorbet, gem packaging, docs, and two-pass thermo review all passed.

PR #422, opened 2026-08-15
fix: ignore response stream keepalive events

### LINKED ISSUE #287: Unknown Stream Events Classified as a ResponseErrorEvent
Intermittently, when using `OpenAI::Client.new(...).responses.stream(...)` we will receive a `keepalive` event: e.g.,`{"sequence_number":3,"type":"keepalive"}`. We are using Azure OpenAI, so my guess is that these `keepalive` events are sent to keep the connection open when there's a gap between received events or a larger time-to-first-token (although there is no mention of these events in the Azure documentation of this).

In our code, we have a `case` statement branching on different types of `ResponseStreamEvent`:
```ruby
      # Translates events from the OpenAI::Models::Responses::ResponseStreamEvent to StreamResponseEvent
      def parse_stream_event(event:)
        case event
        when OpenAI::Models::Responses::ResponseCreatedEvent
          parse_created_event(event:)
        when OpenAI::Models::Responses::ResponseTextDeltaEvent
          parse_output_text_delta_event(event:)
        when OpenAI::Models::Responses::ResponseCompletedEvent
          parse_completed_event(event:)
        when OpenAI::Models::Responses::ResponseFailedEvent
          parse_failed_event(event:)
        when OpenAI::Models::Responses::ResponseIncompleteEvent
          report_token_usage(response: event.response)
          parse_incomplete_event(event:)
        when OpenAI::Models::Responses::ResponseErrorEvent
          parse_error(event:)
        end
      end
```

We are seeing these `keepalive` events being classified as a `ResponseErrorEvent`, resulting in our code treating these `keepalive` events as errors. Looking at the `OpenAI::Internal::Type::Union#coerce` method, it seems as though it tries to match the best-fit variant but does not enforce `type` matching the expected value for the variant.

We've since added logic to skip these events if `type == 'keepalive'` but I wanted to share this in case this is not expected behavior and the logic for `Union#coerce` should be altered to be more strict

jbeckwith-oai:
## Summary

- drop transport-level `keepalive` frames before typed stream-event coercion
- prevent an undocumented heartbeat from being misclassified as a known Responses event
- keep raw/helper, stable/beta, and retrieved response streams on their existing typed event surfaces
- leave a general `UnknownStreamEvent` contract to separate cross-SDK work

## Root cause

When a discriminated union receives an unknown `type`, `Union#coerce` falls back to structural ranking across its known variants. A sparse heartbeat such as `{type: "keepalive", sequence_number: 3}` can therefore be materialized as an unrelated generated event model while retaining the original discriminator.

## Behavior

`keepalive` is an undocumented transport heartbeat, not an actionable Responses event. The decoded frame is now skipped at the stream boundary before generated model coercion, so callers continue to receive only modeled events and the heartbeat cannot affect response accumulation or `starting_after` handling.

Other unknown event types retain their existing behavior. A typed unknown-event variant that preserves the discriminator and original payload is intentionally out of scope for this PR.

## Generator scope

No Castiron change is required. The fix is confined to the SDK-owned stream decoder and regression coverage; generated models, resources, RBI, and RBS remain unchanged.

## Validation

- targeted Responses streaming and retrieval tests: 39 runs / 113 assertions
- full suite with mock server on Ruby 4.0.6: 725 runs / 3,357 assertions
- RuboCop: 2,625 files, no offenses
- Sorbet: no errors
- RBS validation: 1,214 files, no errors

Fixes #287


jbeckwith-oai:
@codex review

jbeckwith-oai:
@codex review

jbeckwith-oai:
@apcha-oai, there seem to be two high-level ways to handle the Azure `keepalive` event:

### Option 1: Raw pass-through compatibility shim

Intercept `type: "keepalive"` at the Responses stream-union boundary, preserve the payload as a raw hash, and skip it during response-state accumulation.

**Advantages**
- Smallest and most surgical change.
- Preserves the undocumented payload exactly.
- Tolerates additional or changing Azure fields without inventing an SDK contract.

**Downsides**
- The high-level stream no longer has a uniform model-like event API: callers using `event.type` or `event.sequence_number` must branch for a hash.
- RBI/RBS stream element types must include a broad hash type, weakening exhaustiveness.
- It is essentially “snag the unknown event from the stream and hope for the best.”

### Option 2: SDK-local typed keepalive model

Add a small handwritten `OpenAI::Streaming::ResponseKeepAliveEvent` with a constant `type` and typed `sequence_number`, and have the compatibility hook coerce the Azure payload into it.

**Advantages**
- Preserves the existing `event.type`, `event.sequence_number`, and `event[:type]` interface.
- Keeps Sorbet/RBS stream unions precise and exhaustive.
- Gives callers an explicit class to match.

**Downsides**
- Creates a de facto SDK API for an undocumented, Azure-specific transport event.
- Adds handwritten Ruby/RBI/RBS maintenance outside the generated model surface.
- Requires a migration or aliasing plan if the canonical schema later adds an official keepalive event.
- Validating fields such as `sequence_number` assumes the undocumented payload remains stable.

Which direction do you prefer here: the minimal raw compatibility shim, or the SDK-local typed event?

apcha-oai:
I’d suggest a third option here: just drop `keepalive` for now. It’s an undocumented transport heartbeat, not an actionable Responses event. I don’t think exposing it is worth making callers branch between model objects and hashes.

Longer term, I think the right approach is something like `UnknownStreamEvent`. Looking across the SDKs, there doesn’t seem to be a consistent solution today:

| SDK | Current behavior |
|---|---|
| **Python** | An unknown discriminator can fall back to constructing a known union variant, preserving `.type` but giving the event a misleading class. [Source](https://github.com/openai/openai-python/blob/main/src/openai/_models.py#L619-L654) |
| **Node** | Raw streams preserve the JSON object and `.type`, but the high-level accumulator explicitly tolerates `keepalive` and throws for other unhandled events. [Raw stream](https://github.com/openai/openai-node/blob/main/src/core/streaming.ts#L72-L87), [accumulator](https://github.com/openai/openai-node/blob/main/src/internal/responses/response-accumulator.ts#L792-L845) |
| **Java** | Preserves unknown JSON in the event union and provides a `Visitor.unknown` fallback. The accumulator ignores unknown variants. [Event model](https://github.com/openai/openai-java/blob/main/openai-java-core/src/main/kotlin/com/openai/models/responses/ResponseStreamEvent.kt#L650-L658), [accumulator](https://github.com/openai/openai-java/blob/main/openai-java-core/src/main/kotlin/com/openai/helpers/ResponseAccumulator.kt#L355-L360) |
| **Go** | Preserves `Type` and raw JSON on the common event struct; `AsAny()` returns `nil` when no known variant matches. [Event union](https://github.com/openai/openai-[go/blob/main/responses/response.go#L22239-L22275](https://www.golinks.io/blob/main/responses/response.go#L22239-L22275?trackSource=github)), [variant selection](https://github.com/openai/openai-[go/blob/main/responses/response.go#L22493-L22612](https://www.golinks.io/blob/main/responses/response.go#L22493-L22612?trackSource=github)) |

Java and Go are useful precedents, but I think we should establish a clearer cross-SDK contract:

- `event.type` always preserves the original discriminator.
- Unknown types are never coerced into unrelated known event classes.
- New event types become an explicit unknown-event variant, with the original payload available to callers.

That would make both type-based and class-based dispatch more predictable. It also matters because [adding new streaming event types is considered a backward-compatible API change](https://developers.openai.com/api/reference/overview#backwards-compatibility).

I’d keep that broader work out of this PR, though, and filter this specific heartbeat for now and we can backlog the more correct fix

HAYDEN-OAI (approved):
Reviewed keepalive filtering across stable/beta Responses creation, retrieval, and helper streams; checked typed-event compatibility, shared decoder callers, terminal/error propagation, resume sequencing, SSE lifecycle, and exact-head versus base behavior. No substantive issues found.

PR #425, opened 2026-08-17
fix: validate RuboCop directives without a Git checkout

apcha-oai:
## Summary
The suppression guard added in #408 assumes every source tree has a Git index. That breaks linting in temporary SDK-generation directories and source archives with `git ls-files failed`. It also silently skips new Ruby files until they are staged. A source check should inspect the files RuboCop would inspect, regardless of how the tree was created.

Use RuboCop’s configured targets directly, preserving the existing suppression rules and dependency/tool-output exclusions. Regression tests cover a source tree without Git metadata, excluded dependency files, and a new untracked file. The guard tests and full RuboCop task pass.

### DIFF
--- a/scripts/rubocop_directive_guard.rb
+++ b/scripts/rubocop_directive_guard.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "date"
-require "open3"
 require "ripper"
 require "rubocop"
 
@@ -62,17 +61,11 @@ module RuboCopDirectiveGuard
     false
   end
 
-  def tracked_ruby_sources
-    paths, status = Open3.capture2("git", "ls-files", "-z")
-    raise "git ls-files failed" unless status.success?
-
-    tracked_paths = paths.split("\0").to_h { [_1, true] }
+  def ruby_sources
     root = File.expand_path(".")
     root_prefix = "#{root}/"
-    rubocop_target_paths.filter_map do |target|
+    rubocop_target_paths.map do |target|
       path = File.expand_path(target).delete_prefix(root_prefix)
-      next unless tracked_paths.key?(path)
-
       [path, File.read(target)]
     end
   end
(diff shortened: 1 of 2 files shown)


scripts/rubocop_directive_guard.rb
```diff
@@ -62,17 +61,11 @@ def valid_iso_date?(value)
...
-    paths, status = Open3.capture2("git", "ls-files", "-z")
-    raise "git ls-files failed" unless status.success?
-
-    tracked_paths = paths.split("\0").to_h { [_1, true] }
+  def ruby_sources
     root = File.expand_path(".")
     root_prefix = "#{root}/"
-    rubocop_target_paths.filter_map do |target|
+    rubocop_target_paths.map do |target|
       path = File.expand_path(target).delete_prefix(root_prefix)
```
HAYDEN-OAI, line 68: [P2] Keep discovered sources inside the project root

`RuboCop::TargetFinder` follows directory symlinks, but `File.expand_path(target)` only normalizes the lexical path before `File.read` follows the link. In a checkout or Git-free archive containing `external_link -> ../external`, with `../external/outside.rb` containing `# rubocop:disable Lint`, this exact commit fails with `external_link/outside.rb:1: department-wide rubocop:disable Lint is forbidden`; the parent commit succeeds because Git tracks only `external_link`, not the outside file. This newly reads files outside the project and makes unrelated linked dependency/build trees cause false failures or unbounded traversal. Constrain discovered targets to the canonical project root and avoid descending through directory symlinks that resolve outside it.

HAYDEN-OAI (commented):
Found one source-discovery regression that allows directory symlinks to escape the project root; the inline comment includes an exact-head reproduction and fix direction.

jbeckwith-oai (approved):
Reviewed the exact diff and RuboCop 1.81.7 target-discovery behavior. Removing the git ls-files intersection makes RuboCop TargetFinder the single source of truth, correctly covers no-Git and untracked sources, preserves configured includes/excludes and path behavior, and improves the tests by exercising the real CLI. Strict thermo is clean and exact CI is green.

PR #432, opened 2026-08-17
fix: attach partial completion to streaming length errors

thegoodengineer:
## Summary

`OpenAI::LengthFinishReasonError` exposes a `completion` reader so callers can recover the content the model produced before it ran into its token limit. When the error is raised from a streaming chat completion, that reader is always `nil`.

`ChatCompletionStreamState#handle_finish_reason` builds the error from `@chat_completion`:

```ruby
# lib/openai/helpers/streaming/chat_completion_stream.rb:276
def handle_finish_reason(finish_reason)
  return unless parseable_input?

  case finish_reason
  when :length
    raise LengthFinishReasonError.new(completion: @chat_completion)
```

`@chat_completion` is never assigned anywhere in the SDK, so it always reads as `nil`:

```console
$ grep -rn "@chat_completion" lib rbi sig test
lib/openai/helpers/streaming/chat_completion_stream.rb:281:            raise LengthFinishReasonError.new(completion: @chat_completion)
```

That single read is the only occurrence in the repository.

## Reproduction

No network needed, the transport is stubbed:

```ruby
require "openai"

class Person < OpenAI::BaseModel
  required :name, String
  required :age, Integer
end

SSE = <<~SSE
  data: {"id":"chatcmpl-123","object":"chat.completion.chunk","created":1234567890,"model":"gpt-4o-mini","choices":[{"index":0,"delta":{"role":"assistant","content":"{\\"name\\":"},"finish_reason":null}]}

  data: {"id":"chatcmpl-123","object":"chat.completion.chunk","created":1234567890,"model":"gpt-4o-mini","choices":[{"index":0,"delta":{"content":"\\"John\\","},"finish_reason":null}]}

  data: {"id":"chatcmpl-123","object":"chat.completion.chunk","created":1234567890,"model":"gpt-4o-mini","choices":[{"index":0,"delta":{},"finish_reason":"length"}]}

  data: [DONE]

SSE

class StubTransport < OpenAI::HTTPClient
  def execute(_request)
    OpenAI::HTTPClient::Response.new(
      status: 200,
      headers: {"content-type" => "text/event-stream"},
      body: [SSE.b]
    )
  end
end

client = OpenAI::Client.new(api_key: "x", base_url: "http://localhost", http_client: StubTransport.new)

begin
  client.chat.completions
        .stream(model: "gpt-4o-mini", messages: [{role: "user", content: "hi"}], response_format: Person)
        .each { |_event| nil }
rescue OpenAI::LengthFinishReasonError => e
  puts "completion: #{e.completion.inspect}"
end
```

Before:

```
completion: nil
```

After:

```
completion: #<OpenAI::Models::Chat::ParsedChatCompletion:0x118 {id: "chatcmpl-123", choices: [{finish_reason: :length, index: 0, logprobs: nil, message: {content: "{\"name\":\"John\",", role: :assistant}}], created: 1234567890, model: "gpt-4o-mini", object: :"chat.completion", service_tier: nil, system_fingerprint: nil, usage: nil}>
```

The truncated content is the useful part of this error. Without it a caller that hits the token limit has no way to reach what was already generated, because the exception unwinds the stream before `get_final_completion` can be called.

## Fix

`accumulate_choice!` already holds the accumulated `completion_snapshot`, so it is threaded into `handle_finish_reason` and used to build the error.

This restores parity with two places that already do the right thing:

- `openai-python` raises `LengthFinishReasonError(completion=completion_snapshot)` from the equivalent branch of `ChatCompletionStreamState._accumulate_chunk`.
- The non-streaming path in this same file, `parse_chat_completion`, already passes the completion it has.

`ContentFilterFinishReasonError` takes no completion argument, so it is unchanged.

## Tests

- `test_length_finish_reason_exposes_partial_completion` is new and covers the `response_format` path.
- `test_length_finish_reason_with_strict_tool_raises` previously asserted only that the error was raised. It now also asserts the completion carries the accumulated tool call, which covers the `input_tools` path.

Both fail against the unpatched `chat_completion_stream.rb` and pass with it:

```
# new tests, old lib
test_length_finish_reason_exposes_partial_completion    1 runs, 2 assertions, 1 failures, 0 errors, 0 skips
test_length_finish_reason_with_strict_tool_raises       1 runs, 2 assertions, 1 failures, 0 errors, 0 skips

# this branch
test_length_finish_reason_exposes_partial_completion    1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
test_length_finish_reason_with_strict_tool_raises       1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
```

The rest of `test/openai/resources/chat/completions/streaming_test.rb` passes unchanged.

`rubocop` and `scripts/validate-rubocop-directives` are clean on the changed files. No RBI or RBS signatures reference `handle_finish_reason`, so none needed updating.


jbeckwith-oai (approved):
Thanks @thegoodengineer — this is a clear, focused fix. I verified that the base implementation reads an unassigned `@chat_completion`, reproduced the nil completion before the change, and confirmed the accumulated snapshot is attached after it. The focused streaming tests pass on CRuby 3.3, 3.4, and 4.0, along with the adjacent snapshot tests, RuboCop, Sorbet, and RBS validation. The change is idiomatic and surgical, introduces no security or public API concerns, and requires no RBI/RBS or Castiron updates. Approved!

thegoodengineer:
Thank you @jbeckwith-oai! Really appreciate the review and approval.

PR #436, opened 2026-08-17
chore: enforce redundant enable directives

jbeckwith-oai:
## Summary

- enable `Lint/RedundantCopEnableDirective`
- keep the existing temporary allowance for narrow disable directives
- reject unmatched or otherwise unnecessary `rubocop:enable` comments

## Why

The formatting migration still makes some narrow disable directives harmless, but their closing enable directives should remain structurally valid. This separates the two policies instead of disabling both directive checks together.

## Ratchet evidence

- Rule: `Lint/RedundantCopEnableDirective`
- Baseline: 0 offenses across 2,632 current Ruby/RBI files
- Final: 0 offenses; no suppressions or exclusions added
- Regression proof: a temporary untracked Ruby file with an unmatched `rubocop:enable Lint/EmptyBlock` was rejected by the cop, then removed before commit

## Castiron impact

No companion PR is needed. Castiron's Ruby renderer emits one paired `Lint/RedundantRequireStatement` disable/enable region in the generated root file, and its existing renderer fixture covers that region. Current generated output is clean; this SDK-owned config change cannot be overwritten by regeneration, while assembled-SDK validation already runs the repository RuboCop task.

## Validation

- focused `Lint/RedundantCopEnableDirective`: clean
- full `rake lint`: RuboCop clean, directive guard clean, Sorbet clean, 1,214 RBS files validated
- gem package build: successful
- `git diff --check`: clean
- thermo-nuclear maintainability review: no findings

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -39,8 +39,9 @@ Lint/UselessAssignment:
 # The directive guard still rejects broad or unowned suppressions.
 Lint/RedundantCopDisableDirective:
   Enabled: false
+# Every enable directive must still close an active disable.
 Lint/RedundantCopEnableDirective:
-  Enabled: false
+  Enabled: true
 
 # Dependency ordering is layout, not package correctness.
 Bundler/OrderedGems:


HAYDEN-OAI (approved):
Reviewed the directive-enforcement path, suppression guard, SDK generation boundaries, and packaged gem contents. This change remains confined to development configuration, preserves existing narrow suppressions, and does not alter generated SDK code, runtime behavior, public APIs, or shipped dependencies.

.rubocop.yml
```diff
@@ -39,8 +39,9 @@ Lint/UselessAssignment:
 # The directive guard still rejects broad or unowned suppressions.
 Lint/RedundantCopDisableDirective:
   Enabled: false
+# Every enable directive must still close an active disable.
 Lint/RedundantCopEnableDirective:
-  Enabled: false
+  Enabled: true
```
HAYDEN-OAI, line 44: **[P1] Disable autocorrection for the newly enabled directive cop**

RuboCop is locked to 1.81.7, where `Lint/RedundantCopEnableDirective` deletes only `DirectiveComment#range`; that range includes the leading `#` but excludes a supported trailing `-- reason`. Consequently, normal safe autocorrection rewrites `# rubocop:enable Lint/EmptyBlock -- puts(42)` into `-- puts(42)`, which Ruby parses as executable `puts(42)` rather than a comment; ordinary reason text can instead leave invalid Ruby. Enabling the cop here therefore introduces a silent code-execution/control-flow change in editor or `rubocop -a` workflows that was unavailable while the cop was disabled. Add `AutoCorrect: false` alongside `Enabled: true` until the upstream fix lands: https://github.com/rubocop/rubocop/pull/15561.

HAYDEN-OAI (commented):
Correcting my earlier approval after confirming a substantive unsafe-autocorrection regression in the pinned RuboCop release.

PR #438, opened 2026-08-17
feat: Add client.with_options support

apcha-oai:
## Summary

Python and Node already let callers derive a client with a different base URL while retaining the original configuration. Ruby was the only one of these SDKs missing that flow: callers had to construct another client and repeat its settings. Add `Client#with_options` to close that gap without mutating a shared client.

| SDK | Select the EU endpoint | Support |
| --- | --- | --- |
| Python | `client.with_options(base_url="https://eu.api.openai.com/v1")` | Existing |
| Node | `client.withOptions({ baseURL: "https://eu.api.openai.com/v1" })` | Existing |
| Ruby | `client.with_options(base_url: "https://eu.api.openai.com/v1")` | Added here |

Include a standalone EU-routing example for eligible projects. This is a client-configuration feature, not a new API request parameter; a residency shorthand remains out of scope.

## Implementation details

The derived client rebuilds its resources, inherits resolved settings, and shares the HTTP transport unless replaced. Credential and provider changes discard incompatible inherited configuration. The example runs against a mock transport in ordinary CI, and the existing `Response#output_text` helper now has RBI and RBS declarations.

`with_options` preserves the receiver’s dynamic class and returns the self type, so subclass request/authentication behavior is retained. The chosen tradeoff is that subclasses with extra required constructor state or a private `.new` must override `with_options` if they want to support copying. We deliberately propagate those constructor errors instead of silently returning a base `OpenAI::Client`; no separate copy-factory hook is added until a concrete use case needs one.


apcha-oai:
@codex review

lib/openai/helpers/client_options.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          headers = options.fetch(:default_headers)
+          if overrides.key?(:default_headers)
+            supplied = overrides[:default_headers]
+            headers = supplied.nil? ? {} : OpenAI::Internal::Util.normalized_headers(headers, supplied)
+          end
+
+          options.merge!(overrides)
+          options[:default_headers] = ResolvedHeaders.new.replace(headers)
+          OpenAI::Client.new(**options)
```
jbeckwith-oai: Medium: This always constructs OpenAI::Client, so calling the inherited with_options on a subclass silently drops overridden auth_headers, prepare_request, instrumentation/error behavior, specialized resources, and class-sensitive headers. That also diverges from the Python/Node parity claimed here, which preserve the dynamic client type. Construct through the receiver dynamic class or a protected overridable copy factory, type the result as the self type in RBI/RBS, and let subclasses with extra constructor state override the factory; returning an unrelated base instance is not a safe fallback.

test/openai/client_options_test.rb
```diff
@@ -0,0 +1,235 @@
...
+    assert_equal(GLOBAL, client.base_url.to_s)
+    assert_equal(EU, eu.base_url.to_s)
+    assert_same(@transport, eu.requester)
+    refute_same(client.responses, eu.responses)
+    assert(@transport.requests.all? { _1.headers["authorization"] == "Bearer test-key" })
+  end
+
+  def test_copy_without_overrides_preserves_every_constructor_option
+    logger = Logger.new(StringIO.new)
+    callback = ->(_event) {}
```
jbeckwith-oai: Low: Please use an explicit no-op body here, e.g. ->(_event) { nil }, rather than an empty block. This keeps the test compliant without a Lint/EmptyBlock suppression.

jbeckwith-oai (changes requested):
The option-copying behavior is otherwise well covered, but with_options currently breaks inherited client behavior by silently returning a base client. I also left one small test-style comment.

jbeckwith-oai (approved):

PR #439, opened 2026-08-17
chore(deps): configure Ruby and GitHub Actions Dependabot

HAYDEN-OAI:
## Summary

- Add the repository's first `.github/dependabot.yml` for the root Bundler dependency graph (`Gemfile`, `Gemfile.lock`, and `openai.gemspec`) and all GitHub Actions workflows.
- Run weekly Monday updates at 09:00/09:30 UTC with the same eight-day release-age cooldown and five-open-PR cap used by the sibling SDKs.
- Group minor/patch production and development Ruby gems separately; leave major upgrades and release-sensitive Actions independently reviewable while keeping paired CodeQL actions together.
- Existing security updates remain enabled and are not delayed by version-update cooldowns or PR limits ([GitHub Dependabot options reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown)).

## Verification

- Strict duplicate-key-safe YAML parsing and JSON Schema validation derived from current GitHub-supported Dependabot options.
- Repository-specific assertions verified actual Bundler manifests, both supported ecosystems, weekly UTC schedules, cooldown/PR limits, production/development grouping, separately reviewable major/security updates, five workflow files, SHA-pinned Actions, and the exact paired CodeQL pattern.
- `git diff origin/main...HEAD --check`
- Two independent read-only reviews found no configuration or release-safety issues; only `.github/dependabot.yml` changed.

Ruby/Bundler are unavailable locally; no runtime code, dependencies, lockfiles, or workflows changed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,14 +26,16 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: weekly
-      day: monday
+      interval: daily
       time: "09:30"
       timezone: Etc/UTC
     cooldown:
       default-days: 8
+      # SHA-pinned actions receive version updates, not security alert updates.
+      exclude:
+        - "*"
     groups:
       codeql:
         patterns:
-          - "github/codeql-action/*"
+          - "github/codeql-action"
     open-pull-requests-limit: 5
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,39 @@
+version: 2
+
+updates:
+  - package-ecosystem: bundler
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+      time: "09:00"
+      timezone: Etc/UTC
+    cooldown:
+      default-days: 8
+    groups:
+      production-dependencies:
+        dependency-type: production
+        update-types:
+          - minor
+          - patch
+      development-dependencies:
+        dependency-type: development
+        update-types:
+          - minor
+          - patch
+    open-pull-requests-limit: 5
+
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+      time: "09:30"
+      timezone: Etc/UTC
+    cooldown:
+      default-days: 8
+    groups:
+      codeql:
+        patterns:
+          - "github/codeql-action/*"
+    open-pull-requests-limit: 5


.github/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+      time: "09:30"
+      timezone: Etc/UTC
+    cooldown:
+      default-days: 8
```
jbeckwith-oai, line 33: Medium: Every external action here is pinned to a full SHA, and GitHub does not create vulnerability alerts/security-update PRs for SHA-pinned Actions. Their only automated path to a fixed release is therefore a version update, which this cooldown plus the weekly schedule delays roughly 8 to just under 15 days. Security PRs are technically exempt; the issue is that SHA pins never get one. To bypass release-age delay, keep a GitHub Actions cooldown block but add exclude: ["*"] (removing the block alone still leaves GitHub's default cooldown), and use a more frequent schedule if sub-week response is required. The Bundler cooldown can remain.

.github/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+    groups:
+      codeql:
```
jbeckwith-oai, line 38: Low: Dependabot collapses composite action usages such as github/codeql-action/init and /analyze to dependency name github/codeql-action, so this slash-wildcard pattern never matches and the named group is dead. Use github/codeql-action (or github/codeql-action*) instead. Same-repo occurrences still update together today, so this is configuration correctness rather than a runtime break.

jbeckwith-oai (changes requested):
The Bundler setup is sound, but the GitHub Actions entry does not currently provide the claimed security-update behavior and its CodeQL grouping rule is ineffective.

.github/dependabot.yml
HAYDEN-OAI: Fixed in 6908fc6. The CodeQL group now matches Dependabot’s canonical dependency name, `github/codeql-action`, so both init and analyze usages remain aligned under the intended group. YAML and targeted grouping/security policy checks pass.

.github/dependabot.yml
HAYDEN-OAI: Fixed in 6908fc6. GitHub Actions now run daily and explicitly exclude every action from release-age cooldown, so full-SHA-pinned actions receive timely version-update PRs despite GitHub not issuing vulnerability alerts/security-update PRs for SHA pins. The Bundler schedule and eight-day cooldown remain unchanged. GitHub documents the SHA-pin limitation here: https://docs.github.com/en/actions/reference/security/secure-use#monitoring-the-actions-in-your-workflows

jbeckwith-oai (approved):
Independent two-pass re-review confirms prior Dependabot concerns are fixed: GitHub Actions updates run daily with explicit cooldown exclusion, and CodeQL grouping matches the canonical dependency name. Bundler grouping retains the desired routine-update cooldown without delaying security updates. The actual Gemfile, lockfile, pinned workflows, and substantive CI are consistent.

PR #442, opened 2026-08-18
feat: update streaming models and regional endpoint selection

apcha-oai:
## Summary

Expose shell-call streaming events and their WebSocket variants, plus chat-completion metadata, image-generation metadata, and the new audit-log event type.

Keep inherited RBS and Sorbet signatures in sync with the Ruby models. Previously, a WebSocket subclass accepted inherited image fields at runtime while its generated constructor and `to_hash` signatures omitted them. The regenerated signatures include those fields, including `background`, `output_format`, `quality`, and `size`, without changing runtime inheritance.

Also add `data_residency` for selecting the `global`, `us`, `eu`, or `ae` OpenAI endpoint. It works with copied clients and rejects conflicting explicit endpoints or third-party providers; it does not change project eligibility or provide regional fallback.


sig/openai/models/responses/response_image_gen_call_partial_image_event.rbs
```diff
@@ -24,12 +28,32 @@ module OpenAI
...
+
+        def size=: (String) -> String
+
         def initialize: (
           item_id: String,
           output_index: Integer,
           partial_image_b64: String,
           partial_image_index: Integer,
           sequence_number: Integer,
+          ?background: String,
```
HAYDEN-OAI, line 53: [P2] Regenerate inherited image fields in WebSocket event signatures

Adding `background`, `output_format`, `quality`, and `size` here updates the base image event's public signature, but the existing `ResponsesServerEvent::ResponseImageGenCallPartialWsImage` and `BetaResponsesServerEvent::BetaResponseImageGenCallPartialWsImage` subclasses were not regenerated. At the exact PR head, real Ruby plus RBS/RBI parsing confirms both WebSocket subclasses accept and expose all four fields at runtime while their RBS/RBI constructors advertise none; their `to_hash` override shapes also omit fields newly present in the parent signature. Exact-base runtime and signatures agree, so this is introduced here. Sorbet/RBS consumers therefore reject valid WebSocket-event keywords, and the subclass hash override no longer matches its parent. Regenerate the stable and beta `responses_server_event.{rbs,rbi}` WebSocket image subclasses, including constructor keywords and `to_hash` shapes.

HAYDEN-OAI (commented):
Found one reproducible P2 public API compatibility regression affecting stable and beta WebSocket image-event type signatures; details and remediation are in the inline comment.

sig/openai/models/responses/response_image_gen_call_partial_image_event.rbs
apcha-oai: Stainless bug, fixed in the castiron compiler

jbeckwith-oai (approved):
Reviewed the complete generated update and its compatibility boundaries. Regional endpoint selection, provider conflicts, environment precedence, client copying and subclass behavior are covered. Shell-call streaming variants and metadata remain coherent, and stable/beta inherited image fields are now present in both RBS and Sorbet signatures. Exact-head Ruby 3.3/3.4/4.0, lint, packaging, RBI/RBS, and security checks are green. No actionable findings.

PR #444, opened 2026-08-18
chore: synchronize generated response signatures and tests

jbeckwith-oai:
## Summary

- Generate inherited WebSocket event fields in RBI/RBS constructor and `to_hash` signatures for both beta and stable Responses namespaces.
- Remove obsolete formatting-only customizations from four generated union resource tests; their position-independent Ruby ASTs are unchanged.
- Preserve runtime Ruby model declarations and direct property accessors unchanged.
- Refresh generated annotation typing and source documentation while reusing the exact pinned API and configuration inputs.

## Validation

- Castiron Ruby renderer unit suite: 578 passed
- Internal candidate scope: 4 signature files, 4 generated tests, and `.castiron.stats.yml`
- OpenAPI and Stainless configuration hashes are unchanged from the prior generation

### DIFF
--- a/.castiron.stats.yml
+++ b/.castiron.stats.yml
@@ -1,6 +1,6 @@
 schema_version: 1
-generation_id: e0a62312-b47a-4bc8-a154-1b0ff22ba46b
+generation_id: 2eec67ae-b82c-4e77-b588-32365f621ee7
 openapi_spec_hash: cc58e691f33751629ca4de3100082edb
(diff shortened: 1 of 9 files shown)


HAYDEN-OAI (approved):
Reviewed exact head 6e92ff9dcf969b52a85d4e276fb38535b0d311ad. The beta and stable WebSocket annotation signatures now match their inherited nullable union contracts; runtime SDK code is unchanged, and the generated resource changes are whitespace-only. No substantive issues found.

Issue #376, opened 2026-08-12
Track model conversion correctness after #295 rollback


jbeckwith-oai:
## Context

PR #375 intentionally reverts #295 to restore the published v0.78 raw-request / parsed-response `BaseModel` contract before redesigning nested coercion. The exact rollback also restores several pre-#295 converter behaviors that deserve independent investigation and regression coverage.

These concerns must not be fixed by widening global coercion or changing `BaseModel#[]` / `#to_h` away from raw caller-owned values. Each should be reproduced against the published contract and solved narrowly.

## Follow-up behaviors

- [ ] Isolate conversion errors between union variant attempts so a failed earlier variant cannot poison a later successful match (for example Realtime `max_output_tokens: :inf`).
- [ ] Preserve explicitly allowed `nil` values without recording a conversion error; distinguish `optional` (omittable) from `nil?: true` (nullable) deliberately.
- [ ] Preserve the incoming strictness state across the discriminated-union fast path before evaluating later fields.
- [ ] Return an already-coerced instance of the target `BaseModel` unchanged, retaining identity and out-of-band state such as `last_response`.
- [ ] Retain collection element conversion failures deterministically across `ArrayOf` and `HashOf`, independent of element order.
- [ ] Make nullable `ArrayOf` / `HashOf` matchers agree with their coercion and generated type contracts.

## Required coverage

For each behavior, add a minimal runtime-level regression and at least one representative generated-model test where applicable. Include order permutations for collection/union error state, symbol/string discriminator forms where relevant, and assertions covering raw request access separately from parsed response materialization.

## Related

- #295 introduced the global coercion change now being rolled back.
- #375 is the standalone rollback and intentionally does not solve these follow-ups.
- #287 remains the separate user-facing unknown-stream-event issue.

PR #446, opened 2026-08-18
fix: isolate union coercion state

jbeckwith-oai:
## Summary

- isolate conversion errors between union variant attempts so rejected variants cannot poison the selected result
- preserve the incoming strictness state across the discriminated-union fast path
- preserve a pre-existing parent error unless the selected variant reports a more specific error
- add runtime regression coverage plus a representative Realtime generated-model contract test

Addresses the first and third follow-up behaviors in #376. This intentionally does not close the tracking issue.

## Root cause

Union alternatives shared `state[:error]`, so an error from a rejected variant survived into a later successful match. The discriminated fast path also returned before capturing `state[:strictness]`, while the method's `ensure` still restored that uninitialized local.

The fix adds an internal, typed converter helper that isolates each attempt's error, then commits only the chosen variant's error. Strictness is captured before discriminator resolution.

## Generated-code ownership

No Castiron generator or generated model changes are required. The behavior belongs to SDK-owned internal coercion plumbing; the Realtime model is used only as representative regression coverage. Raw request values remain caller-owned, while parsed response materialization still normalizes `"inf"` / `:inf` to `:inf`.

## Validation

- focused regressions: 37 runs, 367 assertions, 0 failures/errors
- full suite: 834 runs, 4,176 assertions, 0 failures/errors, 1 skip
- Sorbet: no errors
- RBS: 1,219 files valid
- RuboCop: 2,648 files inspected, no offenses
- generated directive validation: passed
- `git diff --check`: passed

Note: `./scripts/test` exits 1 after the green suite because its cleanup trap attempts to kill an already-exited mock-server PID (`kill: No such process`).

### DIFF
--- a/lib/openai/internal/type/converter.rb
+++ b/lib/openai/internal/type/converter.rb
@@ -270,6 +270,26 @@ module OpenAI
 
           # @api private
           #
+          # Coerces a value while isolating its error from sibling coercions.
+          #
+          # @param target [OpenAI::Internal::Type::Converter, Class]
+          #
+          # @param value [Object]
+          #
+          # @param state [Hash{Symbol=>Object}]
+          #
+          # @return [Array(Object, StandardError, nil)]
+          def coerce_with_error(target, value, state:)
+            previous_error = state.fetch(:error)
+            state[:error] = nil
+            coerced = coerce(target, value, state: state)
+            [coerced, state.fetch(:error)]
+          ensure
+            state[:error] = previous_error
+          end
+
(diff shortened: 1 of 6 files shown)


HAYDEN-OAI (approved):
Reviewed exact commit 4daec6b7844e73ee5992cc7591de02b9ab06b340, including union error-state isolation, strictness restoration, generated model contracts, and current-main Realtime integration. No substantive issues found.

PR #450, opened 2026-08-18
fix: preserve nullable nil values

jbeckwith-oai:
## Summary

- treat explicit `nil` as a valid setter value only when the field declares `nil?: true`
- avoid recording a latent conversion error for nullable request values while preserving raw `#[]`, `#to_h`, accessor, and request-dump behavior
- distinguish an omitted optional field from explicitly assigning `nil` to an optional non-nullable field
- add focused runtime coverage and a representative generated `BatchCreateParams` contract test

Addresses the second follow-up behavior in #376. This intentionally does not close the tracking issue.

## Root cause

`BaseModel.coerce` already recognizes nullable `nil` during parsed-response materialization, but generated-model setters always invoked the declared converter first. For nullable scalar, collection, and model fields, that converter recorded a `TypeError`; the raw value remained `nil`, but the typed accessor later raised a `ConversionError`.

The setter now records nullable `nil` as a valid raw value without invoking its non-nil target converter.

## Compatibility boundary

This change is intentionally limited to declared nullability in setters. Optional means a request field may be omitted; it does not make explicit request `nil` valid without `nil?: true`. Existing parsed-response handling of present optional `nil` remains an inexact compatibility match—tightening that path breaks Responses and Chat streaming accumulators and is outside this follow-up.

## Generated-code ownership

No Castiron generator or generated-model changes are required. Nullability metadata is already emitted correctly; the defect is in SDK-owned `BaseModel` setter plumbing. `BatchCreateParams` is used only as representative generated-model coverage.

## Validation

- focused runtime/generated regressions: 6 runs, 30 assertions, 0 failures/errors
- existing BaseModel suite: 22 runs, 286 assertions, 0 failures/errors
- affected Responses/Chat streaming suites: 25 runs, 95 assertions, 0 failures/errors
- raw-value and prior union regressions: 14 runs, 69 assertions, 0 failures/errors
- full suite: 916 runs, 4,852 assertions, 0 failures/errors, 1 skip
- Sorbet: no errors
- RBS: 1,236 files valid
- RuboCop: 2,700 files inspected, no offenses
- rubyfmt and directive validation: passed
- `git diff --check`: passed

## Related

- #446 fixed union-attempt error isolation and discriminated-union strictness.

### DIFF
--- a/lib/openai/internal/type/base_model.rb
+++ b/lib/openai/internal/type/base_model.rb
@@ -78,18 +78,23 @@ module OpenAI
             }
 
             define_method(setter) do |value|
-              target = type_fn.call
-              state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
-              coerced = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
-              error = @coerced.store(name_sym, state.fetch(:error) || true)
-              stored = case [target, error]
-              in [OpenAI::Internal::Type::Converter | Symbol, nil]
-                coerced
+              if value.nil? && nilable
+                @coerced.store(name_sym, true)
+                @data.store(name_sym, nil)
               else
-                value
-              end
+                target = type_fn.call
+                state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
+                coerced = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
+                error = @coerced.store(name_sym, state.fetch(:error) || true)
+                stored = case [target, error]
+                in [OpenAI::Internal::Type::Converter | Symbol, nil]
+                  coerced
+                else
+                  value
+                end
 
-              @data.store(name_sym, stored)
+                @data.store(name_sym, stored)
+              end
             end
 
             # rubocop:disable Style/CaseEquality
--- a/lib/openai/internal/type/base_model.rb
+++ b/lib/openai/internal/type/base_model.rb
@@ -78,7 +78,7 @@ module OpenAI
             }
 
             define_method(setter) do |value|
-              if value.nil? && nilable
+              if nil.equal?(value) && nilable
                 @coerced.store(name_sym, true)
                 @data.store(name_sym, nil)
               else
--- a/test/openai/internal/type/base_model_nilability_test.rb
+++ b/test/openai/internal/type/base_model_nilability_test.rb
@@ -32,6 +32,22 @@ class OpenAI::Test::BaseModelNilabilityTest < Minitest::Test
     assert_instance_of(TypeError, error.cause)
   end
 
+  def test_constructor_treats_only_the_nil_singleton_as_nullable_nil
+    value = {"nil?" => true, "effort" => "high"}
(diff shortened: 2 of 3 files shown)


lib/openai/internal/type/base_model.rb
```diff
@@ -78,18 +78,23 @@ def fields
...
 
             define_method(setter) do |value|
-              target = type_fn.call
-              state = OpenAI::Internal::Type::Converter.new_coerce_state(translate_names: false)
-              coerced = OpenAI::Internal::Type::Converter.coerce(target, value, state: state)
-              error = @coerced.store(name_sym, state.fetch(:error) || true)
-              stored = case [target, error]
-              in [OpenAI::Internal::Type::Converter | Symbol, nil]
-                coerced
+              if value.nil? && nilable
```
HAYDEN-OAI: [P2] Use actual nil identity for the nullable shortcut

`nil?` is overridable by ordinary standard-library JSON objects, so this branch can mistake a real request value for nil and overwrite the raw value. For example, `value = JSON.parse('{"nil?":true,"effort":"high"}', object_class: JSON::GenericObject)` is not nil but returns `value.nil? == true`. Passing it to `OpenAI::Models::Responses::InputTokenCountParams.new(reasoning: value)` produces `{"reasoning":{"nil?":true,"effort":"high"}}` from `dump_request` on the merge base, but this head silently changes the wire payload to `{"reasoning":null}` and `params[:reasoning]` also loses the original object. `OpenStruct` reproduces the same issue with user-controlled keys. Check identity with `nil.equal?(value)` before taking this shortcut so only the actual nil singleton can erase the stored value.

HAYDEN-OAI (commented):
One merge-base-verified P2 compatibility/data-integrity regression in the nullable setter; see the inline reproduction.

lib/openai/internal/type/base_model.rb
jbeckwith-oai: Fixed in `60419a20db310415c50eae2ec7dbd96584060382`: the nullable shortcut now checks `nil.equal?(value)`, so objects with an overridden `nil?` predicate remain on the normal coercion/raw-value path. I added a regression test using a Hash whose `nil?` returns true and verified object identity through `[]`/`to_h` plus the serialized JSON wire payload. Validation: focused nilability 5 runs/21 assertions, related generated-model and raw-value contract suites 11 runs/52 assertions, full suite 917 runs/4,856 assertions (0 failures/errors, 1 skip), and full lint/type/RBS checks clean.

HAYDEN-OAI (approved):
Re-reviewed exact head 60419a20db310415c50eae2ec7dbd96584060382 against the merge base and previously reviewed head. The nil.equal?(value) follow-up fixes spoofable nil?/proxy data loss while preserving raw object identity and JSON wire payloads. Exercised all 1,804 nullable fields and 2,430 optional non-nullable fields across 2,465 SDK model classes, plus recursive models, collections/unions, generated setters, omitted versus explicit nulls, adversarial Ruby objects, and Sorbet/RBI compatibility. No substantive findings.

PR #451, opened 2026-08-18
chore(api): enforce signature formatting and canonicalize generated Ruby

apcha-oai:
## Summary

Close the signature-formatting gap in the existing Ruby CI lint job. Rubyfmt 0.14.1 now owns Ruby source and RBI layout; RBS keeps the existing pinned Syntax Tree formatter. Both formats are checked by the current `lint:rubocop` task, without adding a CI job or changing workflow permissions.

- Remove Syntax Tree's competing RBI pass and retain `format:rbi` as a compatibility task.
- Share one RBS implementation between formatting and non-mutating check mode, preserving the class/module-alias workaround.
- Normalize the existing signatures and bound one RBI RuboCop suppression that rubyfmt moves off its original line.
- Regenerate Castiron's pure-codegen checkpoint so the formatter migration is not recorded as new custom code.
- Keep Castiron's formatter-processed output canonical rather than adding renderer workarounds to preserve the old Syntax Tree RBI layout.

Ruby-only formatter migration; the transformed API reference is unchanged.

## Custom code before / after

The custom-code cleanup was audited and committed as a separate second pass before promotion.

Against the same canonical codegen checkpoint (`73d61ce694e4030efcd4b72016b86db0ffd5e133`), the 3,683 generator-owned files contain **49 → 47 mixed/custom files** and **3,917 → 2,772 added-plus-deleted custom-diff lines** (`git diff --numstat`). Handwritten-only files are outside that denominator. Ten files were simplified; the two `responses_server_event.rbs` files now exactly match codegen. The other eight retain their real customizations while dropping equivalent layout or syntax differences.

Each accepted Ruby/RBI replacement preserves the normalized Ruby AST and comment/directive sequence. RBS replacements preserve parsed declarations and comments, ignoring source locations and keyword-map ordering. The remaining 47 mixed files contain meaningful behavior, types, or documentation differences under that audit; no whole-file formatting-only matches remain. In particular, client/provider behavior, structured-output converters, streaming/polling/webhook helpers, useful documentation, and existing type/name compatibility changes are preserved.

## Validation

- Before formatting, `lint:rubyfmt`, `lint:rbs_format`, and the existing `lint:rubocop` task each rejected the current public-main signatures. The checks left those files unchanged.
- The shared RBS implementation matches the old formatter's output across all 1,236 RBS files.
- Formatter-policy tests: 9 runs, 190 assertions, no failures.
- `bundle exec rake lint:rubocop`: 2,698 files inspected, no offenses.
- `bundle exec rake typecheck:sorbet validate:rbs`: Sorbet passed; all 1,236 RBS files validated.
- Parsed-signature comparison against public main: 1,229 changed RBI files and 11 changed RBS files retain the same meaning.
- Focused formatter, Responses, and webhook tests through the SDK mock-server runner: 31 runs, 239 assertions, no failures or errors.
- Repeated the full Ruby-only Castiron generation from the same pinned source after the cleanup: all 3,683 pure-generated files were byte-identical, no merge conflicts recurred, and the only resulting SDK change was `.castiron.stats.yml`. Promotion combines the verified internal commits into one public commit.

## Rollout

This is intentionally a larger, one-time signature-layout migration. Rubyfmt is already pinned to 0.14.1 with checksum-verified installation, and the pinned monorepo includes the Castiron Ruby-runtime selection fix. No new formatter dependency, renderer compatibility shim, or immutable Stainless-parity change is needed. The complete candidate passed internal CI before promotion.


jbeckwith-oai (approved):
Independent two-pass review covered the complete 1,248-file change, representative Ruby/RBI syntax-tree equivalence, and all 11 RBS signature transformations. Runtime security-sensitive code remains semantically identical; formatter ownership, aliases, nonmutating checks, pinned/checksummed rubyfmt, package contents, generated boundaries, and CI are consistent. No actionable findings.

PR #452, opened 2026-08-18
chore: remove Tapioca development dependency

jbeckwith-oai:
## Summary

- remove the direct Tapioca development dependency and its tooling-only transitive closure
- declare the existing `sorbet-runtime` test requirement directly, retaining version `0.6.12690`
- remove the orphaned repository-only `sorbet/tapioca` directory task and clobber target
- preserve the runtime compatibility guard for downstream consumers that run Tapioca against the gem
- explicitly skip Sorbet's bundled-gem RBI discovery because this repository already typechecks its shipped `rbi/` directory

## Why

The repository ships and directly typechecks its RBI files, so it does not use Tapioca to generate repository signatures. The test suite does directly exercise runtime Sorbet conversion, which was previously available only through Tapioca's transitive dependency graph; this PR makes that requirement explicit.

The resulting dependency graph keeps 21 direct dependencies while replacing Tapioca with the actual test dependency, and reduces locked specs from 104 to 95.

Sorbet's launcher implicitly suppresses bundled-gem RBI discovery while Tapioca is present in the lockfile. Without an explicit replacement, removing Tapioca loads this gem's shipped RBIs twice. The scoped `SRB_SKIP_GEM_RBIS=1` setting preserves the existing typecheck behavior.

## Validation

- `bundle exec rake test TEST=test/openai/internal/type/array_of_sorbet_type_test.rb` — 1 run, 2 assertions, 0 failures
- `bundle exec rake test TEST=test/openai/load_order_test.rb` — 3 runs, 3 assertions, 0 failures
- `bundle check`
- `bundle exec rubocop Gemfile Rakefile`
- `bundle exec rake lint:rubyfmt`
- thermo-nuclear code quality review after the CI fix — no findings

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ group :development, :test do
   gem "minitest-hooks"
   gem "minitest-proveit"
   gem "minitest-rg"
+  gem "sorbet-runtime"
   gem "webmock"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,6 @@ group :development do
   gem "steep"
   gem "syntax_tree"
   gem "syntax_tree-rbs", github: "ruby-syntax-tree/syntax_tree-rbs", branch: "main"
-  gem "tapioca"
 end
 
 group :development, :test do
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Reviewed the exact head across the dependency closure, Ruby compatibility, shipped RBI/generator ownership, package/release boundaries, and the locked Sorbet launcher's Tapioca detection. No substantive correctness, compatibility, or supply-chain issues found.

PR #455, opened 2026-08-19
Remove runtime CGI dependency

jbeckwith-oai:
## Summary

- replace `CGI.parse` in internal query decoding with a small `URI`-based implementation
- preserve duplicate keys, decoding, semicolon separators, valueless keys, empty segments, malformed percent escapes, binary/invalid UTF-8 bytes, and the frozen non-inserting hash default
- remove `cgi` from the gem runtime load/dependency/manifest graph while retaining it as a development/test dependency for existing multipart tests

## Consumer graph

The built gem's direct runtime dependencies drop from four to three:

- before: `base64`, `cgi`, `connection_pool`, `logger`
- after: `base64`, `connection_pool`, `logger`

## Validation

- `bundle check`
- focused util, client options, path-parameter query, and path-encoding tests: 60 runs / 272 assertions
- package test: 1 run / 8 assertions
- built-gem runtime dependency inspection
- Sorbet and RBS validation
- RuboCop, rubyfmt, RBS formatting, and suppression-directive checks
- full non-live suite excluding one sandbox-specific Realtime proxy test: 926 runs / 4,893 assertions / 0 failures / 0 errors
- 1,111,111 exhaustive query-parser differential cases against `CGI.parse`
- 500,000 randomized byte-query differential cases across UTF-8, binary, US-ASCII, ISO-8859-1, and Shift_JIS

The complete local runner consistently reaches 927 runs / 4,893 assertions with only `RealtimeNetworkInvariantsTest#test_http_proxy_uses_connect_and_isolates_proxy_and_origin_credentials` erroring before the local proxy accepts a connection in this sandbox. The six neighboring network-invariant tests pass; CI remains authoritative for the full supported environment and Ruby 3.3/3.4/4.0 matrix.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ group :development, :test do
   gem "async"
   gem "async-websocket"
   gem "aws-sdk-core", "~> 3"
+  gem "cgi"
   gem "minitest", "~> 5.27"
   gem "minitest-hooks"
   gem "minitest-proveit"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -12,7 +12,6 @@ PATH
   remote: .
   specs:
     openai (0.80.0)
-      cgi
       connection_pool (>= 2.2.3)
       logger
 
@@ -225,6 +224,7 @@ DEPENDENCIES
   async
   async-websocket
   aws-sdk-core (~> 3)
+  cgi
   minitest (~> 5.27)
   minitest-hooks
   minitest-proveit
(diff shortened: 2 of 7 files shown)


HAYDEN-OAI (approved):
Reviewed exact head 620c49407b44d05aa53f9e7e20ed085fa29ba7de. No substantive query/URL compatibility, credential-handling, or supported-Ruby runtime-packaging issues found.

PR #468, opened 2026-08-19
fix: normalize filenames inferred from file handles

jbeckwith-oai:
## Summary

- Normalize inferred `OpenAI::FilePart` filenames for path-backed `IO` objects with the platform-native basename while preserving pathless streams, explicit filename overrides, and `Pathname` behavior.
- Preserve multipart content types, file contents, retry classification, caller-owned stream lifetime, and vector-store staging/cleanup.
- Add direct `client.files.create`, multipart wire-format, platform-path, metadata, and staged vector-store regression coverage.

## Compatibility

`OpenAI::FilePart#filename` now returns `document.txt` instead of `/nested/document.txt` when its name is inferred from a path-backed `IO`. This matches existing `Pathname` and explicit-filename normalization; callers that need the original filesystem path can continue using `file_part.content.to_path`. Public method signatures, visibility, RBI/RBS declarations, supported Ruby versions, dependencies, and generated interfaces are unchanged.

## Validation

- `bundle exec rake test` on Ruby 3.3.12, 3.4.10, and 4.0.6: **932 tests, 4,942 assertions** per runtime; the live Bedrock integration is intentionally skipped.
- `bundle exec rake lint`: Sorbet, 1,236 RBS signatures, formatter/directive checks, and RuboCop across 2,700 files.
- Reviewed generator ownership: production and regression-test changes are SDK-owned; no Castiron changes are required.

### DIFF
--- a/lib/openai/file_part.rb
+++ b/lib/openai/file_part.rb
@@ -70,7 +70,7 @@ module OpenAI
       in [nil, Pathname]
         content.basename.to_path
       in [nil, IO]
-        content.to_path
+        content.to_path&.then { ::File.basename(_1) }
       else
         filename
       end
--- a/test/openai/file_part_test.rb
+++ b/test/openai/file_part_test.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "test_helper"
+require "tempfile"
 
 class OpenAI::Test::FilePartTest < Minitest::Test
   def test_to_json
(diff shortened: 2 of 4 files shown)


HAYDEN-OAI (approved):
Independent deep security and compatibility review completed at `9e79d871738861b521e7e67d46784cf7e9675bb4`. Verified actual direct multipart uploads and vector-store staging, permissions, stream ownership, and cleanup on Ruby 3.3.12, 3.4.10, and 4.0.6. No introduced substantive issues found.

PR #481, opened 2026-08-19
Replace Steep with RBS CLI validation

jbeckwith-oai:
## Summary

- remove the direct Steep dependency and its unused transitive dependencies
- replace the Steep-backed signature check with `rbs parse` and `rbs validate`
- derive the loaded RBS libraries from `manifest.yaml`, including the `net/http` to `net-http` mapping
- preserve original-file diagnostics, file-scoped directives, malformed UTF-8/NUL handling, and GitHub Actions annotations
- remove `Steepfile`, obsolete `STEEP_JOBS` configuration, and formatting-policy references

## Why

The repository only needs signature validation for generated RBS files. The RBS CLI provides the parsing, name-resolution, and type-application validation needed here without carrying Steep and its worker/configuration layer.

## Validation

- `bundle check`
- `bundle exec rake validate:rbs` — 1,236 files
- `bundle exec rake lint` — Sorbet, RBS validation, rubyfmt, RuboCop directives, and 2,702-file RuboCop pass
- validator and formatting-policy tests — 22 runs, 253 assertions
- thermo-nuclear code-quality review

Timing on the rebased tree: 24.86s for 1,236 files. A pre-rebase Steep baseline was 23.42s for 1,219 files, so those measurements are not directly comparable.


scripts/validate-rbs
```diff
@@ -1,49 +1,51 @@
...
+        [rbs_command(env), "parse", "sig"],
+        stdout: stdout,
+        stderr: stderr,
+        env: env
+      )
+      return 1 unless parse_status.success?
+
+      command = [rbs_command(env), "-I", "sig", "--no-collection"]
+      libraries(root).each { command.push("-r", _1) }
+      command << "validate"
```
HAYDEN-OAI, line 41: [P2] Preserve subtype-aware signature validation

`rbs validate` checks name resolution and generic arity, but does not enforce the subtype constraints previously checked by Steep. With the locked RBS 4.1.3, this previously-rejected signature now exits successfully and prints `Validated 1 RBS file with no errors.`:

```rbs
module Example
  class Box[A < Numeric]
  end

  class InvalidBox < Box[String]
  end
end
```

The old validator rejected this exact case with `RBS::UnsatisfiableTypeApplication`. I also reproduced silent acceptance of invalid bounded defaults, incompatible inherited instance-variable types, and unsatisfied mixin self-type constraints. These false negatives allow invalid public `.rbs` contracts to pass validation and reach the packaged gem, where downstream Steep users reject them. Preserve subtype-aware semantic validation, or implement equivalent checks, before reporting success.

HAYDEN-OAI (commented):
One verified P2 semantic-validation regression; details are in the inline comment.

jbeckwith-oai:
Closing after rebasing against current main and reassessing the premise. `rbs validate` guarantees name resolution and generic arity, but it does not preserve the subtype-aware semantic validation Steep provides for bounded generic applications and defaults, inherited instance-variable compatibility, or mixin self-types. Retaining Steep would negate the dependency-removal goal, while recreating those checks here would amount to maintaining an unsupported subtype checker. This PR therefore cannot meet its stated goal without weakening signature validation, so we are closing it rather than expanding the wrong abstraction. Thanks @HAYDEN-OAI for identifying the P2 regression.

scripts/validate-rbs
jbeckwith-oai: Thanks for the concrete bounded-generic, bounded-default, inherited-instance-variable, and mixin-self-type repros. We reassessed the premise and agree that the RBS CLI replacement cannot preserve Steep’s subtype-aware semantics. Retaining Steep would negate the dependency-removal goal, while implementing equivalent checks here would amount to maintaining an unsupported subtype checker. We therefore closed PR #481 rather than claiming a code fix or shipping weaker validation; no replacement change was pushed.

PR #482, opened 2026-08-19
chore: isolate documentation dependencies

jbeckwith-oai:
## Summary

- move Redcarpet, WEBrick, and YARD into a dedicated `docs/Gemfile` and lockfile
- add one `scripts/docs` entry point for installing, building, and previewing documentation
- preserve the existing Rake aliases while clearing the parent Bundler environment
- document the separate contributor workflow and dependency-review boundary

## Why

Documentation-only gems currently live in the primary contributor bundle, so normal installs and CI resolve tooling they do not use. Keeping a separate docs bundle removes those three specs from the primary lock graph while preserving reproducible documentation builds and previews.

## Validation

- `bundle check`
- `BUNDLE_GEMFILE=docs/Gemfile bundle check`
- `bundle exec rake lint`
- `bundle exec rake build:docs` (2,358 files; both lockfile hashes unchanged)
- `PORT=18810 bundle exec rake docs:preview` (WEBrick reached server startup and was stopped cleanly)
- targeted contributor-script tests pass
- full local suite reached 986 passing tests; one unrelated Realtime proxy invariant cannot reach its test listener in the Codex sandbox, so GitHub CI will provide the clean-network result

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,9 @@ To set up the repository, run:
 $ ./scripts/bootstrap
 ```
 
-This will install all the required dependencies.
+This installs the dependencies used by the normal test, lint, typecheck, and
+package workflows. Documentation tooling has a separate bundle; see
+[Documentation](#documentation).
 
 ## Security requirements
 
@@ -30,10 +32,10 @@ This will install all the required dependencies.
   error messages, fixtures, and CI artifacts. Clearly fake or sanitized
   payloads may remain in tests and diagnostics.
 - **Dependencies:** Review `Gemfile`, `Gemfile.lock`, `gemfiles/*.gemfile`,
-  `gemfiles/*.gemfile.lock`, and `openai.gemspec`
-  changes, including direct and transitive gems, sources, locked Git revisions,
-  native extensions, and install/build scripts. Do not run unreviewed scripts
-  or accept unexplained lockfile changes.
+  `gemfiles/*.gemfile.lock`, `docs/Gemfile`, `docs/Gemfile.lock`, and
+  `openai.gemspec` changes, including direct and transitive gems, sources,
+  locked Git revisions, native extensions, and install/build scripts. Do not
+  run unreviewed scripts or accept unexplained lockfile changes.
 - **CI and publishing:** Pin GitHub Actions to full commit SHAs, grant minimum
   per-job permissions, and preserve `permissions: {}`,
   `persist-credentials: false`, protected release environments, and RubyGems
@@ -190,10 +192,26 @@ This can be installed along side Ruby LSP.
 
 2. For each generic type in `*.rbi` files, a spurious "Duplicate type member" error is present.
 
-## Documentation Preview
+## Documentation
+
+Documentation dependencies are intentionally isolated from the primary
+contributor bundle. Install them with:
+
+```bash
+$ ./scripts/docs install
+```
+
+To build the documentation, run:
+
+```bash
+$ ./scripts/docs build
+```
 
 To preview the documentation, run:
 
 ```bash
-$ bundle exec rake docs:preview [PORT=8808]
+$ PORT=8808 ./scripts/docs preview
 ```
+
+The existing `bundle exec rake build:docs` and
+`bundle exec rake docs:preview` tasks delegate to the same docs bundle.
--- a/Rakefile
+++ b/Rakefile
@@ -27,7 +27,7 @@ end
 
 desc("Preview docs; use `PORT=<PORT>` to change the port")
 multitask(:"docs:preview") do
-  Bundler.with_unbundled_env { sh(*%w[./scripts/docs preview]) }
+  Bundler.with_original_env { sh(*%w[./scripts/docs preview]) }
 end
 
 bedrock_tests = FileList["test/openai/providers/bedrock*_test.rb"]
@@ -161,7 +161,7 @@ multitask(lint: [:"lint:rubocop", :"lint:rubocop_directives", :typecheck])
 
 desc("Build yard docs")
 multitask(:"build:docs") do
-  Bundler.with_unbundled_env { sh(*%w[./scripts/docs build]) }
+  Bundler.with_original_env { sh(*%w[./scripts/docs build]) }
 end
 
 desc("Build ruby gem")
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "bundler"
 require "pathname"
 require "tempfile"
 
@@ -26,7 +27,7 @@ end
 
 desc("Preview docs; use `PORT=<PORT>` to change the port")
 multitask(:"docs:preview") do
-  sh(*%w[yard server --reload --quiet --bind \[::\] --port], ENV.fetch("PORT", "8808"))
+  Bundler.with_unbundled_env { sh(*%w[./scripts/docs preview]) }
 end
 
 bedrock_tests = FileList["test/openai/providers/bedrock*_test.rb"]
@@ -160,7 +161,7 @@ multitask(lint: [:"lint:rubocop", :"lint:rubocop_directives", :typecheck])
 
 desc("Build yard docs")
 multitask(:"build:docs") do
-  sh(*%w[yard])
+  Bundler.with_unbundled_env { sh(*%w[./scripts/docs build]) }
 end
 
 desc("Build ruby gem")
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Keep documentation tooling out of the primary contributor dependency graph.
-source "https://rubygems.org"
+source "https://rubygems.org", cooldown: 7
 
 gem "redcarpet"
 gem "webrick"
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Keep documentation tooling out of the primary contributor dependency graph.
+source "https://rubygems.org"
+
+gem "redcarpet"
+gem "webrick"
+gem "yard"
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -16,9 +16,9 @@ DEPENDENCIES
 
 CHECKSUMS
   bundler (4.0.17) sha256=214e21431b5665dd2f99df8a5511c6b151d7a72e8015c8b38f8b775b61cbb6c1
-  redcarpet (3.6.1)
-  webrick (1.9.2)
-  yard (0.9.45)
+  redcarpet (3.6.1) sha256=d444910e6aa55480c6bcdc0cdb057626e8a32c054c29e793fa642ba2f155f445
+  webrick (1.9.2) sha256=beb4a15fc474defed24a3bda4ffd88a490d517c9e4e6118c3edce59e45864131
+  yard (0.9.45) sha256=52e211493f7cb8a3ebf7e104a25a1e73937a3103092545d34cb88fafebb3dc51
 
 BUNDLED WITH
   4.0.17
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -1,0 +1,24 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    redcarpet (3.6.1)
+    webrick (1.9.2)
+    yard (0.9.45)
+
+PLATFORMS
+  ruby
+  x86_64-linux
+
+DEPENDENCIES
+  redcarpet
+  webrick
+  yard
+
+CHECKSUMS
+  bundler (4.0.17) sha256=214e21431b5665dd2f99df8a5511c6b151d7a72e8015c8b38f8b775b61cbb6c1
+  redcarpet (3.6.1)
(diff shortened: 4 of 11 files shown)


docs/Gemfile
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Keep documentation tooling out of the primary contributor dependency graph.
+source "https://rubygems.org"
```
HAYDEN-OAI: [P2] Preserve the existing seven-day RubyGems cooldown

The root `Gemfile` protects these same gems with `source "https://rubygems.org", cooldown: 7`, but the new independent Gemfile omits that option. `scripts/docs` selects this Gemfile directly, and Bundler 4.0.17 stores source cooldowns only in the active Gemfile; when the docs dependencies are added, unlocked, or refreshed, its resolver now accepts a release published moments earlier. This removes the repository's supply-chain vetting window specifically for executable documentation tooling and Redcarpet's native extension. Add `cooldown: 7` to this source as well.

docs/Gemfile
```diff
@@ -0,0 +1,8 @@
...
+
+gem "redcarpet"
+gem "webrick"
+gem "yard"
```
HAYDEN-OAI, line 8: [P2] Add the separate docs bundle to Dependabot

The existing `.github/dependabot.yml` has exactly one Bundler update entry, scoped to `directory: /`. Dependabot's Bundler fetcher only discovers a nested Gemfile from that root when the root Gemfile explicitly references it with `eval_gemfile`, which this intentionally isolated bundle does not. Moving Redcarpet, WEBrick, and YARD here therefore removes all three from the existing scheduled weekly version-update flow, including routine vulnerability fixes and non-advisory hardening. Their docs lockfile will no longer receive the configured automatic dependency bumps. Add a Bundler update entry for `/docs` with the same update/cooldown policy.

Rakefile
```diff
@@ -26,7 +27,7 @@ end
 
 desc("Preview docs; use `PORT=<PORT>` to change the port")
 multitask(:"docs:preview") do
-  sh(*%w[yard server --reload --quiet --bind \[::\] --port], ENV.fetch("PORT", "8808"))
+  Bundler.with_unbundled_env { sh(*%w[./scripts/docs preview]) }
```
HAYDEN-OAI: [P2] Preserve caller-provided Bundler policy when switching bundles

Bundler 4.0.17 implements `with_unbundled_env` by deleting every environment variable whose name begins with `BUNDLE_`. Both delegated docs tasks consequently discard caller-provided `BUNDLE_PATH`, `BUNDLE_FROZEN`, `BUNDLE_DEPLOYMENT`, `BUNDLE_COOLDOWN`, and source/mirror configuration; `scripts/docs` restores only `BUNDLE_GEMFILE`. A contributor using a custom bundle path cannot resolve the already-installed docs gems, and security-enforcing frozen/cooldown/mirror settings silently fail open. Use an environment transition that removes the parent bundle activation while preserving user-supplied Bundler settings, such as `Bundler.with_original_env`, for both docs aliases.

HAYDEN-OAI (commented):
Found three verified P2 regressions in documentation dependency isolation and Bundler security configuration.

docs/Gemfile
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. Added a Dependabot Bundler entry for /docs that mirrors the root schedule, 8-day update cooldown, dependency groups, and PR limit. The focused YAML-policy regression test asserts parity with the root entry; the focused harness passed 6 tests / 21 assertions, and lint passed.

Rakefile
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. The docs Rake tasks now use Bundler.with_original_env, which removes parent bundle activation while retaining caller-provided Bundler policy. End-to-end coverage verifies the mirror survives, docs/Gemfile becomes active, and BUNDLE_BIN_PATH is removed; the focused harness passed 6 tests / 21 assertions.

docs/Gemfile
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. docs/Gemfile now applies cooldown: 7 to the RubyGems source, matching repository policy. A focused regression assertion covers the exact source declaration; docs bundle check and docs build both passed.

CONTRIBUTING.md
```diff
@@ -29,10 +31,11 @@ This will install all the required dependencies.
   request/response bodies, prompts, or uploaded files from logs, exceptions,
   error messages, fixtures, and CI artifacts. Clearly fake or sanitized
   payloads may remain in tests and diagnostics.
-- **Dependencies:** Review `Gemfile`, `Gemfile.lock`, and `openai.gemspec`
-  changes, including direct and transitive gems, sources, locked Git revisions,
-  native extensions, and install/build scripts. Do not run unreviewed scripts
-  or accept unexplained lockfile changes.
+- **Dependencies:** Review `Gemfile`, `Gemfile.lock`, `docs/Gemfile`,
+  `docs/Gemfile.lock`, and `openai.gemspec` changes, including direct and
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. The root AGENTS dependency-review rule now explicitly includes docs/Gemfile and docs/Gemfile.lock. Focused coverage asserts both manifests remain named; dependency review confirmed unchanged locks and only the expected RubyGems docs dependencies.

docs/Gemfile
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. This duplicate cooldown report is covered by the same coherent fix: the docs RubyGems source now declares cooldown: 7, with an exact focused regression assertion. Docs bundle check and docs generation passed.

docs/Gemfile
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. This duplicate Dependabot report is covered by the new /docs Bundler entry, including the same weekly schedule, cooldown, grouping, and PR-limit policy as the root entry. The focused test compares those policy sections structurally and passed.

Rakefile
jbeckwith-oai: Fixed in 0bad10ecb4ced266fb757f58ca3dfc827ca683a6. This duplicate environment-policy report is covered by switching both docs tasks to Bundler.with_original_env. The regression test explicitly preserves BUNDLE_PATH, BUNDLE_FROZEN, BUNDLE_DEPLOYMENT, BUNDLE_COOLDOWN, and the RubyGems mirror while proving parent BUNDLE_BIN_PATH and BUNDLE_GEMFILE activation are removed; the focused harness passed.

scripts/docs
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd -- "$(dirname -- "$0")/.."
+
+export BUNDLE_GEMFILE="$PWD/docs/Gemfile"
```
HAYDEN-OAI, line 12: [P2] Preserve repository-local Bundler security configuration

Changing `BUNDLE_GEMFILE` also changes the application root: Bundler 4.0.17 derives `root` from the selected Gemfile and defaults `app_config_path` to `<root>/.bundle`. Consequently this script and both Rake aliases now load `docs/.bundle/config` instead of the existing repository `.bundle/config`; locally configured RubyGems mirrors, `frozen`/`deployment`, and install paths are silently ignored even though `with_original_env` preserves environment variables. A contributor with a required local mirror will fetch from public RubyGems or fail to install, and a locally enforced frozen policy no longer applies. Preserve an explicit `BUNDLE_APP_CONFIG`, or default it to `$PWD/.bundle` before selecting the docs Gemfile.

HAYDEN-OAI (commented):
The three earlier findings are fixed, but two verified P2 Bundler configuration and lockfile-isolation regressions remain.

scripts/docs
```diff
@@ -0,0 +1,26 @@
...
+
+case "${1:-}" in
+  install)
+    shift
```
HAYDEN-OAI, line 17: [P2] Prevent inherited lockfile overrides from rewriting the primary bundle

This command changes `BUNDLE_GEMFILE` but leaves a caller-provided `BUNDLE_LOCKFILE` untouched. Bundler 4.0.17 checks `BUNDLE_LOCKFILE` before deriving the selected Gemfile’s lockfile, so `BUNDLE_LOCKFILE=$PWD/Gemfile.lock ./scripts/docs install` resolves the docs manifest against the primary lockfile and `bundle install` can rewrite that primary SDK dependency graph; frozen/deployment callers instead fail on the manifest mismatch. `Bundler.with_original_env` also preserves a caller-provided override for the Rake entry points. Explicitly select `$PWD/docs/Gemfile.lock`, or clear an inherited lockfile override, when switching to the documentation bundle.

scripts/docs
jbeckwith-oai: Fixed in 7925f6774471e666605cd96b02fb790ec5e256c9. `scripts/docs` now sets `BUNDLE_APP_CONFIG` to the repository `.bundle` directory only when the caller did not explicitly supply a value, so repository frozen/path/mirror/security policy remains in scope even after selecting `docs/Gemfile`. Regression coverage exercises `install`, `build`, and `preview` directly plus both `build:docs` and `docs:preview` Rake aliases; it asserts Bundler resolves the repository config by default and preserves an explicit config path. The disposable Bundler 4 install also proves the primary `.bundle/config` remains byte-for-byte unchanged.

scripts/docs
jbeckwith-oai: Fixed in 7925f6774471e666605cd96b02fb790ec5e256c9. `scripts/docs` now forces the absolute `docs/Gemfile.lock` via `BUNDLE_LOCKFILE`, overriding any inherited primary lockfile selection while retaining the caller's other Bundler policy settings. The fail-first Bundler 4 regressions cover all three direct docs commands and both Rake aliases, and a real `bundle install --local` fixture starts with `BUNDLE_LOCKFILE` pointed at the primary lock. It verifies the primary Gemfile.lock, repository `.bundle/config`, and docs/Gemfile.lock all remain byte-for-byte unchanged. Final repository hashes also remained unchanged after bundle checks and a full docs build.

docs/Gemfile.lock
```diff
@@ -0,0 +1,24 @@
...
+  x86_64-linux
+
+DEPENDENCIES
+  redcarpet
+  webrick
+  yard
+
+CHECKSUMS
+  bundler (4.0.17) sha256=214e21431b5665dd2f99df8a5511c6b151d7a72e8015c8b38f8b775b61cbb6c1
+  redcarpet (3.6.1)
```
HAYDEN-OAI: [P2] Populate the documentation gem checksums

The new `CHECKSUMS` section records no digest for `redcarpet` (and the `webrick`/`yard` entries below are also empty). Bundler 4.0.17 treats each checksum-less RubyGems entry as a required lockfile update; a real run against this lockfile with `BUNDLE_FROZEN=true` raises `Your lockfile has an empty CHECKSUMS entry for "redcarpet", but can't be updated because frozen mode is set`. Because `scripts/docs` intentionally preserves frozen/deployment policy and repository `.bundle/config`, `./scripts/docs install` and the `build`/`preview` commands fail in frozen or deployment environments instead of using the committed docs bundle. Regenerate the lockfile with valid sha256 values for all three documentation gems.

HAYDEN-OAI (commented):
Found one verified P2: the new documentation lockfile breaks frozen/deployment installs and documentation commands.

docs/Gemfile.lock
jbeckwith-oai: Fixed in fdf1f7c52c1415e1445d17b3bbeb7856158d1dd7. I reproduced Bundler 4 frozen mode rejecting the empty `redcarpet` checksum, then regenerated `docs/Gemfile.lock` with the repository’s Bundler 4.0.17 `bundle lock --add-checksums` workflow. The lock now records SHA-256 values for redcarpet 3.6.1, webrick 1.9.2, and yard 0.9.45; each value was independently matched against the gem artifacts installed by both clean frozen and deployment installs. A focused regression requires a 64-hex SHA-256 for every docs gem (10 tests / 54 assertions on Ruby 3.3, 3.4, and 4.0). Clean `BUNDLE_FROZEN=true` and `BUNDLE_DEPLOYMENT=true` installs and full YARD builds passed in isolated paths; direct preview under both policies and the Rake preview alias reached WEBrick startup and were stopped cleanly with no listeners left. Primary bundle checks passed on all three Rubies, Bedrock check/tests passed (38 runs / 377 assertions), lint/types passed, package build passed, and primary Gemfile/lock hashes remained unchanged. Two-pass generalized review, thermo-nuclear review, Castiron, and the complete supply-chain/security diff scan found no issues.

HAYDEN-OAI (approved):
Verified official gem checksums, clean frozen and deployment Bundler installs, isolated YARD and WEBrick runtime, dependency cooldown and Dependabot coverage, and unchanged primary and Bedrock lockfiles. No substantive correctness or security findings.

PR #483, opened 2026-08-19
fix: require authority when validating request origins

jbeckwith-oai:
## Summary
- Require both configured and resolved request origins to have a non-empty host before applying the existing origin comparison.
- Preserve the established validation order and existing `OpenAI::Errors::Error` behavior before authentication, workload token materialization, or transport execution.
- Keep standard HTTP(S), IPv6/IPvFuture, Azure, Bedrock, realtime, and authority-bearing custom transports unchanged.

## Compatibility
Authorityless configured endpoints, such as `unix:///trusted.sock/v1`, are now rejected because a generic URI cannot establish a reliable scheme-independent destination identity. Custom transports can continue using an authority-bearing endpoint such as `unix://trusted-socket/v1` or an HTTP(S)-shaped `base_url`; relative, absolute same-origin, and network-path requests remain supported for those destinations.

No public Ruby methods, keyword arguments, return types, exception families, RBI/RBS declarations, generated SDK files, dependencies, or shared URI helpers changed.

## Validation
- Fail-first regressions reproduced on Ruby 3.3.12, 3.4.10, and 4.0.6 with the exact locked Minitest 6.0.6 dependencies.
- Focused origin/custom-transport suite: **24 tests / 412 assertions** on each supported Ruby.
- Complete suite: **1,014 tests / 6,391 assertions** on each supported Ruby; zero failures/errors and one existing opt-in live-provider skip.
- Expanded Azure/Bedrock, workload identity, realtime, retries, and custom-transport suite: **137 tests / 1,129 assertions**.
- `bundle exec rake lint`: rubyfmt, RuboCop (**2,705 files**), Sorbet, and **1,236 RBS signatures** all pass.
- `bundle exec rake build:gem` passes.
- Generalized, strict maintainability, public API compatibility, and focused security reviews completed with no outstanding findings.

### DIFF
--- a/lib/openai/internal/transport/base_client.rb
+++ b/lib/openai/internal/transport/base_client.rb
@@ -390,7 +390,9 @@ module OpenAI
             origin.hostname = address.to_s if address.ipv6?
           end
 
-          unless OpenAI::Internal::Util.uri_origin(request_origin) == OpenAI::Internal::Util.uri_origin(base_origin)
+          unless !request_origin.host.to_s.empty? &&
+              !base_origin.host.to_s.empty? &&
+              OpenAI::Internal::Util.uri_origin(request_origin) == OpenAI::Internal::Util.uri_origin(base_origin)
             raise OpenAI::Errors::Error, "Request path must resolve to the configured base URL origin"
           end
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed request-origin validation, authentication and provider credential sequencing, custom transport compatibility, and packaging. No substantive issues found.

PR #486, opened 2026-08-20
chore(api): refresh image model docs and generated lint policy

apcha-oai:
## Summary

Refresh the generated API reference and image-generation model documentation in the stable and beta Responses tool models and RBI comments. Public type signatures and runtime behavior are unchanged.

## Changes

- Document the supported image-model names, including `chatgpt-image-latest`, and preserve the existing `gpt-image-1` default.
- Refresh the transformed OpenAPI snapshot, including GPT Image 2 names for JSON image editing.
- Add CastIron's generated RuboCop policy for `Lint/NumberConversion` in the generated webhook resource.

## Review follow-up

The model-alias documentation correction is included. The [separate generated webhook lint-policy feedback](https://github.com/openai/openai-ruby/pull/486#discussion_r3824046606) remains open; this refresh does not change webhook behavior.

## (manual)

Resolved the three-way generation conflict in `lib/openai/client.rb` by preserving [the existing public-main `auth_headers` implementation](https://github.com/openai/openai-ruby/blob/8d716b657973ac759e2fe8e56e3158698b7f6451/lib/openai/client.rb) byte-for-byte. This retains provider-managed authentication, disabled-scheme handling, normal-key precedence, and missing-credential errors. No new handwritten behavior was added.

Validation: `./scripts/format` and `bundle exec ruby -Itest test/openai/client_authentication_test.rb` passed (11 tests, 25 assertions).


lib/openai/.rubocop.generated.yml
```diff
@@ -0,0 +1,6 @@
+# Generated by Castiron from api/castiron/policies/ruby/lint.yaml.
+# Do not edit; update the canonical policy manifest instead.
+---
+"Lint/NumberConversion":
```
jbeckwith-oai, line 4: **[Low] Avoid adding a blanket lint suppression.** This new generated policy suppresses `Lint/NumberConversion` for the webhook implementation, where `timestamp_header.to_i` already fails to provide the strict parsing implied by its `rescue ArgumentError`. Update the canonical generator to use a compliant strict conversion such as `Integer(timestamp_header, 10)` and regenerate without the exclusion; alternatively omit this unnecessary generated suppression.

jbeckwith-oai (changes requested):
The newly added exhaustive supported-image-model list omits `chatgpt-image-latest`, even though the adjacent schema enum and generated SDK model surfaces continue to support it. Please fix the authoritative source description and regenerate the SDK batch so the published documentation accurately lists every supported model. This change also introduces an avoidable generated webhook lint exclusion; please fix the canonical generator/policy instead of adding a blanket suppression.

jbeckwith-oai (changes requested):
The previous image-model documentation issue is fixed, but the existing review thread on `lib/openai/.rubocop.generated.yml` remains actionable: this PR still adds an avoidable `Lint/NumberConversion` exclusion for webhook timestamp parsing. Please fix the canonical generator with a strict compliant conversion and remove the generated exclusion. Activating the exclusion would conflict with the no-unnecessary-suppressions policy.

lib/openai/.rubocop.generated.yml
jbeckwith-oai: Addressed the policy-adoption follow-up in #702 (6309901aec40161fde291b1078cb9f178710fac1). The root config now inherits the existing generated fragment and enables Lint/NumberConversion, with tests proving that only lib/openai/resources/webhooks.rb is excluded and neighboring generated/handwritten paths remain checked. All 14 handwritten offenses are removed. The webhook exception remains intentional: its generator contract preserves malformed timestamps taking the InvalidWebhookSignatureError / ‘Webhook timestamp is too old’ path. Strict Integer(timestamp_header, 10) parsing would change that established error contract and requires a separate approved migration. The generator-owned fragment and webhook implementation are unchanged. Full lint and the 1,962-test suite pass.

PR #489, opened 2026-08-20
fix: preserve derived pagination cursors

jbeckwith-oai:
## Summary

- make page-derived follow-up cursors authoritative over an original `request_options[:extra_query][:after]`
- preserve unrelated extra query parameters on subsequent pages
- cover both ID-derived and server-provided cursors

## Reproduction

Start a paginated list with `request_options: {extra_query: {after: "old", trace: "enabled"}}` and let the first page return a follow-up cursor of `"new"`. The page implementation updates `req[:query][:after]`, but request construction deep-merges `extra_query` afterward, so `after=old` wins again. Repeated auto-pagination can therefore request the same page indefinitely.

## Compatibility

First-request behavior is unchanged: caller-supplied extra query values still override generated query values. After the first response, `BasePage` stores the effective merged query and clears only its higher-precedence `extra_query` copy. Existing page implementations can then replace the stored cursor for follow-up requests, while unrelated parameters such as `trace=enabled` remain in the query. This changes no public method, argument, return type, RBI, RBS, or generated API surface.

## Verification

- `bundle exec ruby -Itest test/openai/path_parameter_query_test.rb` — 16 runs, 23 assertions
- `bundle exec ruby -Itest test/openai/client_test.rb` — 36 runs, 189 assertions
- `./scripts/mock --daemon && bundle exec rake test` — 1,094 runs, 10,000 assertions, 0 failures, 0 errors, 1 intentional skip
- `bundle exec rake lint` — RuboCop 2,707 files, Sorbet, RBS 1,236 files, rubyfmt, and directive validation passed
- `bundle exec rake test:examples:inventory` — 23/29 exercised, 6 explicitly excluded
- `git diff --check`

### DIFF
--- a/lib/openai/internal/type/base_page.rb
+++ b/lib/openai/internal/type/base_page.rb
@@ -58,8 +58,16 @@ module OpenAI
         # @param response_metadata [OpenAI::ResponseMetadata]
         # @param page_data [Object]
         def initialize(client:, req:, response_metadata:, page_data:)
+          options = req[:options].to_h
+
           @client = client
-          @req = req
+          # Keep the effective query, but clear its higher-precedence copy so a page's
+          # follow-up cursor can replace the caller's original cursor.
+          @req = {
+            **req,
+            query: OpenAI::Internal::Util.deep_merge(req[:query].to_h, options[:extra_query].to_h),
+            options: {**options, extra_query: {}}
+          }
           @model = req.fetch(:model)
           @last_response = response_metadata
           super()
--- a/lib/openai/internal/type/base_page.rb
+++ b/lib/openai/internal/type/base_page.rb
@@ -61,11 +61,11 @@ module OpenAI
           options = req[:options].to_h
           query = OpenAI::Internal::Util
             .deep_merge(req[:query].to_h, options[:extra_query].to_h)
-            .to_h { |key, value| [key.is_a?(String) ? key.to_sym : key, value] }
+            .to_h { |key, value| [key == "after" ? :after : key, value] }
 
           @client = client
-          # Keep the normalized effective query, but clear its higher-precedence copy
-          # so a page's follow-up cursor can replace the caller's original cursor.
+          # Keep the effective query with a canonical cursor key, but clear its
+          # higher-precedence copy so a page can replace the caller's original cursor.
           @req = {
             **req,
             query: query,
--- a/lib/openai/internal/type/base_page.rb
+++ b/lib/openai/internal/type/base_page.rb
@@ -59,13 +59,16 @@ module OpenAI
         # @param page_data [Object]
         def initialize(client:, req:, response_metadata:, page_data:)
           options = req[:options].to_h
+          query = OpenAI::Internal::Util
+            .deep_merge(req[:query].to_h, options[:extra_query].to_h)
+            .to_h { |key, value| [key.is_a?(String) ? key.to_sym : key, value] }
 
           @client = client
-          # Keep the effective query, but clear its higher-precedence copy so a page's
-          # follow-up cursor can replace the caller's original cursor.
+          # Keep the normalized effective query, but clear its higher-precedence copy
+          # so a page's follow-up cursor can replace the caller's original cursor.
           @req = {
             **req,
-            query: OpenAI::Internal::Util.deep_merge(req[:query].to_h, options[:extra_query].to_h),
+            query: query,
             options: {**options, extra_query: {}}
           }
           @model = req.fetch(:model)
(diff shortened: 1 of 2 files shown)


lib/openai/internal/type/base_page.rb
```diff
@@ -58,8 +58,16 @@ def to_enum = super(:auto_paging_each)
...
         def initialize(client:, req:, response_metadata:, page_data:)
+          options = req[:options].to_h
+
           @client = client
-          @req = req
+          # Keep the effective query, but clear its higher-precedence copy so a page's
+          # follow-up cursor can replace the caller's original cursor.
+          @req = {
+            **req,
+            query: OpenAI::Internal::Util.deep_merge(req[:query].to_h, options[:extra_query].to_h),
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `b55ca9a2`.

The shared `BasePage` boundary now merges the effective query, canonicalizes top-level String keys to Symbols, and stores request options with `extra_query` cleared. That means a caller-supplied `"after"` and the internally derived `:after` occupy the same key, so the page-derived cursor is authoritative on follow-up requests. The first HTTP request is unchanged because this normalization happens only when the returned page stores its follow-up request state; unrelated String/Symbol extra-query parameters remain present.

I reproduced this fail-first for both the ID-derived `CursorPage` path and the server-provided `NextCursorPage` path: both tests failed with the stale String-key cursor still stored. After the fix:

- focused pagination regressions: 16 runs, 32 assertions, 0 failures/errors
- full offline suite: 1,094 runs, 10,009 assertions, 0 failures/errors (1 intentional skip)
- RuboCop/rubyfmt, Sorbet, RBS, example inventory, gem packaging, and `git diff --check`: passed
- generalized two-pass review, thermo-nuclear review, and bounded security diff scan: no actionable findings

Both regressions assert the old cursor on the first request, the new cursor on the follow-up, preservation of unrelated `"trace"` and `:debug` parameters, and an empty stored `options[:extra_query]` invariant.

lib/openai/internal/type/base_page.rb
```diff
@@ -58,8 +58,19 @@ def to_enum = super(:auto_paging_each)
         # @param response_metadata [OpenAI::ResponseMetadata]
         # @param page_data [Object]
         def initialize(client:, req:, response_metadata:, page_data:)
+          options = req[:options].to_h
+          query = OpenAI::Internal::Util
+            .deep_merge(req[:query].to_h, options[:extra_query].to_h)
+            .to_h { |key, value| [key.is_a?(String) ? key.to_sym : key, value] }
```
HAYDEN-OAI: **[P2] Preserve string-key overrides for embedded path query parameters**

Converting every query key to a symbol breaks the real paginated `client.beta.responses.input_items.list("resp_123", request_options: {extra_query: {"beta" => "false"}})` endpoint, whose generated path includes `?beta=true`. The first request correctly sends `?beta=false`, but the next page sends `?beta=true&beta=false&after=item_1`: the embedded value is reparsed under the String key `"beta"`, while the stored override is now the Symbol `:beta`. The parent implementation sends only `beta=false` on both requests. The conflicting duplicate can change beta routing or cause the follow-up to be rejected. Preserve non-cursor String keys, or normalize embedded path-query keys consistently before merging.

HAYDEN-OAI (commented):
One validated pagination regression: a real generated endpoint loses the effective override of an embedded query parameter on subsequent pages.

lib/openai/internal/type/base_page.rb
jbeckwith-oai: Implemented and pushed in ce5e1a99.

The BasePage boundary now canonicalizes only the page-owned String after key to :after. It preserves unrelated String keys, so beta=false remains keyed identically to the embedded beta=true query and continues to override it. The initial request remains unchanged; follow-up requests still clear stored extra_query and derive an authoritative after cursor.

Regression: a real generated client.beta.responses.input_items.list call with extra_query {"beta"=>"false"} uses a concrete transport to inspect raw URLs. Fail-first follow-up pairs were beta=false, beta=true, and after=item_1. After the fix they are exactly beta=false and after=item_1; the first request is exactly beta=false. Existing ID-derived and server-provided cursor regressions still preserve unrelated String/Symbol keys and assert stored extra_query is empty.

Verification: focused path suite 17 runs/36 assertions; generated resource 1/4; full offline 1095/10013 (1 expected skip); lint including RuboCop, Sorbet examples, and RBS; example inventory 23/29 with 6 documented exclusions; gem packaging; git diff --check. Sequential generalized review, thermo review, and a complete security diff scan all found no issues. No Castiron regeneration was needed because this is the handwritten BasePage customization boundary and no generated resource changed.

HAYDEN-OAI (approved):
Independently reviewed the exact head: no substantive P1/P2 findings. Pagination preserves unrelated query parameters while keeping derived cursors authoritative.

PR #490, opened 2026-08-20
fix(streaming): keep indexed snapshots compact

jbeckwith-oai:
## Summary

- treat streamed choice and tool-call indices as protocol identifiers while keeping snapshot arrays compact
- preserve deterministic index ordering and independent state for interleaved choices and tool calls
- reject negative indices before mutating stream state
- keep choice/tool-call lookup linear or constant-time and avoid repeated sorting for monotonic streamed appends

The behavior is implemented in the handwritten streaming accumulator; generated transport models remain unchanged.

## Validation

- rebased onto `origin/main` at `2c4abe5e321053be32916d0df823a20f621a8d4c`
- focused stream-index regressions on Ruby 3.3.12, 3.4.10, and 4.0.6: 8 runs / 40 assertions per version
- complete chat streaming and snapshot matrix: 42 runs / 200 assertions per version
- complete suite on Ruby 3.3.12, 3.4.10, and 4.0.6: 1,120 runs / 10,146 assertions per version
- `bundle exec rake lint`
- independent two-pass review, thermo-nuclear maintainability review, and scoped Codex Security diff scan


lib/openai/helpers/streaming/chat_completion_stream.rb
```diff
@@ -402,6 +404,30 @@ def find_input_tool(name)
           @input_tools.find { |tool| tool.dig(:function, :name) == name }
         end
 
+        def find_choice_snapshot(completion_snapshot, index)
+          completion_snapshot.choices.find { |choice| choice.index == index }
+        end
+
+        def find_tool_call_snapshot(tool_calls, index)
+          tool_calls.find { |tool_call| tool_call[:index] == index }
```
HAYDEN-OAI: **[P2] Keep indexed tool-call snapshot lookups constant-time**

`find_tool_call_snapshot` is called for every tool-call delta, so replacing indexed lookup with `Array#find` makes one valid chunk with N distinct, nonnegative tool indices quadratic; the additional scan in `tool_done_event` brings a 10,000-call chunk to roughly 100 million comparisons. Neither streamed tool-call arrays nor SSE events have a cardinality/size bound, and custom provider endpoints are supported, so an otherwise well-formed response can monopolize a Ruby worker. Maintain an identifier-to-snapshot hash alongside the compact ordered array, or reject oversized counts before processing.

HAYDEN-OAI (commented):
Found one introduced P2 reliability/scalability issue in the indexed streaming accumulator.

lib/openai/helpers/streaming/chat_completion_stream.rb
jbeckwith-oai: Fixed in a34a30f6. The accumulator now keeps compact ordered public arrays plus internal choice-index and per-choice tool-call-index hashes, rebuilding the tool-call map whenever a typed snapshot is reconstructed. The recursive indexed-array accumulator also builds one local index map and sorts only once after appends, so both affected paths avoid repeated linear scans without reintroducing sparse allocation. Deterministic TracePoint regressions cover 128 valid tool calls and 128 valid choices with operation-count bounds (no wall-clock threshold); fail-first measured 18,304 comparisons for tool-event lookup and 9,676 for accumulation before the complete fix. Verification: all three streaming suites pass on Ruby 3.3.12, 3.4.10, and 4.0.6; each full suite passes 1,097 runs / 10,028 assertions; lint, rubyfmt, diff checks, and gem build pass; generalized two-pass, thermo, and scoped security reviews found no remaining issues.

lib/openai/helpers/streaming/chat_completion_stream.rb
```diff
@@ -118,39 +121,53 @@ def get_choice_state(choice)
...
 
           completion_snapshot = @current_completion_snapshot
+          appended_choice = false
 
           chunk.choices.each do |choice|
-            accumulate_choice!(choice, completion_snapshot)
+            appended_choice = true if accumulate_choice!(choice, completion_snapshot)
           end
 
+          completion_snapshot.choices.sort_by!(&:index) if appended_choice
```
HAYDEN-OAI: **[P2] Avoid re-sorting every choice after each streamed append**

This full-array sort runs after every chunk that introduces a choice, even when choice identifiers are already dense and increasing. A valid response streaming choices 0 through 9,999 in separate chunks therefore invokes the sort key 50,004,999 times; before this change the same stream needed only 10,000 indexed writes. The SDK does not cap choice counts or streaming events, and custom-provider endpoints are supported, so an otherwise linear-size response can monopolize a Ruby worker. Fast-path already-ordered appends and reposition only out-of-order additions, or enforce a cardinality limit before accumulation.

HAYDEN-OAI (commented):
One substantive P2 performance regression is documented inline.

lib/openai/helpers/streaming/chat_completion_stream.rb
jbeckwith-oai: Fixed in 2e09128d. New choices now use ordered insertion: already-increasing indices append in O(1), while out-of-order additions use bsearch_index plus one compact Array#insert. A one-time sorted-state check preserves the existing observable order of an out-of-order initial chunk until the first later append, when the prior implementation also established sorted order. The deterministic 128-chunk regression failed first with 8,509 ParsedChoice#index reads against a linear bound of 2,048 and now passes; an additional compatibility test covers the initially out-of-order case. Verification passed on Ruby 3.3.12, 3.4.10, and 4.0.6: all three streaming files (41 tests / 198 assertions per version), each full suite (1,099 runs / 10,033 assertions), lint/RBS/RBI/type checks/RuboCop, rubyfmt, diff checks, gem build, and final thermo/security-invariant review.

jbeckwith-oai:
Rebased onto current `origin/main` and completed a comprehensive re-review. The independent pass found one sibling performance gap: monotonically appended tool calls still triggered a full `sort_by!` after every chunk. The deterministic regression failed first with 127 sorts for 127 appends and now passes with zero; out-of-order/interleaved ordering remains covered. Exact head `32bf0169bb08e89687a41f241dcf01859f8e30ad` is fully green. Local verification: chat streaming matrix 42 runs / 200 assertions on Ruby 3.3.12, 3.4.10, and 4.0.6; full suite 1,120 runs / 10,146 assertions on each; lint, types, rubyfmt, gem build, two-pass review, thermo review, and scoped security review all passed.

apcha-oai (approved):
[sdk codex review] Approved after two independent reviews of 15c96822ca9889eb5d921b692e591b8e0e855e88. Both passed the three streaming suites (42 tests/200 assertions) plus independently constructed sparse/interleaved-state probes covering compact arrays, ordering, sidecar identity, exact accumulation, and atomic negative-index rejection. Public SSE/strict-tool behavior was also checked. Both prior performance findings are substantively fixed; this does not claim globally linear processing or eliminate existing message-reconstruction costs. Required CI passes. Local tests used installed compatibility dependencies rather than the exact locked bundle. Final head/base/author/draft/mergeability checks pass; the branch is behind main, and no merge or gate bypass is requested.

PR #495, opened 2026-08-20
fix: retain collection conversion errors

jbeckwith-oai:
## Summary

- isolate each `ArrayOf` / `HashOf` child conversion so a successful sibling cannot clear an earlier element failure
- retain the last actual child failure, preserving the existing single-error-slot behavior for multiple failures, and fall back to an incoming parent error when every child succeeds
- keep exactness accounting and full collection traversal unchanged
- add fail-first runtime coverage plus a representative generated Realtime model regression

Addresses the collection error-retention follow-up in #376. This intentionally does not close the tracking issue.

## Root cause and rule

Collection converters previously passed the shared state directly to every child. `BaseModel` clears `state[:error]` at each field boundary, so a failing model element followed by a successful model element could end with no collection error; reversing the elements retained the failure despite identical exactness.

The fix reuses `Converter.coerce_with_error` for each child. Exactness remains shared, but child errors are isolated from siblings. After every element has been converted, the collection commits the last child failure, or restores the incoming parent error if no child failed. This preserves current last-failure behavior for multiple invalid scalar/model elements while removing success-after-failure order dependence.

## Compatibility boundary

- no change to raw caller-owned request storage, `BaseModel#[]`, `#to_h`, or serialization
- no change to typed accessor behavior: invalid request fields still raise `ConversionError`
- no change to union matching/ranking or coercion breadth
- no RBI/RBS or public API changes

## Generated-code ownership

No Castiron generator or generated model changes are required. Castiron's Ruby exclusion policy keeps `lib/openai/internal/` and the non-resource test tree SDK-owned. The generated `RealtimeMcpListTools` type is used only as representative contract coverage; generated files are unchanged.

## Validation

- fail-first regressions reproduced on `origin/main`: 3 runtime failures and 1 generated-model failure
- focused collection/generated regressions: 5 runs, 56 assertions, 0 failures/errors
- related BaseModel/union suites: 44 runs, 436 assertions, 0 failures/errors
- full suite excluding one pre-existing environment failure: 1,091 runs, 9,844 assertions, 0 failures/errors/skips
  - `RealtimeNetworkInvariantsTest#test_http_proxy_uses_connect_and_isolates_proxy_and_origin_credentials` fails identically on a clean `origin/main` worktree in this environment
- RuboCop: 2,716 files, no offenses
- Sorbet: no errors
- RBS: 1,236 files valid
- rubyfmt and RuboCop directive validation: passed
- example inventory and gem packaging: passed
- required thermo-nuclear code-quality review: passed after removing an unnecessary test-only identity adapter
- `git diff --check`: passed

### DIFF
--- a/lib/openai/internal/type/array_of.rb
+++ b/lib/openai/internal/type/array_of.rb
@@ -82,16 +82,22 @@ module OpenAI
 
           target = item_type
           exactness[:yes] += 1
-          value
-            .map do |item|
-              case [nilable?, item]
-              in [true, nil]
-                exactness[:yes] += 1
-                nil
-              else
-                OpenAI::Internal::Type::Converter.coerce(target, item, state: state)
-              end
+          previous_error = state.fetch(:error)
+          last_error = nil
+          converted = value.map do |item|
+            case [nilable?, item]
+            in [true, nil]
+              exactness[:yes] += 1
+              nil
+            else
+              coerced, item_error = OpenAI::Internal::Type::Converter.coerce_with_error(target, item, state: state)
+              last_error = item_error if item_error
+              coerced
             end
+          end
+
+          state[:error] = last_error || previous_error
+          converted
         end
 
         # @api private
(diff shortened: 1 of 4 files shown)


HAYDEN-OAI (approved):
Reviewed collection conversion error retention, nested collections, union behavior, generated models, and API compatibility; no substantive correctness or security issues found.

PR #496, opened 2026-08-20
refactor: isolate file polling status loops

apcha-oai:
## Summary

Move the existing file and vector-store polling status loops into the SDK-owned
`OpenAI::Helpers::ResourcePolling` helper, with matching private RBI/RBS
declarations. The three public resource methods retain their exact signatures
and delegate to the helper.

This is an extraction only. The shared `Internal::Poller`, terminal states,
request options and headers, timeout/error behavior, upload concurrency,
composite methods, and all existing public RBI/RBS declarations are unchanged.
The helper calls the original resource's `retrieve`, preserving subclass
overrides. A token-level comparison confirms each moved loop body is identical
after accounting for that explicit receiver.

The verified custom-code report remains at **50 mixed files**, while the
generated-file patch shrinks **2,892 → 2,809 lines**. Only the three intended
resource customizations change; the other 47 are unchanged. No compiler,
schema/config, generation metadata, API reference, or dependency changes.

## Tests and covered cases

The existing `OpenAI::Test::Resources::PollingHelpersTest` is unchanged:
37 tests cover server/custom intervals, all terminal states, unknown states,
strict and unbounded deadlines, workload-identity replay, shorter request
timeouts, headers/options, upload concurrency, idempotency, and composite
validation.

New `OpenAI::Test::ResourcePollingTest` in
`test/openai/helpers/resource_polling_test.rb`:

- `test_polling_preserves_retrieve_overrides_and_result_identity`: all three
  methods in synchronous and Async execution; pending-to-terminal transitions,
  exact returned object, direct override dispatch, headers, retries, intervals,
  and unmodified caller options.
- `test_polling_preserves_shorter_request_timeout_error_identity`: all three
  methods preserve the same timeout exception and shorter request timeout.
- `test_polling_preserves_unknown_status_error_messages`: exact error text for
  each resource.
- `test_public_polling_method_parameters_are_unchanged`: positional and keyword
  parameter contracts for all three public methods.

The four new tests pass against both the original and extracted implementation
(90 assertions). Combined focused suite: **41 tests, 281 assertions**.

## Validation

- Ruby 4.0.6, pinned rubyfmt 0.14.1, `./scripts/format`.
- `bundle exec rake lint`: 2,717 Ruby files, Sorbet, 1,237 RBS files; all pass.
- Full suite against pinned Steady 0.22.1: **1,091 tests, 9,892 assertions**,
  no failures or skips.
- Separate Bedrock suite: 38 tests, 377 assertions, one live-test skip.
- Gem build and packed-helper/signature/load checks pass.
- Existing public signatures and generation/API-reference metadata are
  byte-identical to the base; `git diff --check` passes.

### DIFF
--- a/lib/openai/helpers/resource_polling.rb
+++ b/lib/openai/helpers/resource_polling.rb
@@ -1,0 +1,154 @@
+# frozen_string_literal: true
+
+module OpenAI
+  module Helpers
+    # Product-specific status handling for file and vector-store polling.
+    #
+    # @api private
+    module ResourcePolling
+      # @api private
+      #
+      # @return [OpenAI::Models::FileObject]
+      def self.wait_for_file(resource, file_id, poll_interval:, timeout:, request_options:)
+        poller = OpenAI::Internal::Poller.new(
+          operation: "file #{file_id}",
+          poll_interval: poll_interval,
+          timeout: timeout
+        )
+        file = nil
+
+        begin
+          loop do
+            file = poller.request(request_options, resource: file) do |options|
+              resource.retrieve(file_id, request_options: options)
+            end
+
+            case file.status
+            when OpenAI::FileObject::Status::UPLOADED
+              poller.wait(file)
+            when
+                OpenAI::FileObject::Status::PROCESSED,
+                OpenAI::FileObject::Status::ERROR,
+                :deleted,
+                "deleted"
+              return file
+            else
+              raise(
+                OpenAI::Errors::PollingError,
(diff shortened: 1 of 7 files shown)


jbeckwith-oai (approved):
Independently reviewed the typed SDK-owned polling extraction across all three file/vector-store paths. Public signatures, terminal states, deadlines, exceptions, request/header isolation, retrieve overrides, response identity, and fiber behavior are preserved. The helper and RBI/RBS definitions are packaged correctly for supported Ruby versions; focused coverage and CI are green.

PR #497, opened 2026-08-20
fix: align nullable collection matchers

jbeckwith-oai:
## Summary

- make nullable `ArrayOf` and `HashOf` element matchers agree with their existing coercion behavior
- keep collection-level `nil` distinct from nullable elements and preserve non-nullable matching
- make `HashOf#to_sorbet_type` return the symbol-keyed hash type produced by coercion, including nullable values
- include element nullability in collection converter hashes, matching equality and inspect semantics

## Compatibility

An executable differential compared the published `openai` 0.80.0 gem, the refreshed `origin/main` base, and this branch for nullable and non-nullable Array/Hash values, symbol and string hash keys, and unions with nullable collection and broad fallback variants in both orders.

The released gem and current main were identical. This branch changed only:

- nullable matcher results from false to true for values already accepted by coercion
- post-coercion matcher agreement from false to true
- nullable-first union dumping to use the now-matching nullable collection variant

Fallback-first unions, non-nil values, invalid element values, non-nullable variants, and `nil` collections were unchanged. `Union` itself is unchanged.

## Generator ownership

The runtime collection types and tests changed here are SDK-owned. Generated model, RBI, and RBS files were not edited. The handwritten contract test exercises generated `Metadata` and completion prompt declarations; the runtime Sorbet hash output now agrees with the generated symbol-keyed RBS type.

## Validation

- focused nullable collection, Sorbet, BaseModel, and structured output tests
- released gem / current main / branch compatibility differential
- `bundle exec rake lint`
- Bedrock bundle check and test suite: 38 runs, 377 assertions, 0 failures, 1 existing skip
- primary suite: 1,094 of 1,095 tests pass locally; the remaining Realtime proxy invariant is blocked by the sandbox proxy stack and reproduces unchanged against untouched `origin/main`
- thermo-nuclear code quality review
- `git diff --check`

Refs #376

### DIFF
--- a/lib/openai/internal/type/array_of.rb
+++ b/lib/openai/internal/type/array_of.rb
@@ -38,7 +38,21 @@ module OpenAI
         # @param other [Object]
         #
         # @return [Boolean]
-        def ===(other) = other.is_a?(Array) && other.all?(item_type)
+        def ===(other)
+          return false unless other.is_a?(Array)
+
+          target = item_type
+          other.all? do |item|
+            case item
+            in ^target
+              true
+            in nil
+              nilable?
+            else
+              false
+            end
+          end
+        end
 
         # @api public
         #
@@ -52,7 +66,7 @@ module OpenAI
         # @api public
         #
         # @return [Integer]
-        def hash = [self.class, item_type].hash
+        def hash = [self.class, item_type, nilable?].hash
 
         # @api private
         #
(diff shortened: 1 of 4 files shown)


HAYDEN-OAI (approved):
Reviewed nullable collection matching, union selection and serialization, invalid nested values and keys, exactness/error propagation, generated Realtime compatibility, and runtime/Sorbet signatures. No substantive correctness or compatibility issues found.

PR #499, opened 2026-08-20
chore(api): Clarify image background and realtime audio-delta docs

apcha-oai:
## Summary

Update SDK documentation to clarify allowed image background values and to explain that realtime translation audio deltas contain variable-length PCM16 chunks.

## Changes

- Clarify 'background' parameter: explicitly state allowed values are 'transparent', 'opaque', or 'auto', note that transparent support is in preview for some GPT image models and recommend using 'png' or 'webp' when using 'transparent'
- Clarify realtime translation audio delta semantics: the 'delta' field contains a PCM16 audio chunk whose length can vary and clients should decode and queue the complete delta rather than assuming fixed frame sizes


lib/openai/models/image_generate_params.rb
```diff
@@ -18,18 +18,13 @@ class ImageGenerateParams < OpenAI::Internal::Type::BaseModel
       required :prompt, String
 
       # @!attribute background
-      #   Allows to set transparency for the background of the generated image(s). This
-      #   parameter is only supported for GPT image models that support transparent
-      #   backgrounds. Must be one of `transparent`, `opaque`, or `auto` (default value).
-      #   When `auto` is used, the model will automatically determine the best background
-      #   for the image.
+      #   Allows to set transparency for the background of the generated image(s). Must be
```
jbeckwith-oai, line 21: The updated description removes the existing parameter-level restriction that `background` is supported only for GPT Image models. It now limits only transparent backgrounds, incorrectly implying `dall-e-2` / `dall-e-3` accept `background: "opaque"` or `"auto"`. Restore the explicit GPT-Image-only restriction in the canonical OpenAPI description while retaining the new transparency preview and PNG/WebP guidance, then regenerate.

jbeckwith-oai (changes requested):
The updated description removes the existing parameter-level restriction that `background` is supported only for GPT Image models. It now limits only transparent backgrounds, incorrectly implying `dall-e-2` / `dall-e-3` accept `background: "opaque"` or `"auto"`. Restore the explicit GPT-Image-only restriction in the canonical OpenAPI description while retaining the new transparency preview and PNG/WebP guidance, then regenerate.

apcha-oai:
overriding as docstring only change; api ref authoritative here

PR #501, opened 2026-08-20
refactor: isolate Chat Completions structured-output parsing

apcha-oai:
## Summary

Move Chat Completions structured-output preparation, tool-model decoration, and
the response-unwrapping callback into the SDK-owned
`OpenAI::Helpers::StructuredOutput::ChatCompletionParser`, with private RBI/RBS
declarations.

The existing public `get_structured_output_models` and
`build_tools_with_models` methods remain at the same visibility and with the
same parameters. The callback remains a stateful lambda; its captured-model
behavior is intentionally preserved. Both moved methods and the lambda are
token-identical to the original. Public request/streaming methods, signatures,
shared converters, dependencies, and generation/API-reference metadata are
unchanged.

The verified report remains at **50 mixed files**, while custom-patch lines
drop **2,809 → 2,706** against current public main. The Chat Completions resource
changes from +145 to +42 lines; the other 49 customizations are unchanged.

## Tests and covered cases

New `OpenAI::Test::ChatCompletionParserTest` in
`test/openai/helpers/chat_completion_parser_test.rb` passes against the original
and extracted implementation: **7 tests, 93 assertions**.

- `test_response_format_forms_keep_their_precedence_over_tools`: all three
  response-format forms, existing branch precedence, names/strictness, and
  nested object identity.
- `test_tool_conversion_preserves_names_and_input_identity`: direct and nested
  tool models, inferred/explicit names, and untouched tools.
- `test_build_tools_only_copies_matching_symbol_function_tools`: nil tools,
  matching functions, unknown names, string discriminators, and copy identity.
- `test_create_preserves_request_shape_and_unwrap_contract`: method/path/body,
  security/options/model, lambda arity, raw identity, and typed parsed content.
- `test_reused_unwrap_retains_its_captured_model_state`: repeated callback use,
  known and unknown tools, and the original model-state transitions.
- `test_unwrap_preserves_nil_malformed_json_and_missing_key_behavior`: nil and
  malformed JSON plus missing message/content/function/name/arguments keys.
- `test_existing_resource_helper_parameters_and_visibility_are_unchanged`:
  both existing public helper contracts.

No existing tests are removed or modified.

## Validation

- Focused structured-output suites: **27 tests, 205 assertions**.
- Ruby 4.0.6 / rubyfmt 0.14.1; formatting, 2,720-file RuboCop, Sorbet, and
  1,238 RBS files pass.
- Full suite against Steady 0.22.1 on current public main: **1,098 tests,
  9,985 assertions**, no failures or skips.
- Bedrock suite and gem build pass; packed helper/signature/load check passes.
- Exact moved-body/lambda comparison, unchanged public-signature/metadata
  checks, and `git diff --check` pass.

### DIFF
--- a/lib/openai/helpers/structured_output/chat_completion_parser.rb
+++ b/lib/openai/helpers/structured_output/chat_completion_parser.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+module OpenAI
+  module Helpers
+    module StructuredOutput
+      # Request preparation and response coercion for Chat Completions.
+      #
+      # @api private
+      module ChatCompletionParser
+        # @api private
+        def self.build_unwrap(model, tool_models)
+          # rubocop:disable Metrics/BlockLength
+          -> (raw) do
+            if model.is_a?(OpenAI::StructuredOutput::JsonSchemaConverter)
+              raw[:choices]&.each do |choice|
+                message = choice.fetch(:message)
+                begin
+                  content = message.fetch(:content)
+                  parsed = content.nil? ? nil : JSON.parse(content, symbolize_names: true)
+                rescue JSON::ParserError => e
+                  parsed = e
+                end
+
+                coerced = OpenAI::Internal::Type::Converter.coerce(model, parsed)
+                message.store(:parsed, coerced)
+              end
+            end
+
+            raw[:choices]&.each do |choice|
+              choice.dig(:message, :tool_calls)&.each do |tool_call|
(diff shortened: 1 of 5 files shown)


jbeckwith-oai (approved):
Independently verified the extracted Chat Completions structured-output helper preserves existing public/private delegates, request shape, schema parsing, malformed-input behavior, and the intentionally stateful unwrap callback. Existing lint annotations are only relocated, typing/packaging remain intact, and CI is green.

PR #512, opened 2026-08-21
chore(api): refresh SDK generation metadata

apcha-oai:
## Summary

Refresh the SDK generation metadata after compiler and configuration updates. The only changed file is `.castiron.stats.yml`; Ruby source, tests, API-reference content, public types, and runtime behavior are unchanged.


jbeckwith-oai (approved):
Approved after two independent reviews: generation metadata only, unchanged generated content hash, no runtime/public API/workflow/security changes, and passing public checks.

PR #514, opened 2026-08-22
test: preserve large HTTP response payloads

apcha-oai:
## Summary

Protect existing large HTTP payload behavior against future regressions; this PR does not remove or change any existing limit. Large JSON responses and SSE events are valid API payloads. Carry the guidance from https://github.com/openai/openai-node/pull/2433 into the Ruby SDK: prefer incremental processing, efficient buffering, cleanup, and cancellation over arbitrary rejection limits; any new limit requires an owner-approved API contract.

Add three sequential, offline regressions using synthetic 32 MiB + 1 byte text over a local HTTP server:

- Responses JSON decoding and `output_text`.
- Responses streaming deltas, accumulated snapshots, final response, and output-text helpers.
- Chat Completions streaming deltas, its independent accumulator, final completion, and output-text helper.

The streaming helpers exercise the same HTTP/SSE reader as their raw-stream counterparts. The fixture size is a regression probe, not a new API maximum; its high memory use is intentional. No runtime limits, dependencies, generated files, custom-code budgets, or generation metadata change. WebSockets are outside this PR's scope.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,3 +45,21 @@ The checker and effective budget come from main, not the PR. Keep default CODEOW
   [CONTRIBUTING.md](CONTRIBUTING.md).
 - Report suspected vulnerabilities privately as described in
   [SECURITY.md](SECURITY.md), never in public issues or pull requests.
+
+## Large-payload compatibility
+
+Treat large payloads as a normal API contract, not evidence of malformed or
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the serial public HTTP/SSE contract coverage, Responses and Chat Completions accumulators, final-result helpers, and cleanup. No production behavior changes; CI is green.

PR #518, opened 2026-08-23
examples: validate parsed chat function tool calls

jbeckwith-oai:
## Summary

- Require the handwritten structured-output Chat Completions function-calling example to observe at least one real function tool call whose parsed arguments are a `GetWeather` instance.
- Preserve successful weather output and refusal filtering while silently ignoring custom tools and incorrectly parsed values; abort with a deterministic explanation when no matching call exists.
- Add seven isolated, offline regression cases using the existing Responses sibling's wrapped-example loading and verified Minitest mocking patterns.

## Evidence and scope

- The new regression suite failed six of seven cases before the example fix: empty choices, all-refused choices, absent tool calls, incorrect parsed values, custom-only calls, and mixed/nonmatching output.
- Only `examples/structured_outputs_chat_completions_function_calling.rb` and its uniquely named regression-test file change.
- `CONTRIBUTING.md` explicitly identifies `examples/` as handwritten and never modified by the generator; no generated SDK, API contract, transport, authentication, serialization, dependency, or shared example code changes.
- Existing successful function-call output is preserved. `GetWeather === parsed` keeps normal Ruby class matching compatible with this example's `# typed: strong` Sorbet contract.

## Validation

- `bundle exec ruby test/openai/structured_outputs_chat_completions_function_calling_example_test.rb` on Ruby 3.3.12, 3.4.10, and 4.0.6: seven tests and 45 assertions pass on every supported minor.
- `TEST='test/openai/structured_outputs_*function_calling_example_test.rb' bundle exec rake test`: 10 Chat/Responses sibling tests and 62 assertions pass; the offline example inventory also passes.
- `TMPDIR=/private/tmp bundle exec rake test`: 1,198 tests and 10,638 assertions pass, with one existing skip. The explicit temporary directory avoids an unrelated existing macOS `/var` versus `/private/var` path comparison.
- `TMPDIR=/private/tmp bundle exec rake lint`: full rubyfmt, RuboCop (2,747 files), suppression guard, Sorbet, and 1,239 RBS signature validations pass.
- `python3 -m unittest discover -s scripts/castiron -p 'test_custom_code*.py'`: 51 Castiron tests pass, with one existing skip.
- Trusted-main custom-code budget check against base `edfb30b2973d68b4a19ce3f86f1de7313f730702` and head `ddb6323202c067245cad9658002a345d27849f8f`: isolation and budget pass at 2,659 / 4,000 lines.
- Focused Ruby formatting/RuboCop checks, extensive general review, thermo-nuclear maintainability review, `git diff --check`, and protected-base scope review all pass.

### DIFF
--- a/examples/structured_outputs_chat_completions_function_calling.rb
+++ b/examples/structured_outputs_chat_completions_function_calling.rb
@@ -23,6 +23,7 @@ chat_completion = client.chat.completions.create(
   tool_choice: {type: :function, function: {name: "GetWeather"}}
 )
 
+parsed_tool_call_count = 0
 chat_completion
   .choices
   .reject { _1.message.refusal }
@@ -31,8 +32,12 @@ chat_completion
     case tool_call
     when OpenAI::Chat::ChatCompletionMessageFunctionToolCall
       # parsed is an instance of `GetWeather`
-      pp(tool_call.function.parsed)
-    else
-      puts("Unexpected tool call type: #{tool_call.type}")
+      parsed = tool_call.function.parsed
+      next unless GetWeather === parsed
+
+      parsed_tool_call_count += 1
+      pp(parsed)
     end
   end
+
+abort("The chat completion did not contain a parsed GetWeather tool call") if parsed_tool_call_count.zero?
--- a/test/openai/structured_outputs_chat_completions_function_calling_example_test.rb
+++ b/test/openai/structured_outputs_chat_completions_function_calling_example_test.rb
@@ -1,0 +1,160 @@
+# frozen_string_literal: true
+
+require_relative "test_helper"
+
+class OpenAI::Test::StructuredOutputsChatCompletionsFunctionCallingExampleTest < Minitest::Test
+  extend Minitest::Serial
+
+  EXAMPLE_PATH = File.expand_path(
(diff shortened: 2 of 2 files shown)


apcha-oai (approved):
[sdk codex review] Approved at `dc4e8a8aa3087e996497f6008fc9520aa778d573` after two independent full reviews of both changed files, the supporting parser/model and test-isolation contracts, and all existing feedback. No actionable correctness, security, compatibility or maintainability blocker was found. The example now requires a parsed `GetWeather` function call, preserves successful output/refusal filtering, and has seven isolated regression cases.

Existing hosted Ruby 3.3.12, 3.4.10 and 4.0.6 jobs each passed 1,201 tests and 10,659 assertions at merge `55c37c50a55578a8f0b1d83742ba71f78b44fce3` of this head into base `5a07c3e79a8cbaf3e06e5725864ed77e3c090235`. The four newer main commits through `c3fe427` were inspected for integration impact, but those old-base runs are not claimed as execution against newer main. Current CI is green apart from expected skips. No candidate code, tests or scans were executed locally.

Final checks confirmed unchanged head/PR base/actual merge-base, Justin’s authorship, OPEN/non-draft status and mergeability. No duplicate inline comments.

PR #522, opened 2026-08-24
feat(api): add project residency and cost quantity units

apcha-oai:
## Summary

Adds typed residency support for admin organization projects and project creation, including supported residency enum values. Deprecates geography for project creation in favor of residency while retaining geography for backward compatibility, including project updates. Adds optional quantity_unit fields and known unit values to organization cost results returned across usage response models; unknown units remain supported as strings. These changes are additive and preserve existing API compatibility.

Project responses and project-creation inputs now reuse a shared residency model. Usage cost responses reuse a shared quantity-unit model, with unknown unit values and optional/null fields preserved.


lib/openai/models/admin/organization/project_create_params.rb
```diff
@@ -22,15 +22,28 @@ class ProjectCreateParams < OpenAI::Internal::Type::BaseModel
...
-          # @!method initialize(name:, external_key_id: nil, geography: nil, request_options: {})
+          # @!attribute residency
+          #   Create the project with the specified residency configuration. Your organization
+          #   must have access to the requested residency configuration in order to use it.
+          #   See
+          #   [data residency controls](https://platform.openai.com/docs/guides/your-data#data-residency-controls)
+          #   to review the functionality and limitations of setting this field.
+          #
+          #   @return [Symbol, OpenAI::Models::Admin::Organization::ProjectResidency, nil]
+          optional :residency, enum: -> { OpenAI::Admin::Organization::ProjectResidency }, nil?: true
```
jbeckwith-oai, line 44: [P2] Make `geography` and `residency` mutually exclusive in the authoritative schema/generated parameter contract. This new optional residency field remains independent of existing `geography`, even though the adjacent public documentation explicitly says not to provide both; `projects.create(name:, geography:, residency:)` therefore serializes an unsupported request that the real endpoint rejects. Encode the exclusion upstream, regenerate this model, and add separate geography-only/residency-only cases plus rejection coverage for the combined input.

jbeckwith-oai (changes requested):
Requesting changes for a P2 API-contract mismatch: project creation independently accepts geography and residency despite explicitly prohibiting that combination, and no Ruby creation regression covers the new residency mode.

jbeckwith-oai (approved):
LGTM.

PR #531, opened 2026-08-25
feat(api): support backend-mediated Realtime WebRTC calls

apcha-oai:
## Summary

Add `realtime.calls.create` so server-side applications can exchange an SDP offer and optional session configuration for an SDP answer.

This makes WebRTC signaling available through the SDK instead of requiring a handwritten HTTP request. The new operation sends the API’s typed multipart parts and exposes the SDP answer as a `StringIO`.


jbeckwith-oai (changes requested):
Requesting changes for the existing P1 discussion at lib/openai/resources/realtime/calls.rb:50. The new create method returns only a decoded StringIO, while the client attaches response metadata only to BaseModel values. The call identifier is returned in the Location response header, so a successful caller cannot recover the ID required to monitor or hang up its Realtime call. Return a raw-response/header-preserving wrapper or equivalent accessible response metadata and add a Location/call-ID regression. Existing inline discussion already captures this blocker.

jbeckwith-oai (approved):

PR #532, opened 2026-08-26
feat(responses): parse structured outputs on retrieval

### LINKED ISSUE #308: Parse structured outputs when retrieving background Responses
Carried over from #126 and originally requested by @agrobbin.

Structured-output parsing works when creating a Response synchronously and when resuming a response stream with the model supplied again. It is still not available for the ordinary non-streaming background retrieval path: `client.responses.retrieve(response_id)` does not accept a structured-output model, so retrieved output cannot be coerced back into that model through `content.parsed`.

## Desired behavior

Allow callers to supply the original structured-output model when retrieving a stored/background Response, for example:

```ruby
response = client.responses.retrieve(
  response_id,
  text: CalendarEvent
)
```

The exact API can differ, but it should:

- parse matching output content into the supplied model;
- support tool models consistently where applicable;
- leave retrieval without a model backward-compatible;
- document the background create/poll/retrieve flow; and
- include tests for completed and not-yet-completed background responses.

Related partial support: structured-output parsing is already available when resuming streams. Source: #126.

cjc0013:
## Summary

- allow `responses.retrieve` to accept the same local `text` and `tools` structured-output models used when creating a background response
- remove those local parsing hints before serializing retrieval query parameters, then reuse the existing Responses parser for returned text and function calls
- preserve caller-owned tool arrays and nested hashes while extracting parser models
- document the background create/retrieve flow

Closes #308.

## Tests

- Responses-related regression selection: 7 existing test files, 39 runs, 250 assertions, 0 failures, 0 errors, 0 skips
- structured-output parser selection: 3 existing test files (including the 2 new retrieval tests), 22 runs, 161 assertions, 0 failures, 0 errors, 0 skips
- `rubyfmt 0.14.1 --check` on the 2 modified Ruby files
- `rubocop` on the 2 modified Ruby files: no offenses

### DIFF
--- a/lib/openai/helpers/sorbet.rb
+++ b/lib/openai/helpers/sorbet.rb
@@ -259,6 +259,11 @@ module OpenAI
           super
         end
 
+        def retrieve(response_id, params = {})
+          reject_sorbet_function_tools!(params)
+          super
+        end
+
         def stream(params)
           reject_sorbet_streaming_formats!(params)
           super
--- a/lib/openai/helpers/sorbet.rb
+++ b/lib/openai/helpers/sorbet.rb
@@ -286,7 +286,7 @@ module OpenAI
         end
 
         private def reject_sorbet_function_tools!(params = {})
-          tools = Array(params[:tools])
+          tools = Array(params[:tools]).dup
           tools.concat(Array(params["tools"])) if params.is_a?(Hash)
 
           if tools.any? { sorbet_adapter?(_1) }
--- a/lib/openai/helpers/sorbet.rb
+++ b/lib/openai/helpers/sorbet.rb
@@ -286,6 +286,7 @@ module OpenAI
         end
 
         private def reject_sorbet_function_tools!(params = {})
+          params = params.to_h if params.is_a?(OpenAI::Internal::Type::BaseModel)
           tools = Array(params[:tools]).dup
           tools.concat(Array(params["tools"])) if params.is_a?(Hash)
 
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -577,7 +577,7 @@ module OpenAI
       end
 
       def structured_output_response_complete?(raw)
-        return raw[:status] == "completed" unless raw[:status].nil?
+        return false unless [nil, "completed"].include?(raw[:status])
 
         raw[:output].to_a.none? do |output|
           ["queued", "in_progress", "incomplete"].include?(output[:status])
@@ -596,6 +596,7 @@ module OpenAI
             if normalized_key == :type && ["function", "json_schema"].include?(normalized_item)
               normalized_item = normalized_item.to_sym
             end
+
             [normalized_key, normalized_item]
           end
         else
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -392,6 +392,8 @@ module OpenAI
 
         unwrap = if model || !tool_models.empty?
           -> (raw) do
+            next raw unless raw[:status] == "completed"
+
             parse_structured_outputs!(raw, model, tool_models)
           end
         end
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -392,7 +392,7 @@ module OpenAI
 
         unwrap = if model || !tool_models.empty?
           -> (raw) do
-            next raw if ["in_progress", "queued"].include?(raw[:status])
+            next raw unless [nil, "completed"].include?(raw[:status])
 
             parse_structured_outputs!(raw, model, tool_models)
           end
@@ -581,7 +581,10 @@ module OpenAI
         when Array
           value.map { |item| duplicate_structured_output_params(item) }
         when Hash
-          value.to_h { |key, item| [key, duplicate_structured_output_params(item)] }
+          value.to_h do |key, item|
+            normalized_key = key.is_a?(String) ? key.to_sym : key
+            [normalized_key, duplicate_structured_output_params(item)]
+          end
         else
           value
         end
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -378,12 +378,10 @@ module OpenAI
       #
       # @see OpenAI::Models::Responses::ResponseRetrieveParams
       def retrieve(response_id, params = {})
-        structured_output_params = duplicate_structured_output_params(params.slice(:text, :tools))
-        request_params = params.dup
-        request_params.delete(:text)
-        request_params.delete(:tools)
-
-        parsed, options = OpenAI::Responses::ResponseRetrieveParams.dump_request(request_params)
+        parsed, options = OpenAI::Responses::ResponseRetrieveParams.dump_request(params)
+        structured_output_params = duplicate_structured_output_params(parsed.slice(:text, :tools))
+        parsed.delete(:text)
+        parsed.delete(:tools)
         query = OpenAI::Internal::Util.encode_query_params(parsed)
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
@@ -392,8 +390,10 @@ module OpenAI
 
         model, tool_models = get_structured_output_models(structured_output_params)
 
-        unwrap = -> (raw) do
-          parse_structured_outputs!(raw, model, tool_models)
+        unwrap = if model || !tool_models.empty?
+          -> (raw) do
+            parse_structured_outputs!(raw, model, tool_models)
+          end
         end
 
         @client.request(
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -392,7 +392,7 @@ module OpenAI
 
         unwrap = if model || !tool_models.empty?
           -> (raw) do
-            next raw unless structured_output_response_complete?(raw)
+            next raw unless structured_output_response_complete?(raw, model, tool_models)
 
             parse_structured_outputs!(raw, model, tool_models)
           end
@@ -576,12 +576,50 @@ module OpenAI
         OpenAI::Helpers::StructuredOutput::ResponseParser.get_models(parsed)
       end
 
-      def structured_output_response_complete?(raw)
+      def structured_output_response_complete?(raw, model, tool_models)
         return false unless [nil, "completed"].include?(raw[:status])
 
         raw[:output].to_a.none? do |output|
-          ["queued", "in_progress", "incomplete"].include?(output[:status])
+          case output[:type]
+          when "message"
+            model && unfinished_structured_output_message?(output)
+          when "function_call"
+            unfinished_structured_output_function_call?(output, tool_models)
+          else
+            false
+          end
         end
+      end
+
+      def unfinished_structured_output_message?(output)
+        return true if structured_output_pending_status?(output[:status])
+        return false unless output[:status].nil?
+
+        output[:content].to_a.any? do |content|
+          content[:type] == "output_text" &&
+            (!content.key?(:text) || !valid_structured_output_json?(content[:text]))
+        end
+      end
+
+      def unfinished_structured_output_function_call?(output, tool_models)
+        return false if tool_models.empty?
+        return true unless output.key?(:name)
+        return false unless tool_models.key?(output[:name])
+        return true if structured_output_pending_status?(output[:status])
+        return true unless output.key?(:arguments)
+
+        output[:status].nil? && !valid_structured_output_json?(output[:arguments])
+      end
+
+      def structured_output_pending_status?(status)
+        ["queued", "in_progress", "incomplete"].include?(status)
+      end
+
+      def valid_structured_output_json?(value)
+        JSON.parse(value)
+        true
+      rescue JSON::ParserError, TypeError
+        false
       end
 
       def duplicate_structured_output_params(value)
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -582,44 +582,37 @@ module OpenAI
         raw[:output].to_a.none? do |output|
           case output[:type]
           when "message"
-            model && unfinished_structured_output_message?(output)
+            model && unfinished_structured_output_message?(output, response_status: raw[:status])
           when "function_call"
-            unfinished_structured_output_function_call?(output, tool_models)
+            unfinished_structured_output_function_call?(
+              output,
+              tool_models,
+              response_status: raw[:status]
+            )
           else
             false
           end
         end
       end
 
-      def unfinished_structured_output_message?(output)
+      def unfinished_structured_output_message?(output, response_status:)
         return true if structured_output_pending_status?(output[:status])
-        return false unless output[:status].nil?
+        return false unless response_status.nil? && output[:status].nil?
 
-        output[:content].to_a.any? do |content|
-          content[:type] == "output_text" &&
-            (!content.key?(:text) || !valid_structured_output_json?(content[:text]))
-        end
+        output[:content].to_a.any? { |content| content[:type] == "output_text" }
       end
 
-      def unfinished_structured_output_function_call?(output, tool_models)
+      def unfinished_structured_output_function_call?(output, tool_models, response_status:)
         return false if tool_models.empty?
-        return true unless output.key?(:name)
+        return false unless output.key?(:name)
         return false unless tool_models.key?(output[:name])
         return true if structured_output_pending_status?(output[:status])
-        return true unless output.key?(:arguments)
 
-        output[:status].nil? && !valid_structured_output_json?(output[:arguments])
+        response_status.nil? && output[:status].nil?
       end
 
       def structured_output_pending_status?(status)
         ["queued", "in_progress", "incomplete"].include?(status)
-      end
-
-      def valid_structured_output_json?(value)
-        JSON.parse(value)
-        true
-      rescue JSON::ParserError, TypeError
-        false
       end
 
       def duplicate_structured_output_params(value)
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -356,7 +356,7 @@ module OpenAI
       #
       # Retrieves a model response with the given ID.
       #
-      # @overload retrieve(response_id, include: nil, include_obfuscation: nil, starting_after: nil, request_options: {})
+      # @overload retrieve(response_id, include: nil, include_obfuscation: nil, starting_after: nil, text: nil, tools: nil, request_options: {})
       #
       # @param response_id [String] The ID of the response to retrieve.
       #
@@ -366,23 +366,41 @@ module OpenAI
       #
       # @param starting_after [Integer] The sequence number of the event after which to start streaming.
       #
+      # @param text [OpenAI::StructuredOutput::JsonSchemaConverter, nil] The structured-output model used to parse retrieved text output. This is a
+      #   local parsing hint and is not sent to the API.
+      #
+      # @param tools [Array<OpenAI::StructuredOutput::JsonSchemaConverter>, nil] Structured-output models used to parse retrieved function tool calls. These
+      #   are local parsing hints and are not sent to the API.
+      #
       # @param request_options [OpenAI::RequestOptions, Hash{Symbol=>Object}, nil]
       #
       # @return [OpenAI::Models::Responses::Response]
       #
       # @see OpenAI::Models::Responses::ResponseRetrieveParams
       def retrieve(response_id, params = {})
-        parsed, options = OpenAI::Responses::ResponseRetrieveParams.dump_request(params)
+        structured_output_params = duplicate_structured_output_params(params.slice(:text, :tools))
+        request_params = params.dup
+        request_params.delete(:text)
+        request_params.delete(:tools)
+
+        parsed, options = OpenAI::Responses::ResponseRetrieveParams.dump_request(request_params)
         query = OpenAI::Internal::Util.encode_query_params(parsed)
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
           raise ArgumentError.new(message)
         end
 
+        model, tool_models = get_structured_output_models(structured_output_params)
+
+        unwrap = -> (raw) do
+          parse_structured_outputs!(raw, model, tool_models)
+        end
+
         @client.request(
           method: :get,
           path: ["responses/%1$s", response_id],
           query: query,
+          unwrap: unwrap,
           model: OpenAI::Responses::Response,
           security: {bearer_auth: true},
           options: options
@@ -554,6 +572,17 @@ module OpenAI
 
       def get_structured_output_models(parsed)
         OpenAI::Helpers::StructuredOutput::ResponseParser.get_models(parsed)
+      end
+
+      def duplicate_structured_output_params(value)
+        case value
+        when Array
+          value.map { |item| duplicate_structured_output_params(item) }
+        when Hash
+          value.to_h { |key, item| [key, duplicate_structured_output_params(item)] }
+        else
+          value
+        end
       end
     end
   end
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -392,7 +392,7 @@ module OpenAI
 
         unwrap = if model || !tool_models.empty?
           -> (raw) do
-            next raw unless raw[:status] == "completed"
+            next raw if ["in_progress", "queued"].include?(raw[:status])
 
             parse_structured_outputs!(raw, model, tool_models)
           end
--- a/lib/openai/resources/responses.rb
+++ b/lib/openai/resources/responses.rb
@@ -392,7 +392,7 @@ module OpenAI
 
         unwrap = if model || !tool_models.empty?
           -> (raw) do
-            next raw unless [nil, "completed"].include?(raw[:status])
+            next raw unless structured_output_response_complete?(raw)
 
             parse_structured_outputs!(raw, model, tool_models)
           end
@@ -576,6 +576,14 @@ module OpenAI
         OpenAI::Helpers::StructuredOutput::ResponseParser.get_models(parsed)
       end
 
+      def structured_output_response_complete?(raw)
+        return raw[:status] == "completed" unless raw[:status].nil?
+
+        raw[:output].to_a.none? do |output|
+          ["queued", "in_progress", "incomplete"].include?(output[:status])
+        end
+      end
+
       def duplicate_structured_output_params(value)
         case value
         when Array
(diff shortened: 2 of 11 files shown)


lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,41 @@ def stream_raw(params = {})
...
+      #   are local parsing hints and are not sent to the API.
+      #
       # @param request_options [OpenAI::RequestOptions, Hash{Symbol=>Object}, nil]
       #
       # @return [OpenAI::Models::Responses::Response]
       #
       # @see OpenAI::Models::Responses::ResponseRetrieveParams
       def retrieve(response_id, params = {})
-        parsed, options = OpenAI::Responses::ResponseRetrieveParams.dump_request(params)
+        structured_output_params = duplicate_structured_output_params(params.slice(:text, :tools))
```
jbeckwith-oai: **[P1] Keep string-keyed parsing hints local.**

The SDK normally accepts string-keyed parameter hashes and normalizes them in `dump_request`, but this extraction/deletion handles only symbol keys. I reproduced `responses.retrieve(id, {"text" => CalendarEvent})` issuing `GET /responses/<id>?text=CalendarEvent`, which both leaks the supposedly local model name into the URL and sends an unsupported retrieval query parameter. Please normalize parameter keys before extracting/removing `text` and `tools`, and add regression coverage for both string-keyed hints.

jbeckwith-oai: **[P2] Preserve existing request-parameter model arguments.**

`responses.retrieve(id, OpenAI::Responses::ResponseRetrieveParams.new)` succeeds on the current base branch because `dump_request` accepts SDK request-model instances. This new unconditional `params.slice` call instead raises `NoMethodError: undefined method 'slice'` before any request is made. Please coerce supported parameter objects to a hash before inspecting local hints, and add a backward-compatibility regression test.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,41 @@ def stream_raw(params = {})
...
+        request_params.delete(:tools)
+
+        parsed, options = OpenAI::Responses::ResponseRetrieveParams.dump_request(request_params)
         query = OpenAI::Internal::Util.encode_query_params(parsed)
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
           raise ArgumentError.new(message)
         end
 
+        model, tool_models = get_structured_output_models(structured_output_params)
```
jbeckwith-oai: **[P2] Support the same typed text configurations accepted by creation.**

`config = OpenAI::Responses::ResponseTextConfig.new(format_: CalendarEvent)` correctly populates `content.parsed` when supplied to `responses.create(..., text: config)`, but `responses.retrieve(id, text: config)` silently returns `content.parsed == nil`. Creation normalizes the model and translates `format_` to `format` before extracting structured-output models; retrieval passes the unnormalized object straight into the parser. Please normalize supported typed configurations before model extraction and add a create/retrieve parity test.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,41 @@ def stream_raw(params = {})
...
+
+        unwrap = -> (raw) do
```
jbeckwith-oai: **[P2] Preserve the no-hints retrieval path.**

`unwrap` now invokes `ResponseParser.parse!` on every retrieval, including calls without `text` or `tools`. An in-progress response containing a partially populated `{type: "function_call"}` item succeeds on the base branch but now raises `KeyError: key not found: :name`, even though no tool parsing was requested. Please install the parser only when a usable text/tool model is present, and add a regression test showing ordinary retrieval remains backward-compatible.

jbeckwith-oai (changes requested):
Thanks for tackling #308! This is a thoughtful improvement, and I especially appreciate the clear background-response example, the focused completed/in-progress coverage, and the quick follow-up adding both RBI and RBS signatures.

I reviewed the updated head and confirmed the type-signature concern is already addressed. Before merging, please address the four remaining compatibility/correctness issues in the inline comments:

1. String-keyed local parsing hints currently leak into the HTTP query string.
2. Existing `ResponseRetrieveParams` objects now fail before request conversion.
3. Supported `ResponseTextConfig` instances that parse correctly during creation silently fail to parse during retrieval.
4. Retrieval without parsing hints now runs the structured-output parser and can reject previously accepted partial function-call output.

The existing focused suites still pass (114 tests, 627 assertions), but targeted regression cases reproduce each of these four issues. Please add coverage for these cases, preserve the existing no-hints/request-model behavior, and request another review once they are fixed. Thanks again for moving this feature forward!

cjc0013:
Addressed the four requested compatibility issues in 8b44e9cf:

- normalize retrieval params before extracting local string/symbol parsing hints
- preserve ResponseRetrieveParams arguments
- normalize typed ResponseTextConfig values for create/retrieve parity
- skip structured-output parsing when retrieval has no usable hints

Added focused regressions for each case. Fresh verification: 26 runs, 171 assertions, 0 failures/errors; rubyfmt 0.14.1 clean; RuboCop clean. Ready for another review.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,41 @@ def stream_raw(params = {})
...
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
           raise ArgumentError.new(message)
         end
 
+        model, tool_models = get_structured_output_models(structured_output_params)
+
+        unwrap = if model || !tool_models.empty?
+          -> (raw) do
+            parse_structured_outputs!(raw, model, tool_models)
```
jbeckwith-oai: **[P1] Defer structured-output parsing until the background response is complete.**

This callback currently parses every response whenever a text/tool hint is present, including `status: "in_progress"` responses that already contain partial output. I reproduced three failures: partial output-text JSON raises `OpenAI::Errors::ConversionError` when `content.parsed` is accessed; partial function arguments raise the same error when `output.parsed` is accessed; and an unfinished `{type: "function_call"}` without `name` raises `KeyError` during retrieval even when only a text hint was provided. The current pending-response regression has `output: []`, so it does not cover this normal background-polling state. Please run the parser only for completed responses and add focused in-progress regressions containing partial text and function-call output.

rbi/openai/resources/responses.rbi
```diff
@@ -876,6 +876,13 @@ module OpenAI
           include: T::Array[OpenAI::Responses::ResponseIncludable::OrSymbol],
           include_obfuscation: T::Boolean,
           starting_after: Integer,
+          text: T.nilable(
+            T.any(
+              OpenAI::Responses::ResponseTextConfig::OrHash,
+              OpenAI::StructuredOutput::JsonSchemaConverter::Input
+            )
+          ),
+          tools: T.nilable(T::Array[OpenAI::StructuredOutput::JsonSchemaConverter::Input]),
```
jbeckwith-oai: **[P2] Include supported explicitly named tool definitions in the retrieval signature.**

At runtime, `responses.retrieve(id, tools: [{type: :function, function: {name: "custom_lookup", parameters: LookupModel}}])` successfully parses the matching function call and preserves the caller-owned definition. However, this signature permits only converter objects/classes, so Sorbet rejects that working call with `Expected T.nilable(T::Array[T.any(T::Class[T.anything], ...JsonSchemaConverter)]) but found ... Hash`. This also prevents a typed caller from reusing an explicitly named tool definition across create and retrieve. Please widen the retrieval `tools:` signature to include the supported tool/hash representation and add a focused Sorbet regression.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,41 @@ def stream_raw(params = {})
...
+        structured_output_params = duplicate_structured_output_params(parsed.slice(:text, :tools))
+        parsed.delete(:text)
+        parsed.delete(:tools)
         query = OpenAI::Internal::Util.encode_query_params(parsed)
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
           raise ArgumentError.new(message)
         end
 
+        model, tool_models = get_structured_output_models(structured_output_params)
```
jbeckwith-oai: **[P2] Preserve the newly merged optional Sorbet adapter's supported-feature boundary.**

Current main now includes #537, which supports `OpenAI::StructuredOutput.from_sorbet(MyStruct)` for non-streaming text output but intentionally rejects Sorbet function tools. Testing this PR together with that merged helper showed `retrieve(id, text: adapter)` correctly hydrates the `T::Struct`, but `retrieve(id, tools: [adapter])` also succeeds instead of rejecting the explicitly unsupported function-tool case. Please update against current main and add focused regressions showing Sorbet text retrieval works while Sorbet function tools remain unsupported; avoid expanding the adapter's public feature surface or changing BaseModel/create/stream semantics.

jbeckwith-oai (changes requested):
Thank you for the thoughtful follow-up and for carefully addressing the previous compatibility issues. The normalized string-keyed hints, preserved request-parameter models, typed text configuration, no-hints fast path, and updated signatures are all meaningful improvements.

I re-reviewed the latest head against the original background-retrieval requirements and the newer Sorbet integration. There are three remaining items to address before this is ready:

1. Do not attempt structured-output parsing while a background response is still queued or in progress; partial text/function-call output can currently raise during normal polling.
2. Make the shipped Sorbet signature accept the explicitly named tool definitions that retrieval already supports at runtime.
3. Update against current main and preserve the optional Sorbet adapter's existing boundary: typed text retrieval should work, but intentionally unsupported Sorbet function tools must not become supported accidentally.

Please keep the changes narrowly scoped; there is no need to change the existing canonical flat-tool parser behavior or refactor create/stream handling. Thanks again for sticking with this and for improving the regression coverage!

lib/openai/resources/responses.rb
cjc0013: Thank you for the detailed follow-up. Addressed the three remaining items in 3d4f6cae after updating to current main:

- retrieval now defers structured-output parsing until `status == "completed"`, returning queued/in-progress partial output without parsing
- the shipped `retrieve` RBI now accepts explicitly named `FunctionTool::OrHash` definitions alongside converter inputs
- the optional Sorbet boundary is preserved: completed text retrieval hydrates the `T::Struct`, while Sorbet function-tool adapters are rejected before any request

Added regressions for partial output text, partial function arguments, unfinished unnamed function calls, explicitly named retrieval tools, Sorbet text hydration, and unsupported Sorbet function tools.

Verification:
- topology-adjacent Responses/structured-output suite: 12 files, 93 runs, 502 assertions, 0 failures/errors/skips
- shipped RBI and examples Sorbet typecheck: no errors
- rubyfmt 0.14.1: clean
- RuboCop: 5 files inspected, no offenses
- Castiron public custom-code report: no new custom-code files; 3,102 / 4,000 custom lines, leaving 898 lines of headroom

I also attempted the full non-Bedrock suite on Windows. It completed 1,508 runs and 13,184 assertions with 33 failures and 30 errors, all outside Responses retrieval, structured outputs, Sorbet, and the changed files. Those failures were in platform/process/network and repository-workflow tests, so I am not representing that broader run as green.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,43 @@ def stream_raw(params = {})
...
+
+        unwrap = if model || !tool_models.empty?
+          -> (raw) do
+            next raw unless raw[:status] == "completed"
```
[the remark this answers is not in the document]
cjc0013: `Response#status` is optional, so requiring `raw[:status] == "completed"` skipped otherwise-valid retrieved responses. I changed the guard to defer only the known pending statuses (`in_progress` and `queued`) and added a regression test where the top-level status is omitted but the output message is completed. The existing in-progress and queued deferral cases continue to pass.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,43 @@ def stream_raw(params = {})
...
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
           raise ArgumentError.new(message)
         end
 
+        model, tool_models = get_structured_output_models(structured_output_params)
+
+        unwrap = if model || !tool_models.empty?
+          -> (raw) do
+            next raw if ["in_progress", "queued"].include?(raw[:status])
```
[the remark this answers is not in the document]
cjc0013: Updated in d45bb9dd. Retrieval now parses only responses whose top-level status is completed or omitted; any explicit non-completed status, including incomplete, is returned without attempting to parse partial text or function arguments. Added an incomplete regression covering both output forms.

lib/openai/resources/responses.rb
```diff
@@ -365,23 +365,43 @@ def stream_raw(params = {})
...
+        structured_output_params = duplicate_structured_output_params(parsed.slice(:text, :tools))
+        parsed.delete(:text)
+        parsed.delete(:tools)
         query = OpenAI::Internal::Util.encode_query_params(parsed)
         if parsed[:stream]
           message = "Please use `#retrieve_streaming` for the streaming use case."
           raise ArgumentError.new(message)
         end
 
+        model, tool_models = get_structured_output_models(structured_output_params)
```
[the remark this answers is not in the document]
cjc0013: Updated in d45bb9dd. Retrieval now recursively normalizes string keys in its duplicated local hint tree before model extraction. The regression now uses a nested string-keyed named function tool and also verifies the caller-owned hash is unchanged.

lib/openai/resources/responses.rb
```diff
@@ -554,6 +574,20 @@ def parse_structured_outputs!(raw, model, tool_models)
...
       end
+
+      def duplicate_structured_output_params(value)
+        case value
+        when Array
+          value.map { |item| duplicate_structured_output_params(item) }
+        when Hash
+          value.to_h do |key, item|
+            normalized_key = key.is_a?(String) ? key.to_sym : key
+            [normalized_key, duplicate_structured_output_params(item)]
```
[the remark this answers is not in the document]
cjc0013: Updated in e63c1dcc. The retrieval-local copy now normalizes only the exact type="function" discriminator to the function symbol; it does not alter other string values or the canonical create/stream parser. Added a flattened string-keyed named-tool regression and verified the caller-owned hint remains unchanged. The 12-file topology suite passes 96 runs and 513 assertions with no failures, errors, or skips.

cjc0013:
Updated in 68fc903e. Retrieval now:

- defers structured-output parsing when the top-level status is omitted but an output item is queued, in_progress, or incomplete, including the unfinished function-call case
- normalizes symbol- and string-keyed format_ hints to format, plus the string json_schema discriminator, on the retrieval-local copy
- documents that retrieved structured output is parsed only after completion

Added focused regressions for each path while preserving caller-owned hint hashes.

Validation:
- bundle exec rake test TEST=test/openai/helpers/{structured_output_api_names,sorbet_structured_output}_test.rb: 32 runs, 127 assertions, 0 failures, 0 errors, 0 skips
- bundle exec rubocop --force-exclusion lib/openai/resources/responses.rb test/openai/helpers/structured_output_api_names_test.rb: 2 files inspected, no offenses

lib/openai/resources/responses.rb
```diff
@@ -554,6 +574,33 @@ def parse_structured_outputs!(raw, model, tool_models)
       def get_structured_output_models(parsed)
         OpenAI::Helpers::StructuredOutput::ResponseParser.get_models(parsed)
       end
+
+      def structured_output_response_complete?(raw)
+        return raw[:status] == "completed" unless raw[:status].nil?
```
jbeckwith-oai: **[P1] Check item completion even when the response is completed.**

`structured_output_response_complete?` returns immediately when the top-level status is `"completed"`, so it never inspects output-item statuses in that case. A response shaped like `{status: "completed", output: [{status: "incomplete", ...}]}` therefore reaches `ResponseParser.parse!`; truncated JSON can surface as a conversion error, and an unfinished function call can raise while retrieval is supposed to preserve partial output. Please reject queued/in-progress/incomplete output items before parsing regardless of the top-level status, and add a focused regression for a completed response containing an incomplete text or function-call item.

jbeckwith-oai (changes requested):
Thank you for the careful, iterative follow-up on this PR. We really appreciate how much attention you have given the compatibility cases, the Sorbet boundary, the documentation, and the focused regression coverage—it is clear that a lot of thoughtful work went into making this feature safe for SDK users.

I reviewed the latest head again and resolved all 17 outstanding threads that are now addressed by the current implementation and tests. There is one remaining correctness issue in the new inline comment: a top-level completed response can still contain an incomplete output item, and that item is parsed because the completion gate returns before checking item status.

Focused verification on Ruby 4.0.6:
- `test/openai/helpers/structured_output_api_names_test.rb`: 28 runs, 109 assertions, passing
- `test/openai/helpers/sorbet_structured_output_test.rb`: 4 runs, 18 assertions, passing
- RuboCop on the four changed Ruby/test files: no offenses
- `./scripts/rubyfmt --check` still reports small formatting diffs in `lib/openai/resources/responses.rb` and `test/openai/helpers/structured_output_api_names_test.rb`

Once the completion gate and its regression are updated, plus the formatter output is applied, this should be very close. Thanks again—we genuinely appreciate your persistence and the quality of the follow-through here.

lib/openai/resources/responses.rb
cjc0013: Updated in 953345da. The completion gate now checks output-item statuses regardless of whether the top-level status is omitted or completed, so queued, in-progress, and incomplete items remain unparsed. Added a combined regression covering a completed response with incomplete text and an unfinished function call, and applied the pinned rubyfmt output.

Validation:
- focused structured-output suites: 33 runs, 131 assertions, 0 failures, 0 errors, 0 skips
- `./scripts/rubyfmt --check`: passed
- RuboCop on the changed Ruby/test files: no offenses
- gem build: passed

lib/openai/resources/responses.rb
```diff
@@ -555,6 +575,34 @@ def parse_structured_outputs!(raw, model, tool_models)
       def get_structured_output_models(parsed)
         OpenAI::Helpers::StructuredOutput::ResponseParser.get_models(parsed)
       end
+
+      def structured_output_response_complete?(raw)
+        return false unless [nil, "completed"].include?(raw[:status])
+
+        raw[:output].to_a.none? do |output|
+          ["queued", "in_progress", "incomplete"].include?(output[:status])
```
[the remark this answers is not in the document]
jbeckwith-oai: Addressed in 471af2f7.

Retrieval now defers a statusless function call when the fields required for parsing are absent, and it also defers statusless partial JSON arguments/text. The readiness check is hint-aware, so unrelated malformed output no longer blocks parsing of completed hinted text or tools. I also made the shared parser skip function-call inspection entirely when no tool models were supplied.

Added focused regressions for:
- a standalone statusless unfinished function call
- statusless partial function arguments
- statusless partial output text
- valid hinted text beside an unhinted partial function call
- valid hinted tools beside unhinted partial text

Verification on Ruby 4.0.6: the seven adjacent Responses/structured-output suites pass (91 runs, 389 assertions); rubyfmt, RuboCop, diff hygiene, aggregate lint/typecheck, security review, and two consecutive clean adversarial-review rounds all passed.

lib/openai/helpers/sorbet.rb
```diff
@@ -259,6 +259,11 @@ def create(...)
           super
         end
 
+        def retrieve(response_id, params = {})
+          reject_sorbet_function_tools!(params)
```
HAYDEN-OAI, line 263: [P2] Copy tool hints before the Sorbet retrieval guard

When `openai/helpers/sorbet` is loaded, `responses.retrieve(id, tools: [LookupArguments].freeze)` with an ordinary supported `OpenAI::BaseModel` tool class raises `FrozenError` before sending the GET. This new wrapper invokes `reject_sorbet_function_tools!` on the original parameters; that helper retains the array with `Array(params[:tools])` and then calls `concat`, even when the string-keyed tools value is absent and the appended array is empty. The resource's later defensive copy is never reached. Build a fresh array in the guard, or inspect both key forms without mutating either caller-owned array, so loading the optional helper does not break ordinary frozen tool hints.

HAYDEN-OAI (commented):
One compatibility issue remains in the new Sorbet retrieval wrapper: frozen arrays of ordinary tool hints fail before the GET. See the inline P2 comment. Static source review of 471af2f7d.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Addressed in 0bcfc763 after merging current main.

The Sorbet retrieval guard now duplicates the symbol-keyed tools collection before concatenating the string-keyed form, so it never mutates a caller-owned frozen array. Added a focused regression that loads the optional helper and retrieves with a frozen ordinary structured-output tool-hint array.

Verification on Ruby 4.0.6: the Sorbet structured-output suite passes; the adjacent Responses/structured-output selection passes; rubyfmt, RuboCop, diff hygiene, aggregate lint/typecheck, and two consecutive clean adversarial-review rounds all passed.

lib/openai/resources/responses.rb
```diff
@@ -555,6 +575,72 @@ def parse_structured_outputs!(raw, model, tool_models)
...
+      end
+
+      def unfinished_structured_output_function_call?(output, tool_models)
+        return false if tool_models.empty?
+        return true unless output.key?(:name)
+        return false unless tool_models.key?(output[:name])
+        return true if structured_output_pending_status?(output[:status])
+        return true unless output.key?(:arguments)
+
+        output[:status].nil? && !valid_structured_output_json?(output[:arguments])
```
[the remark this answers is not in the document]
HAYDEN-OAI: Confirmed in this re-review at `0bcfc763b0`: the response-wide guard also leaves otherwise valid hinted text/tool siblings unparsed. Please preserve explicit pending item statuses, but let an explicitly completed response reach the shared parser when the item status is absent, so malformed final arguments surface through the existing conversion-error path.

HAYDEN-OAI (commented):
The frozen-array Sorbet guard issue is fixed. One P2 remains in the existing [completed-response parsing thread](https://github.com/openai/openai-ruby/pull/532#discussion_r3899036096): a completed response with malformed arguments and an omitted function-call status skips structured-output parsing, including otherwise valid hinted output. I independently confirmed that path. No additional findings from the static re-review of 0bcfc763b0.

lib/openai/resources/responses.rb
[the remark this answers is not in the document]
jbeckwith-oai: Addressed in e40d496f after rebasing onto current main.

Retrieval now treats an explicitly completed top-level response as final even when an output item omits its optional status, so malformed final function arguments reach the existing parser and surface the normal conversion error instead of remaining unparsed. When both response and item statuses are omitted, parser-relevant structured text and named hinted tool calls still defer because completion is ambiguous. Nameless tool calls are ignored for hinted-tool parsing and no longer block valid hinted text siblings.

Added focused regressions for completed malformed statusless arguments/text, schema-incomplete fully statusless arguments/text, and valid hinted text beside a nameless tool call. Verification on Ruby 4.0.6: seven adjacent Responses/structured-output suites pass, full rake lint/typecheck passes, diff hygiene is clean, and two consecutive fresh adversarial-review rounds were clean on the rebased diff.

lib/openai/resources/responses.rb
```diff
@@ -555,6 +575,65 @@ def parse_structured_outputs!(raw, model, tool_models)
...
+        output[:content].to_a.any? { |content| content[:type] == "output_text" }
+      end
+
+      def unfinished_structured_output_function_call?(output, tool_models, response_status:)
+        return false if tool_models.empty?
+        return false unless output.key?(:name)
+        return false unless tool_models.key?(output[:name])
+        return true if structured_output_pending_status?(output[:status])
+
+        response_status.nil? && output[:status].nil?
```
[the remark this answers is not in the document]
jbeckwith-oai: Addressed in 8a794253.

For the both-statuses-omitted path, retrieval now checks whether the JSON is syntactically complete and exactly viable for the hinted schema. Complete schema-valid function arguments/text are parsed; malformed, schema-incomplete, or missing-required-nullable payloads continue to defer. Explicit pending statuses still defer, while an explicitly completed top-level response still reaches the shared parser so malformed final JSON surfaces its normal conversion error.

I also fixed the CI-discovered shared-parser compatibility regression by keeping missing-key errors in the shared parser and filtering nameless calls only in retrieval, plus covered typed string-keyed Sorbet tool hints. Added focused regressions for schema-valid statusless text/arguments and missing required nullable fields. Verification: targeted parser/retrieval/Sorbet suites pass, full rake lint/typecheck passes, diff hygiene is clean, and two consecutive fresh adversarial-review rounds are clean.

lib/openai/resources/responses.rb
```diff
@@ -552,9 +572,92 @@ def parse_structured_outputs!(raw, model, tool_models)
...
+      end
+
+      def structured_output_pending_status?(status)
+        ["queued", "in_progress", "incomplete"].include?(status)
+      end
+
+      def structured_output_value_ready?(value, model)
+        parsed = JSON.parse(value, symbolize_names: true)
+        state = OpenAI::Internal::Type::Converter.new_coerce_state
+        OpenAI::Internal::Type::Converter.coerce(model, parsed, state: state)
```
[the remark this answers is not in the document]
HAYDEN-OAI: Confirmed at 8a79425319. When both the response and a relevant item omit status, a complete schema-valid payload reaches `structured_output_value_ready?`, which JSON-parses and fully converts it but retains only the exactness counters. The successful unwrap then calls the shared `ResponseParser`, which parses and converts the same payload again. This is two full materialization passes for supported statusless structured output, including nested model/array payloads; retaining the readiness result would avoid the redundant work. Explicitly completed statuses bypass this readiness conversion, so the finding is limited to the both-statuses-absent path.

HAYDEN-OAI (commented):
The earlier frozen-array and completed/statusless parsing issues are fixed in 8a79425319. I confirmed one remaining P2 in the existing [statusless parsing thread](https://github.com/openai/openai-ruby/pull/532#discussion_r3906823343): successful readiness discards a fully parsed and converted result, and the final parser repeats that work. I added the exact execution path there rather than duplicating the finding. No additional substantive findings from this static re-review.

lib/openai/resources/responses.rb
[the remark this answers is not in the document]
jbeckwith-oai: Addressed in 8a2bfb40. Statusless readiness now performs only the JSON syntax parse and retains the decoded value. Once every hinted output in the response is ready, retrieval converts each staged value exactly once and stores it directly, instead of reparsing/recoercing through the shared parser. A mixed ready+partial regression also verifies no conversion runs until the whole response is ready, and a Sorbet regression verifies syntactically complete invalid JSON surfaces normal hydration errors rather than being treated as partial.

jbeckwith-oai (approved):

sylvesterkaczmarek (commented):
Checked the current head after the earlier retrieval fixes. `text`/`tools` are removed before query encoding, typed/hash configs go through the retrieval normalization path, and statusless values stay staged until the hinted output is ready. The latest path also avoids the earlier double conversion. I don't see another blocker here.

PR #537, opened 2026-08-26
feat: add optional Sorbet structured-output models

jbeckwith-oai:
## Summary

- Add an explicitly loaded, optional Sorbet structured-output adapter: `require "openai/helpers/sorbet"` and `OpenAI::StructuredOutput.from_sorbet(MyStruct)`.
- Generate strict JSON Schema for application-owned `T::Struct` models, including nested structs, arrays, nullable fields, booleans, numeric fields, serialized field aliases, and actual `T::Enum` values.
- Hydrate non-streaming Responses and Chat Completions directly into checked application-owned models without changing `OpenAI::BaseModel`, generated resources/models, shared parsers, transport semantics, or runtime dependencies.
- Document the supported model subset, optional dependency, explicit casts, and intentionally unsupported streaming/function-tool behavior.

Refs #309.

## Compatibility and security

- Ordinary SDK loading does not require or load `sorbet-runtime`; the gem's runtime dependency list is unchanged.
- Existing transport-model raw-value identity, mutation, dumping, and serialization semantics remain unchanged.
- Unknown fields, missing required fields, ambiguous wire names, unsupported unions/hashes/recursive models, and invalid enum values fail explicitly.
- Hydration diagnostics identify the field path and expected type without including untrusted response contents.
- The generated/custom-code budget passes without changing its policy or limit.

## Verification

- Full repository suite: 1,425 tests, 13,078 assertions, 0 failures, 0 errors; one existing skip.
- Optional-helper integration and public Sorbet signature checks on Ruby 3.3, 3.4, and 4.0.
- Repository-wide formatting, RuboCop, Sorbet, and RBS validation: clean.
- Published-gem contents and unchanged runtime dependencies verified.
- Independent adversarial architecture and deserialization/security review completed.

## Scope

This first change supports non-streaming `client.responses.create` and `client.chat.completions.create` only. Streaming structured outputs, function-tool adaptation, broader union support, and generated-signature parity remain separate follow-up work.


lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,242 @@
...
+            unless keys.one?
+              reason = keys.empty? ? "missing required field" : "duplicate field"
+              raise HydrationError, "#{field_path}: #{reason}"
+            end
+
+            [field.name, hydrate(field.node, value.fetch(keys.first), path: field_path)]
+          end
+
+          model.new(**attributes)
+        rescue TypeError, ArgumentError => e
```
HAYDEN-OAI: **[P1] Sanitize every ordinary constructor failure, not only type errors**

Application-owned `T::Struct#initialize` can raise any `StandardError`, but this rescue handles only `TypeError` and `ArgumentError`. For example, a model constructor calling `super` and then `raise RuntimeError, "model-output=#{secret}"` propagates its untrusted/private response content verbatim through **both** Responses and Chat Completions parsing; neither public parser sanitizes it. This violates the documented no-response-content diagnostic boundary (the existing check exercises `ArgumentError` only). Rescue ordinary `StandardError`, preserve/re-raise existing `HydrationError`, and replace all other constructor failures with the scrubbed `HydrationError` using `cause: nil`.

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,242 @@
...
+            end
+
+          in :boolean
+            unless value.equal?(true) || value.equal?(false)
+              fail_hydration(path, expected: "Boolean", actual: value)
+            end
+
+            value
+          in :scalar
+            value = value.to_f if node.value == Float && value.is_a?(Integer)
```
HAYDEN-OAI, line 205: **[P2] Reject integer-to-float conversions that overflow to infinity**

Valid JSON integers have no fixed magnitude, so a model output containing a 400-digit integer in a schema `number` field reaches this branch and `value.to_f` becomes `Float::INFINITY`. It then passes `is_a?(Float)`, is installed in the supposedly checked `T::Struct`, and cannot be serialized back to JSON (`JSON::GeneratorError: Infinity not allowed in JSON`), breaking downstream numeric/business invariants. Verify the converted float is finite and raise the existing redacted hydration error on overflow.

HAYDEN-OAI (commented):
Found one sensitive-output disclosure and two additional structured-output contract/integrity issues.

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,242 @@
...
+
+module OpenAI
+  module Helpers
+    module StructuredOutput
+      # Adapts application-owned Sorbet models to structured output requests.
+      #
+      # Load explicitly with `require "openai/helpers/sorbet"`.
+      class SorbetAdapter
+        include OpenAI::Internal::Type::Converter
+        include OpenAI::Helpers::StructuredOutput::JsonSchemaConverter
```
HAYDEN-OAI, line 22: **[P2] Reject unsupported Chat Completions streaming before opening a request**

Including `JsonSchemaConverter` also makes this adapter match the existing streaming parser. `client.chat.completions.stream(response_format: adapter, ...)` therefore accepts it, generates/sends a billable strict-schema request, and constructs `ChatCompletionStreamState`; that state only treats a **Class** as a rich response format, so `content.done`/the final completion expose a raw `Hash` instead of the promised application `T::Struct`. README declares streaming unsupported, but unsupported inputs should fail explicitly before network dispatch, not silently return unvalidated/unhydrated output after charging for a request. Reject Sorbet adapters at the unsupported streaming boundary (or fully implement streaming hydration).

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,242 @@
...
+            end
+
+            [field.name, hydrate(field.node, value.fetch(keys.first), path: field_path)]
+          end
+
+          model.new(**attributes)
+        rescue TypeError, ArgumentError => e
+          raise if e.is_a?(HydrationError)
+
+          raise HydrationError, "#{path}: invalid #{model.name} structured output", cause: nil
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 65d17ab1 by narrowing the rescue to the checked `model.new(**attributes)` constructor call. `model` is therefore initialized before the rescue can reference it; existing `HydrationError` instances still propagate unchanged, and wrapped constructor errors continue to suppress causes so response values cannot leak. Added a focused regression test for hydration-error identity.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Fixed in 7669a5b4: constructor error translation now rescues every ordinary `StandardError`, preserves existing `HydrationError` instances unchanged, and raises a sanitized `HydrationError` with `cause: nil` for all other constructor failures. The regression now covers both `ArgumentError` and a secret-bearing `RuntimeError`, including `full_message` and exception-cause redaction.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Fixed in 7669a5b4: Float hydration now requires the resulting value to be finite after integer conversion. Focused regressions cover a 400-digit JSON integer overflowing to infinity, `Float::INFINITY`, and `Float::NAN`; each fails with the existing redacted, path-specific `HydrationError` and no exception cause.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Intentionally deferred to preserve the architecture approved for this first PR: the Sorbet adapter explicitly supports only non-streaming `responses.create` and `chat.completions.create`, and that limitation is documented in the README. There is no adapter-owned hook before Chat Completions streaming dispatch; fail-fast rejection would require changing the generated resource, modifying the shared structured-output parser, or monkey-patching a global resource, all of which would violate the agreed no-generated/no-shared-parser scope. Streaming validation or full hydration should be designed and approved as a separate follow-up.

README.md
```diff
@@ -761,6 +761,56 @@ response = client.responses.create(
...
+client = OpenAI::Client.new
+
+response = client.responses.create(
+  model: "gpt-5.2",
+  input: "Extract the event information.",
+  text: schema
+)
+
+event = T.cast(response.output.first.content.first.parsed, CalendarEvent)
+event.participants.first.name
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 1428188c. The README now narrows response items with `grep(OpenAI::Responses::ResponseOutputMessage)` and output content with `grep(OpenAI::Responses::ResponseOutputText)`, uses `fetch(0)` for non-null array access, and uses `fetch(0)` for Chat Completions choices and typed participants. The shipped-RBI regression now typechecks the real documented response traversal and both endpoint calls under Sorbet `# typed: strict` instead of casting a synthetic `nil`.

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,244 @@
...
+
+        private def compile_struct(model, path:, ancestors:)
+          if ancestors.include?(model)
+            raise ArgumentError, "#{path}: recursive Sorbet models are not supported"
+          end
+
+          fields = model.props.map do |field_name, rules|
+            wire_name = rules.fetch(:serialized_form)
+            type = rules.fetch(:type_object)
+            child = compile_type(type, path: "#{path}.#{field_name}", ancestors: [*ancestors, model])
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 1428188c using the SDK's existing canonical schema-converter machinery. Sorbet structs are memoized by model class during compilation, while actual recursive ancestor cycles remain rejected; object schemas now use `JsonSchemaConverter.cache_def!` so repeated nested classes share escaped `$defs`/`$ref` entries rather than recursively duplicating their schemas. Added a multi-level repeated-model regression that verifies each struct is compiled once and both parent/child references share the expected definitions.

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,256 @@
...
+              reason = keys.empty? ? "missing required field" : "duplicate field"
+              raise HydrationError, "#{field_path}: #{reason}"
+            end
+
+            [field.name, hydrate(field.node, value.fetch(keys.first), path: field_path)]
+          end
+
+          begin
+            model.new(**attributes)
+          rescue HydrationError
```
HAYDEN-OAI: **[P1] Sanitize application-raised `HydrationError` too**

This `begin` encloses only the application-owned `model.new(**attributes)` call, so any `HydrationError` caught here was produced by application-controlled constructor code, not by this adapter's field validation (which runs before the `begin`). A constructor can therefore `raise OpenAI::StructuredOutput::SorbetAdapter::HydrationError, "model-output=#{secret}"`, and this special case re-raises the private response content verbatim through both Responses and Chat Completions parsers, bypassing the intended redaction fix. The new `test_constructor_hydration_errors_are_preserved` actually codifies the bypass. Remove this special case and sanitize **all** `StandardError` raised by the constructor into a fresh `HydrationError` with `cause: nil`; adapter-generated validation errors outside the constructor will still propagate normally.

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,256 @@
...
+
+module OpenAI
+  module Helpers
+    module StructuredOutput
+      # Adapts application-owned Sorbet models to structured output requests.
+      #
+      # Load explicitly with `require "openai/helpers/sorbet"`.
+      class SorbetAdapter
+        include OpenAI::Internal::Type::Converter
+        include OpenAI::Helpers::StructuredOutput::JsonSchemaConverter
```
HAYDEN-OAI, line 22: **[P2] Unsupported streaming still sends a billable request and silently skips hydration**

The previous streaming thread was marked resolved, but its response confirms the behavior was intentionally deferred and the current head still accepts `client.chat.completions.stream(response_format: OpenAI::StructuredOutput.from_sorbet(Model), ...)`. Because this adapter includes `JsonSchemaConverter`, `Completions#stream` converts it into a strict schema and invokes `@client.request` before constructing `ChatCompletionStreamState`; that state accepts rich output only when `response_format.is_a?(Class)`, so the streamed/final parsed value remains a raw, unvalidated `Hash`. A README limitation does not prevent the billable request or preserve the advertised checked-model contract. Reject this adapter at a pre-dispatch streaming boundary, or implement streaming hydration.

HAYDEN-OAI (commented):
Re-reviewed the updated head: integer-to-float overflow and ordinary constructor exceptions are now handled, and schema definitions are reused, but the constructor-redaction fix still has a sensitive-output bypass and the previously identified unsupported streaming path still dispatches billable requests.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Fixed in 8a3295d4. The constructor boundary no longer special-cases `HydrationError`: every application-raised `StandardError`, including an application-created `HydrationError` containing private model output, is replaced with a fresh path-aware `HydrationError` using `cause: nil`. Adapter-generated field-validation errors still occur before the constructor boundary and propagate normally. The regression now verifies that a secret-bearing constructor `HydrationError` is not preserved and cannot appear in `full_message` or `cause`.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Fixed in 8a3295d4 at the existing handwritten `ChatCompletionParser.get_models` request-preparation boundary. Streaming already supports only class-based rich response models, so object-based schema adapters now raise `ArgumentError` before `@client.request` is invoked. This is a generic invariant check with no Sorbet constant, generated-resource change, monkey patch, or change to existing class-based streaming. Added a real `client.chat.completions.stream` regression that verifies the clear error and asserts the POST request was never sent; the existing 21 Chat streaming tests and all three nested response-format shapes also remain green.

lib/openai/helpers/structured_output/chat_completion_parser.rb
```diff
@@ -110,6 +110,10 @@ def self.get_models(parsed)
           else
           end
 
+          if parsed[:stream] && model && !model.is_a?(Class)
```
HAYDEN-OAI: **[P1] Reject Sorbet adapters on Responses streaming too**

This guard fixes only `chat.completions.stream`; the equally unsupported `client.responses.stream(model: ..., input: ..., text: adapter)` still passes the adapter through `ResponseParser.get_models` and sends a billable POST (`resources/responses.rb:204-243`). In particular, `stream` sets `parsed[:stream] = true` only **after** model extraction, so copying this `parsed[:stream]` check into `ResponseParser` would not work without adjusting call order. The stream then calls `ResponseStreamState#parse_structured_text`; if generated output is malformed JSON, its error embeds `Raw text: #{text.inspect}`, exposing the complete potentially sensitive model output despite this adapter's documented redaction boundary. Reject object-based Sorbet adapters on the Responses streaming path **before** opening/retrieving a request and add the corresponding no-dispatch regression.

lib/openai/helpers/sorbet.rb
```diff
@@ -0,0 +1,254 @@
...
+            end
+
+            value.map.with_index { |item, index| hydrate(node.value, item, path: "#{path}[#{index}]") }
+          in :enum
+            unless value.is_a?(String)
+              fail_hydration(path, expected: node.value.name, actual: value)
+            end
+
+            begin
+              node.value.deserialize(value)
```
HAYDEN-OAI, line 193: **[P1] Scrub exceptions from application-owned enum deserializers**

`node.value` is an application-owned `T::Enum` subclass, so its `.deserialize` implementation can run application code and raise any `StandardError`, not just `KeyError`. For example, `def self.deserialize(value); raise RuntimeError, "model-output=#{value}"; end` causes the sensitive model-provided enum value to propagate verbatim through both Responses and Chat Completions parsers. This happens while assembling field attributes, before the newly hardened `model.new` rescue, so constructor sanitization cannot catch it. Treat enum deserialization as the same application-code boundary: sanitize every ordinary failure into a path-aware `HydrationError` with `cause: nil`, without including the response value.

HAYDEN-OAI (commented):
Re-reviewed the latest head: constructor exception redaction, finite numeric hydration, and Chat Completions streaming are fixed. Two distinct sensitive-output exposure paths remain through unsupported Responses streaming and application-owned enum deserialization.

lib/openai/helpers/structured_output/chat_completion_parser.rb
jbeckwith-oai: Fixed in c10a9d6e. The shared Chat parser change is gone; the explicitly loaded handwritten Sorbet adapter now guards both Responses and Chat stream and stream_raw before POST or retrieval. Regression coverage includes direct and nested text formats, function tools, string-keyed request hashes, generated typed request-model wrappers, and no-dispatch assertions. No generated resource or Castiron compiler changes are required.

lib/openai/helpers/sorbet.rb
jbeckwith-oai: Fixed in c10a9d6e. Enum deserialization now catches every StandardError and converts it into a path-aware HydrationError without response values or an exception cause. Field-validation errors also explicitly clear implicit active causes. Tests cover RuntimeError, ArgumentError, application-raised HydrationError, and secret-bearing active exceptions.

lib/openai/helpers/structured_output/chat_completion_parser.rb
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in c10a9d6e. The optional Sorbet request guard rejects adapter-backed function tools before dispatch for both Chat and Responses streaming, including stream_raw, nested and string-keyed function definitions, and typed SDK tool objects. Because function-tool schemas are outside this approved text-only integration, nonstreaming create rejects them too.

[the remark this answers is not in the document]
jbeckwith-oai: Fixed in c10a9d6e. Removed the generic shared-parser class check entirely and moved preflight into the explicitly loaded Sorbet helper. Only actual Sorbet adapters are rejected; an existing instance-based JsonSchemaConverter is covered by a successful Chat streaming regression. The shared parser is now identical to main.

HAYDEN-OAI (approved):
Re-reviewed the updated Sorbet integration and confirmed all prior substantive findings are fixed. Application-owned enum deserializers, model constructors, and field failures now produce scrubbed, cause-free hydration errors; integer-to-float conversion rejects non-finite values; and unsupported Sorbet response formats and function tools are rejected before both Responses and Chat Completions streaming/raw-streaming dispatch or retrieval. Existing non-Sorbet instance-based converters remain compatible. No remaining substantive concerns.

PR #539, opened 2026-08-26
feat(api): make function call output call IDs optional

apcha-oai:
## Summary

Align function call output types with the API schema: `call_id` is optional in stable and beta input and response types, and input types also accept `null`.


rbi/openai/models/responses/response_function_tool_call_output_item.rbi
```diff
@@ -127,10 +130,10 @@ module OpenAI
           override.returns(
             {
               id: String,
-              call_id: String,
               output: OpenAI::Responses::ResponseFunctionToolCallOutputItem::Output::Variants,
               status: OpenAI::Responses::ResponseFunctionToolCallOutputItem::Status::TaggedSymbol,
               type: Symbol,
+              call_id: String,
```
jbeckwith-oai, line 136: **[P2] Make the generated output hash contract reflect omitted `call_id`**

This PR makes response-output `call_id` optional, and `BaseModel` skips an absent optional field when populating `@data`; `to_hash` returns that data without a `:call_id` entry. However, this changed RBI still declares the returned hash has a required `call_id: String`, and the matching RBS and beta signatures make the same promise. Consequently `model.to_hash.fetch(:call_id)` type-checks but raises `KeyError` for the newly supported valid response. Generate an accurately optional hash key (and matching RBS/beta contracts) and add a focused regression for an output with no call ID.

jbeckwith-oai (changes requested):
The newly optional output call_id is correctly omitted at runtime, but the generated public Sorbet/RBS to_hash contracts still guarantee a required non-null call_id. Update the stable and beta generated hash shapes to accurately model omission and add a focused omitted-output regression.

rbi/openai/models/responses/response_function_tool_call_output_item.rbi
apcha-oai: hmm this looks like a longstanding issue with the generated RBI code will add a backlog issue for this

apcha-oai: https://linear.app/openai/issue/SDK-389

PR #543, opened 2026-08-27
feat(api): add compute_units to Responses and Chat Completions usage

apcha-oai:
Adds `compute_units` to Responses and Chat Completions usage.


HAYDEN-OAI (approved):
Reviewed the full 17-file diff at 109b0e24bb91664b1d0af8f545a5c29fd8a0b437 and the surrounding usage-model and coercion code. No actionable correctness issues found.

The new field is optional and explicitly nullable in CompletionUsage, ResponseUsage, and BetaResponseUsage, matching the schema without changing existing required token fields. The shared usage models cover normal Chat Completions, final streaming usage chunks, Responses, and compaction; the existing BaseModel handling preserves omitted, null, and integer values. RBI/RBS accessor and constructor declarations track the addition, and the two audit-log event values are consistent across the schema, runtime enums, and signatures.

Validation was source inspection; I did not run tests or use CI status as the basis for this review.

sig/openai/models/responses/response_usage.rbs
```diff
@@ -21,20 +22,24 @@ module OpenAI
...
         ) -> void
 
         def to_hash: -> {
           input_tokens: Integer,
           input_tokens_details: OpenAI::Responses::ResponseUsage::InputTokensDetails,
           output_tokens: Integer,
           output_tokens_details: OpenAI::Responses::ResponseUsage::OutputTokensDetails,
-          total_tokens: Integer
+          total_tokens: Integer,
+          compute_units: Integer?
```
jbeckwith-oai, line 42: **[P2] Mark omitted compute_units as an optional hash key**

`compute_units` is explicitly optional on the wire, and `BaseModel#to_hash` returns its raw `@data` without inserting absent fields. For an existing valid Responses payload that omits `compute_units`, `usage.to_hash` therefore has no `:compute_units` key, but this RBS record declares that key as required. Consequently a type-valid `usage.to_hash.fetch(:compute_units)` raises `KeyError`. Generate `?compute_units: Integer?` and the equivalent optional-key Sorbet contract; the same defect appears in the beta/Chat usage RBS/RBI aliases and `to_hash` signatures. Preserve omission rather than injecting a synthetic nil.

jbeckwith-oai (changes requested):
The new optional compute_units wire field currently violates the generated RBS/Sorbet hash contract when older valid payloads omit it. Please represent the key as optional across all affected stable/beta Responses and Chat usage signatures without changing runtime omission behavior.

HAYDEN-OAI (commented):
Correction to my earlier approval: my clean-review conclusion was mistaken. On rechecking the source, I agree with the existing [P2 on optional compute_units hash keys](https://github.com/openai/openai-ruby/pull/543#discussion_r3873874635).

The runtime correctly preserves an omitted compute_units field: decoding succeeds, the getter returns nil, and to_hash does not insert the key. However, the new RBS aliases/to_hash records and RBI to_hash shapes declare that nullable key as present across stable Responses, beta Responses, and Chat usage. Consequently, a caller relying on that presence promise can get KeyError from usage.to_hash.fetch(:compute_units) for a valid omitted-field payload. The constructors/getters and the RBI OrHash aliases are not the problem; OrHash uses AnyHash.

The hash signatures need an omission-aware contract while retaining runtime omission behavior. This is a source-based correction for head 109b0e24bb91664b1d0af8f545a5c29fd8a0b437; I did not execute a reproducer. The PR has already merged.

PR #545, opened 2026-08-27
test: make realtime cleanup deadline regression deterministic

jbeckwith-oai:
## Summary

- Make the interrupted Realtime call-cleanup redirect regression deterministic by replacing real sleeps and wall-clock assertions with a controlled timeout.
- Verify one cleanup deadline spans multiple redirects, preserves the configured per-request timeout, stops redirecting when that deadline expires, and propagates the original interrupted-call error.
- Preserve real timeout behavior for concurrently executing tests by limiting the timeout stub to the owning test thread.

The previous timing-sensitive assertion exceeded its 150 ms threshold on busy GitHub Actions runners, including observed durations of 262 ms on Ruby 4.0 and 161 ms on Ruby 3.3.

## Verification

- 500 consecutive focused regression-test iterations on Ruby 4.0.6.
- Complete Realtime call-creation suite on Ruby 4.0.6 and Ruby 3.3.12: 31 tests and 223 assertions on each runtime.
- Adjacent Realtime authentication/retry suite: 8 tests and 44 assertions.
- `mise exec ruby@4.0.6 -- bundle exec rake lint:rubocop`: Ruby formatting, RuboCop directives, and all 2,792 inspected files pass.
- Two consecutive clean, independent two-reviewer adversarial-review rounds.

### DIFF
--- a/test/openai/realtime/call_creation_test.rb
+++ b/test/openai/realtime/call_creation_test.rb
@@ -288,30 +288,42 @@ class OpenAI::Test::RealtimeCallCreationTest < Minitest::Test
       body: ""
     )
     requests = []
+    cleanup_deadlines = []
     http = Minitest::Mock.new(Object.new)
     http.expect(:execute, creation_response) do |request|
       requests << request
       true
     end
 
-    20.times do
+    20.times do |attempt|
       http.expect(:execute, redirect) do |request|
         requests << request
-        sleep(0.01)
+        raise Timeout::Error if attempt == 2
+
         true
       end
     end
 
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI (approved):
Reviewed `8894e6944d3215e55d736d6fb44a1ee511b9aa17`. The deterministic regression covers one shared cleanup timeout across redirects, preserves each request's configured timeout, stops at the injected timeout, and retains the original interrupted-call error. The stub delegates timeout behavior to the original implementation on other test threads. No actionable findings.

Validation: static review of the complete diff, production cleanup/redirect flow, and surrounding test support. I did not run tests or make live requests.

PR #548, opened 2026-08-27
fix: resume Responses streams from server-side cursors

jbeckwith-oai:
## Summary

- Forward `starting_after` when `client.responses.stream(response_id:, starting_after:)` retrieves an existing response, so the server resumes from the requested event sequence.
- Handle genuine incremental resumed streams that begin with text or tool deltas, finalized structured text, unknown events, or `response.completed`, while preserving normal stream validation and typed final responses.
- Add public-entrypoint regression coverage for cursor forwarding, partial snapshots, completion-only resumes, structured text/tool outputs, unknown-event filtering, and unchanged non-resumed behavior.

## Verification

- `TMPDIR=/private/tmp bundle exec rake test`
- `BUNDLE_FROZEN=true BUNDLE_GEMFILE=gemfiles/bedrock.gemfile bundle exec rake test:bedrock`
- `bundle exec rake lint:rubocop`
- `bundle exec rake validate:rbs`
- `python3 -m unittest discover -s scripts/castiron -p 'test_custom_code*.py'`
- Trusted-main Castiron custom-code budget check: 3,047 / 4,000 lines.
- Two consecutive clean adversarial-review rounds with two independent reviewers per round, including endpoint/deserialization security review.

@openai/sdks-team

### DIFF
--- a/lib/openai/helpers/streaming/response_stream.rb
+++ b/lib/openai/helpers/streaming/response_stream.rb
@@ -13,7 +13,8 @@ module OpenAI
           @last_response = raw_stream.last_response
           @iterator = iterator
           @state = ResponseStreamState.new(
-            text_format: text_format
+            text_format: text_format,
+            starting_after: starting_after
           )
         end
 
@@ -80,10 +81,12 @@ module OpenAI
       class ResponseStreamState
         attr_reader :completed_response
 
-        def initialize(text_format:)
+        def initialize(text_format:, starting_after: nil)
           @current_snapshot = nil
           @completed_response = nil
           @text_format = text_format
+          @resumed = !starting_after.nil?
+          @resumed_snapshots = {}
         end
 
         def handle_event(event)
@@ -98,11 +101,17 @@ module OpenAI
 
           case event
           when OpenAI::Models::Responses::ResponseTextDeltaEvent
-            output = @current_snapshot.output[event.output_index]
-            assert_type(output, :message)
+            snapshot = if @current_snapshot
+              output = @current_snapshot.output[event.output_index]
+              assert_type(output, :message)
 
-            content = output.content[event.content_index]
-            assert_type(content, :output_text)
+              content = output.content[event.content_index]
+              assert_type(content, :output_text)
+              content.text
+            else
+              key = [:text, event.output_index, event.content_index]
+              (@resumed_snapshots[key] ||= +"").concat(event.delta).dup
+            end
 
             events_to_yield <<
               OpenAI::Streaming::ResponseTextDeltaEvent.new(
@@ -112,17 +121,22 @@ module OpenAI
                 output_index: event.output_index,
                 sequence_number: event.sequence_number,
                 type: event.type,
-                snapshot: content.text
+                snapshot: snapshot
               )
 
           when OpenAI::Models::Responses::ResponseTextDoneEvent
-            output = @current_snapshot.output[event.output_index]
-            assert_type(output, :message)
+            text = if @current_snapshot
+              output = @current_snapshot.output[event.output_index]
+              assert_type(output, :message)
 
-            content = output.content[event.content_index]
-            assert_type(content, :output_text)
+              content = output.content[event.content_index]
+              assert_type(content, :output_text)
+              content.text
+            else
+              event.text
+            end
 
-            parsed = parse_structured_text(content.text)
+            parsed = parse_structured_text(text)
 
             events_to_yield <<
               OpenAI::Streaming::ResponseTextDoneEvent.new(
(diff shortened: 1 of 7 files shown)


lib/openai/helpers/streaming/response_stream.rb
```diff
@@ -98,11 +101,17 @@ def handle_event(event)
...
+              assert_type(output, :message)
 
-            content = output.content[event.content_index]
-            assert_type(content, :output_text)
+              content = output.content[event.content_index]
+              assert_type(content, :output_text)
+              content.text
+            else
+              key = [:text, event.output_index, event.content_index]
+              (@resumed_snapshots[key] ||= +"").concat(event.delta).dup
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8df46bd90995b5608c331b1f48d6464a0f699a59. For a genuine server-directed resume without response.created, text and function-argument deltas now stay as raw model delta events instead of building synthetic accumulated snapshots, so there is no resumed prefix buffer or per-delta prefix copy. ResponseStream#text now accepts the shared Responses text-delta model so incremental resumed text remains consumable; response.output_text.done still parses its authoritative full text, and response.completed remains the authoritative typed final response. Regression coverage now asserts partial resumed text/function deltas have no snapshot, verifies resumed text iteration, and preserves final output/typed completion behavior. Verified with the focused streaming suite (32 tests, 125 assertions), unknown-event suite (13 tests, 96 assertions), full Ruby suite (1502 tests, 13579 assertions), Bedrock (38 tests, 377 assertions), repository-wide RuboCop/formatting, Sorbet, RBS validation, trusted Castiron budget, and two consecutive clean fresh-context adversarial-review rounds.

jbeckwith-oai (commented):
Re-reviewed the exact current head. `starting_after` is forwarded to the retrieval stream, resumed streams no longer require a synthetic `response.created`, raw deltas avoid quadratic prefix materialization, and authoritative done/completed events still drive parsed/final results. No blocking findings. (GitHub does not permit this account to formally approve its own PR.)

lib/openai/helpers/streaming/response_stream.rb
```diff
@@ -98,31 +100,42 @@ def handle_event(event)
...
+                  output_index: event.output_index,
+                  sequence_number: event.sequence_number,
+                  type: event.type,
+                  snapshot: content.text
+                )
+            else
+              # A server-directed resumed stream may begin after response.created.
+              # Without the omitted prefix, a snapshot would be incomplete and
+              # materializing every partial prefix would make streaming quadratic.
+              events_to_yield << event
```
HAYDEN-OAI: [P2] Preserve helper event matching for resumed deltas

Once `starting_after` is forwarded, normal server resumes omit `response.created` and enter this branch. The yielded object is `OpenAI::Models::Responses::ResponseTextDeltaEvent`, which is not an instance of its `OpenAI::Streaming::ResponseTextDeltaEvent` subclass. Existing documented helper handlers therefore silently discard every resumed text delta; the repository's own `examples/responses/streaming_previous_response.rb:146` has exactly that `when` match. The function-argument branch has the same class-identity regression. Broadening the internal `.text` helper does not fix external `.each` consumers. Preserve the helper event class for partial resumes and represent unavailable snapshots explicitly, without rebuilding or copying the omitted prefix.

HAYDEN-OAI (commented):
Reviewed the full diff at `8b8f54b1f1b2240faca2a20bbc16c975255a8fee`, including cursor forwarding, partial-stream handling, structured completion, unknown events, and stream cleanup. The previous quadratic-copy concern is addressed, but one helper-event compatibility regression remains inline.

lib/openai/helpers/streaming/response_stream.rb
jbeckwith-oai: Fixed in `439c5f654bf439deeead263173a6b5034c722da9`.

Partial server-resumed text and function-argument deltas are now emitted as the public `OpenAI::Streaming::*DeltaEvent` helper subclasses, with `snapshot: nil` explicitly representing the unavailable omitted prefix. The rewrap uses the raw event payload plus that nil snapshot, so fields such as `logprobs`, `response_id`, and future additive fields remain available, while no prefix buffer is rebuilt or copied. I also added public-entrypoint regressions for helper-class matching, nil snapshots, preserved payload fields, and string-keyed cursor forwarding.

Verification: focused Responses streaming tests (33 runs, 122 assertions), full Ruby suite (1507 runs, 13639 assertions), Bedrock (38 runs, 377 assertions), RuboCop, Sorbet, RBS, custom-code budget, and the mandatory adversarial review with two consecutive clean rounds. This resolves the P1 quadratic-copying concern and the helper-subclass compatibility issue.

HAYDEN-OAI (approved):
Reviewed the full current diff at `439c5f654bf439deeead263173a6b5034c722da9`. The prior helper-event compatibility issue is fixed for both text and function-argument deltas: partial resumes retain the public helper subclasses and original payload fields, with `snapshot: nil` representing the unavailable prefix without rebuilding it. Cursor forwarding now also uses normalized parameters while preserving request options. Completion/structured-output handling, unknown-event filtering, normal-stream validation, and stream cleanup remain consistent. No substantive issues found.

PR #556, opened 2026-08-29
fix: handle custom chat completion tool calls

jbeckwith-oai:
## Summary
- skip valid custom tool calls in the non-streaming Chat Completions structured-output unwrap helper
- preserve existing function-call parsing and malformed-function error behavior
- add a public-path regression covering custom tool-call hydration and payload preservation

## Validation
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_completion_parser_test.rb
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/resources/chat/completions_test.rb
- mise exec ruby@4.0.6 -- bundle exec rake lint:rubocop
- ruby -c for both changed files
- git diff --check
- two consecutive clean adversarial-review rounds
- security boundary and bypass reviews found no issues

## Full suite
- mise exec ruby@4.0.6 -- bundle exec rake test was attempted: 1518 runs, 13687 assertions, 2 unrelated failures in FastFormatTest temp-path normalization and GemPackagingTest installed-gem x509 smoke coverage.

### DIFF
--- a/lib/openai/helpers/structured_output/chat_completion_parser.rb
+++ b/lib/openai/helpers/structured_output/chat_completion_parser.rb
@@ -28,6 +28,8 @@ module OpenAI
 
             raw[:choices]&.each do |choice|
               choice.dig(:message, :tool_calls)&.each do |tool_call|
+                next if tool_call[:type] == "custom"
+
                 func = tool_call.fetch(:function)
                 next if (model = tool_models[func.fetch(:name)]).nil?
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed 162441f393251ea6288cbaa67ee9c5242018927c. No substantive findings. The custom discriminator is checked before function-only parsing, preserving custom payloads for the existing union hydration path while leaving function argument parsing and errors unchanged.

PR #558, opened 2026-08-29
fix: respect StringIO cursor in multipart uploads

jbeckwith-oai:
## Summary
- encode multipart StringIO values from their current byte cursor
- preserve the cursor so replayable multipart retries keep the same payload
- add regression coverage proving a consumed fake-sensitive prefix is not uploaded

## Tests
- TEST=test/openai/internal/util_test.rb bundle exec rake test
- TEST=test/openai/http_client_test.rb bundle exec rake test
- bundle exec rake lint:rubocop
- git diff --check
- ruby -c lib/openai/internal/util.rb
- ruby -c test/openai/internal/util_test.rb

## Security review
- verified direct, FilePart-wrapped, EOF, and retry/replay behavior
- completed two consecutive clean adversarial-review rounds with two fresh independent reviewers per round

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -510,7 +510,7 @@ module OpenAI
             IO.copy_stream(val, y)
           in StringIO
             y << format(content_line, content_type || "application/octet-stream")
-            y << val.string
+            y << (val.string.byteslice(val.pos..) || "")
           in -> { primitive?(_1) }
             y << format(content_line, content_type || "text/plain")
             y << val.to_s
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed afb5e763ba05904677e3a7367918e3a0dce312e4. No substantive findings. The byte-based slice respects the current StringIO cursor, emits empty content at or beyond EOF, and applies consistently to direct and FilePart-wrapped multipart inputs. It leaves the source cursor unchanged, preserving the existing retry/redirect path that builds a fresh multipart stream for each attempt.

Static code/contract review only; tests, CI, lint, formatting, coverage, and mergeability were not run or assessed.

PR #559, opened 2026-08-29
fix: preserve formatter path boundaries

jbeckwith-oai:
## Summary
- preserve discovered formatter paths through NUL-delimited batching
- pass an option terminator before formatter paths
- add regression coverage for a discovered filename containing a newline

## Tests
- mise exec ruby@4.0.6 -- env BUNDLE_PATH=/tmp/openai-ruby-bundle bundle exec rake test TEST=test/scripts/formatting_policy_test.rb
- mise exec ruby@4.0.6 -- env BUNDLE_PATH=/tmp/openai-ruby-bundle bundle exec rake lint:rubyfmt FORMAT_FILE=/tmp/openai-ruby-rakefile-paths
- mise exec ruby@4.0.6 -- env BUNDLE_PATH=/tmp/openai-ruby-bundle bundle exec rubocop Rakefile test/scripts/formatting_policy_test.rb --force-exclusion
- git diff --check

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -87,9 +87,14 @@ run_rubyfmt = lambda do |mode|
   abort(violations.join("\n")) unless violations.empty?
   unless paths.empty?
     Tempfile.create("rubyfmt-paths") do |file|
-      file.write(paths.join("\n") + "\n")
+      file.binmode
+      paths.each do |path|
+        file.write(path)
+        file.write("\0")
+      end
+
       file.flush
-      sh("./scripts/rubyfmt", mode, "@#{file.path}")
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed bcd275d103a6a2855dcc1cd8f54ff475c2e37efc. No substantive findings. The binary NUL-delimited path file and argv-form xargs invocation preserve discovered path boundaries across batches, and the launcher forwards the option terminator. Empty input skips invocation, while nonzero formatter/batch failures still fail the task. Canonical absolute paths also avoid rubyfmt's leading-`@` response-file interpretation.

Static implementation review only.

PR #562, opened 2026-08-31
feat: add Responses WebSocket mode

jbeckwith-oai:
## Summary

- add a Ruby-native block-scoped `client.responses.connect` WebSocket surface
- extract product-neutral WebSocket transport, lifecycle, protocol, retry, and error helpers shared by Responses and Realtime
- keep Realtime compatibility through thin facades while making Responses permissive: faithfully forward client fields, best-effort decode known events, and preserve unknown data
- generate schema-shaped `response.create` keyword typing and corrected WebSocket event RBI nilability through Castiron
- add RBI/RBS declarations, README guidance, and focused regression coverage

## Verification

- `ruby -Itest test/openai/responses_websocket/connection_test.rb` (47 runs, 202 assertions)
- `SRB_SKIP_GEM_RBIS=1 srb typecheck --dir rbi`
- `rbs parse sig`
- Ruby syntax checks and `git diff --check`
- two consecutive clean fresh-context adversarial-review rounds after the final code change
- all GitHub checks green, including Ruby 3.3/3.4/4.0, bedrock, RBI/RBS, RuboCop, baseline consistency, and CodeQL

## Notes

- `response.create` typing is generated in the Responses resource RBI/RBS; the handwritten helper owns only runtime forwarding and its constructor declaration
- ambiguous writes remain terminal because delivery is unknowable; read transport failures are payload-safe and terminal; ordinary model-shape drift, error events, and payload-safe malformed JSON errors do not poison the connection
- the PR preserves forward compatibility through permissive event handling and keyword rest typing, and does not change the checked-in API reference


HAYDEN-OAI (commented):
Reviewed d1bedefca0baf7383713275615eaf3c4fe1db36a. Found two actionable correctness issues in nested-model serialization and application-exception propagation; details inline. Static source review only.

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,411 @@
...
+
+        @named_stream_ids[stream_id] = true
+      end
+
+      private def normalize_event_keys(value, active = {}.compare_by_identity)
+        case value
+        when OpenAI::Internal::Type::BaseModel
+          guard_cycle!(value, active)
+          begin
+            normalize_event_keys(value.to_h, active)
```
HAYDEN-OAI: [P2] Preserve generated-model serialization before normalizing keys

Using `to_h` here discards the nested model's serializer metadata before the subsequent wire-shape coercion. For example, `connection.response.create(..., tools: [OpenAI::Responses::Tool::ImageGeneration.new])` turns the tool into an empty hash, which is then matched and serialized as a `computer` tool instead of `image_generation`. Valid typed computer-call history is also rejected because click coordinates are stored as Ruby `y_`, while coercion expects wire key `y`. These are supported generated-model inputs on the normal `response.create` path. Preserve the model converter's default discriminators and `api_name` mappings, with cycle/duplicate-key preflight kept separate from serialization.

lib/openai/helpers/responses_websocket/transports/async_websocket.rb
```diff
@@ -0,0 +1,33 @@
...
+      #
+      # @api private
+      class AsyncWebSocket
+        def initialize
+          @transport = OpenAI::Realtime::Transports::AsyncWebSocket.new
+        end
+
+        def open(url:, headers:, timeout:, **options, &block)
+          @transport.open(url: url, headers: headers, timeout: timeout, **options, &block)
+        rescue OpenAI::Errors::RealtimeConnectionError => e
```
HAYDEN-OAI: [P2] Preserve Realtime exceptions raised by the caller's block

This rescue also catches a `RealtimeConnectionError` raised by application code inside the yielded Responses block, such as a failed nested `client.realtime.connect` call. The underlying transport deliberately preserves that application exception, but this wrapper replaces it with `ResponsesConnectionError`, so a surrounding Realtime-specific rescue no longer catches the failing inner operation even though the Responses connection opened successfully. Track exceptions originating in the yielded block and re-raise them unchanged; translate only errors from this transport's own work.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 3e2a2a18. Client-event validation now performs a read-only cycle and duplicate-key preflight, then serializes through Castiron before key normalization. Raw top-level hashes use target-neutral Unknown.dump so nested generated models retain default discriminators and api_name mappings without synthesizing a missing top-level discriminator; typed top-level events keep ResponsesClientEvent dumping. Added regressions for nested serializer metadata, raw missing type, and emitted API-name collisions.

lib/openai/helpers/responses_websocket/transports/async_websocket.rb
jbeckwith-oai: Fixed in 3e2a2a18. The Responses adapter now marks only RealtimeConnectionError raised from its yielded callback and re-raises that same exception unchanged. Delegate-originated Realtime transport failures still translate to payload-safe ResponsesConnectionError, including the missing async-websocket dependency guidance. Added a regression asserting callback error identity is preserved.

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
...
+          return OpenAI::Responses::UnknownServerEvent.new(data: parsed)
+        end
+
+        state = OpenAI::Internal::Type::Converter.new_coerce_state
+        event = OpenAI::Internal::Type::Converter.coerce(
+          OpenAI::Responses::ResponsesServerEvent,
+          parsed,
+          state: state
+        )
+        raise OpenAI::Errors::ResponsesProtocolError.new if coercion_error(state)
```
HAYDEN-OAI: [P1] Exclude SDK-computed attributes from wire validation

A normal `response.output_item.added` frame containing a `function_call` item has no `parsed` key—the SDK computes that response-only attribute locally—but `ResponseFunctionToolCall` declares it required. `BaseModel.coerce` therefore increments `state[:exactness][:no]`, and this check raises `ResponsesProtocolError`; `read_one` then poisons the connection before the application receives the tool call. Ordinary function-calling WebSocket turns consequently fail. The corresponding check at line 200 also rejects function-call items replayed as input after dumping removes `parsed`. Validate wire-required fields separately from SDK-computed attributes so both normal tool events and replay inputs remain usable.

HAYDEN-OAI (commented):
Reviewed 3e2a2a18d689ba1b07338cb103468684ead66835. The previously reported nested-model serialization and callback-error issues are fixed. One high-priority correctness issue remains: valid function-call wire items are rejected because validation requires the SDK-computed parsed attribute. Details inline. Static source review only.

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
+# frozen_string_literal: true
+
+module OpenAI
+  module Responses
+    # A live, typed Responses WebSocket connection.
+    class Connection
+      include Enumerable
+
+      STREAM_ID = /\A[A-Za-z0-9_.-]{1,256}\z/
```
apcha-oai: unnecessary validation; avoid ideally

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
...
+      MAX_NAMED_STREAM_IDS = 32
```
apcha-oai: recommend avoid arbitrary caps like this

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
...
+
+        reject_beta_input_items!(payload[:input]) if payload.key?(:input)
+      end
+
+      private def reject_beta_input_items!(value)
+        items = value.is_a?(Array) ? value : [value]
+        items.each do |item|
+          next unless item.is_a?(Hash)
+
+          beta_types = %i[agent_message multi_agent_call multi_agent_call_output]
```
apcha-oai: why reject this? we should just faithfully pass through imo

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
...
+        end
+
+        return if keys.map(&:to_s).uniq.length == keys.length
+
+        raise OpenAI::Errors::ResponsesClientEventError.new
+      end
+
+      private def reject_beta_event_fields!(payload)
+        return unless payload.is_a?(Hash)
+        raise OpenAI::Errors::ResponsesClientEventError.new if payload.key?(:multi_agent)
```
apcha-oai: similar here; we should avoid client validation like this imo

rbi/openai/helpers/responses_websocket/connection.rbi
```diff
@@ -0,0 +1,66 @@
+# typed: strong
+
+module OpenAI
+  module Models
+    module Responses
+      class Connection
+        include Enumerable
+
+        ServerEvent = T.type_alias do
+          T.any(OpenAI::Responses::ResponsesServerEvent::Variants, OpenAI::Responses::UnknownServerEvent)
```
apcha-oai: **Blocking:** Before exposing `ResponsesServerEvent::Variants` as the public WebSocket event type, please audit the generated event RBIs for soundness. For example, optional `stream_id` has a `T.nilable(String)` reader but non-null constructor, writer, and `to_hash` types. That inconsistency predates this PR, but `receive` and `each` now make these generated types part of the normal consumer surface. We should fix the generated types—or otherwise narrow the typing guarantee—before merging.

Likely an inheritance bug in the generator

rbi/openai/helpers/responses_websocket/connection_resources.rbi
```diff
@@ -0,0 +1,20 @@
...
+  module Models
+    module Responses
+      module ConnectionResources
+        class Response
+          # @api private
+          sig { params(connection: OpenAI::Responses::Connection).returns(T.attached_class) }
+          def self.new(connection)
+          end
+
+          sig { params(params: T.untyped).void }
```
apcha-oai: avoid T.untyped

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
...
+        raise
+      rescue StandardError
+        raise OpenAI::Errors::ResponsesClientEventError.new, cause: nil
+      end
+
+      # Close the connection.
+      def close(code: 1000, reason: "")
+        assert_owner!
+        return if closed?
+        return abort if poisoned?
```
apcha-oai, line 80: This is quite odd. We shouldn't poison the websocket if parsing fails once?

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,433 @@
...
+
+      private def parse_event(data)
+        parsed = JSON.parse(data, symbolize_names: true)
+        type = event_type(parsed)
+        validate_inbound_stream_id!(parsed[:stream_id]) if parsed.key?(:stream_id)
+        unless @server_event_names.key?(type.to_s)
+          return OpenAI::Responses::UnknownServerEvent.new(data: parsed)
+        end
+
+        state = OpenAI::Internal::Type::Converter.new_coerce_state
```
apcha-oai, line 116: [P0] Preserve normal parsing semantics

This treats generated-model coercion scoring as strict wire validation, unlike both normal HTTP responses and Responses streaming. For a concrete non-function-call example, the SDK’s own streaming fixtures accept `response.content_part.added` with `part: {type: "output_text", text: ""}`, while `ResponseOutputText` declares `annotations` required. Normal streaming yields a best-effort typed event; this path records `exactness[:no]`, raises `ResponsesProtocolError`, and permanently poisons the connection. The accepted `response.created` fixtures similarly omit several fields that `Response` marks required.

WebSocket parsing should have roughly the same permissive semantics as the existing HTTP and streaming paths: decode known fields, retain unknown/raw data, and reserve fatal protocol errors for invalid framing or JSON—not generated-model shape drift. Permanently killing a persistent connection because a response model is incomplete or newer than the SDK is not safe.

apcha-oai (changes requested):
**Semantic notes**
Blocking: this client is enforcing too much policy instead of acting as a permissive, faithful transport. It validates IDs, imposes arbitrary caps, rejects beta/newer fields, treats model coercion as wire validation, and permanently poisons the socket on ordinary shape mismatches.

We should match the SDK’s existing semantics here rather than inventing a stricter WebSocket-specific layer: faithfully pass requests through, parse responses flexibly, preserve unknown data, and avoid permanently poisoning the connection for model-shape mismatches. This should reuse the same established behavior as normal HTTP and streaming paths.

**Design notes**
I don’t think we should introduce another largely independent WebSocket stack. Responses is unlikely to be our last WebSocket API, and this causes some scaling issues for both consumers and maintainers.

At a high level, we should reuse and generalize the existing WebSocket machinery rather than fork it into another product-specific implementation.


Detailed design concerns:
- **blocking**: `OpenAI::Responses::Transports::AsyncWebSocket`\*\* delegates to `OpenAI::Realtime::Transports::AsyncWebSocket` and then translates `RealtimeConnectionError` into `ResponsesConnectionError`. I consider this blocking: generic WebSocket transport infrastructure should not be owned by one product API and adapted by every other product. The underlying transport and socket wrapper should move into a shared WebSocket namespace before we add another public WebSocket surface.
- `OpenAI::Realtime::ConnectionManager` and `OpenAI::Responses::ConnectionManager` duplicate transport validation, opening, handshake completion, connection construction, yielding, closing, aborting, and cleanup behavior. That lifecycle machinery should be shared.
- `OpenAI::Helpers::Realtime::ClientExtension` and `OpenAI::Helpers::ResponsesWebSocket::ClientExtension` duplicate WebSocket base-URL parsing, scheme conversion, request-option validation, header filtering, deadline handling, and workload-identity refresh behavior. That request/handshake construction should be shared.
- `OpenAI::Realtime::Connection` and `OpenAI::Responses::Connection` both implement event iteration, reading, writing, closing, aborting, discriminator lookup, coercion validation, and unknown-event fallback. The common connection and protocol machinery should be shared, even if the public product connection classes remain thin separate facades.
- `OpenAI::Errors::RealtimeProtocolError`, `OpenAI::Errors::ResponsesProtocolError`, `OpenAI::Errors::ResponsesClientEventError`, and `OpenAI::Errors::ResponsesSendError`: consider changing t part of a shared WebSocket error hierarchy rather than establishing another product-specific family.
- Consumers shouldn’t need to rescue both `OpenAI::Errors::RealtimeConnectionError` and `OpenAI::Errors::ResponsesConnectionError`, plus additional equivalent errors for every future WebSocket API. These  seem like protocol errors and not specific to the APIs

I feel less strongly about whether Realtime and Responses expose the exact same public connection class. Python and TypeScript retain product-specific connection surfaces, which may make sense given their different resource APIs. But those classes should ideally be thin product facades over one shared WebSocket core—not separate implementations or adapters that translate one product’s transport errors into another’s.

Realtime is new enough that we could either correct its public error surface in a breaking revision or preserve the existing names through compatibility inheritance. This is worth doing imo as we have several WS APIs already (e.g: responses beta)


### Meta notes
Stepping back, most of the concerns above fall out of two guiding principles that I think Codex missed:

- SDKs should be thin, permissive clients. They should generally pass inputs through faithfully, return what the API sends, and avoid imposing additional validation or policy beyond what is minimally necessary
- Generic protocol infrastructure should not be owned by a product API. For something like WebSockets, we should build a shared abstraction that specific API surfaces can use, rather than reusing one product’s implementation from another or rebuilding the same stack for every API.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: removed the Beta-field rejection. Client fields are faithfully serialized and forwarded instead of being rejected by transport-specific policy.

rbi/openai/helpers/responses_websocket/connection.rbi
jbeckwith-oai: Fixed in 8135e8fd: narrowed the public Responses WebSocket event typing to T.anything/top until generated event RBI soundness is repaired, and loosened UnknownServerEvent stream_id to reflect arbitrary JSON.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: known Responses events now follow HTTP and SSE-style best-effort coercion rather than exact wire validation. Added regressions for partial response.content_part.added and ordinary function_call frames, and unknown fields/events remain observable.

rbi/openai/helpers/responses_websocket/connection_resources.rbi
jbeckwith-oai: Fixed in 8135e8fd: replaced T.untyped and RBS untyped in the new helper surface with T.anything and top.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: removed the SDK-only stream ID validation. Stream IDs are now forwarded and unknown events preserve arbitrary JSON values, with regression coverage for a numeric stream_id.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: removed strict client model validation. Outbound events are forwarded with only narrow preflight checks needed for deterministic serialization safety, such as cycles and duplicate serialized keys.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: removed the arbitrary stream ID caps. The Responses WebSocket path no longer imposes SDK-only length or count limits.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: model-shape drift and ordinary error events no longer poison the connection. Malformed JSON remains a payload-safe protocol error while the connection can continue; ambiguous writes still poison because delivery is unknowable.

lib/openai/helpers/responses_websocket/connection.rb
jbeckwith-oai: Fixed in 8135e8fd: Responses now uses best-effort generated-model coercion rather than exactness rejection, so normal function_call output items without SDK-computed parsed decode successfully and replay inputs send without parsed. Added focused regressions in test/openai/responses_websocket/connection_test.rb.

lib/openai/helpers/responses_websocket/connection_resources.rb
```diff
@@ -0,0 +1,21 @@
...
+        # @api private
+        def initialize(connection)
+          @connection = connection
+        end
+
+        # Send a response.create event over this connection.
+        #
+        # @return [nil]
+        def create(**params)
+          @connection.send_event(type: :"response.create", **params)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 9697b901: Response#create now removes both symbol and string type keys before adding the fixed response.create discriminator. Added regressions for type: response.cancel and **{"type" => :"response.cancel"}; both emit response.create.

lib/openai/helpers/responses_websocket/connection.rb
```diff
@@ -0,0 +1,255 @@
...
+        else
+          value
+        end
+      end
+
+      private def connection_error(message)
+        OpenAI::Errors::ResponsesConnectionError.new(url: @url, message: message)
+      end
+
+      private :receive_raw, :send_raw
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 9697b901: removed the Responses private visibility override so inherited receive_raw and send_raw remain callable with an explicit receiver. Both shared runtime methods are marked @api private, and the Responses RBI/RBS now declare them publicly with matching types. Added a focused explicit-receiver regression.

rbi/openai/helpers/responses_websocket/connection.rbi
```diff
@@ -0,0 +1,72 @@
...
+
+module OpenAI
+  module Models
+    module Responses
+      class Connection
+        include Enumerable
+
+        ServerEvent = T.type_alias { T.anything }
+
+        ClientEvent = T.type_alias { T.anything }
```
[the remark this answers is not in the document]
jbeckwith-oai: No code change here: the T.anything/top aliases are intentional and directly address the earlier blocking human review on this same surface. The generated ResponsesServerEvent union remains unsound today: stream_id readers are nilable while generated writers, constructors, and to_hash signatures require non-null String. Re-exposing that union from receive/each would restore the typing bug Alex asked us to avoid. We will keep the narrowed public guarantee until the generator/inherited-field RBI issue is fixed in its own scope.

lib/openai/helpers/websocket/client_request.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module OpenAI
+  module WebSocket
+    # Shared handshake request helpers used by product WebSocket facades.
+    #
+    # @api private
+    module ClientRequest
```
markstuart-oai, line 8: This shared boundary still leaves the security-sensitive handshake orchestration duplicated. `with_websocket_connection_retry` is used by both products, but the other four helpers in this module are Responses-only; Realtime still separately performs request-option handling, base-URL parsing, URL conversion, header filtering, deadline calculation, workload-identity token replacement, request preparation, and remaining-timeout calculation. That means auth or redaction changes must be kept synchronized across two builders. Move the common build/prepare/token/URL/header/timeout sequence behind this boundary and parameterize only the product guards, path/query, and error construction, so both product extensions become thin policy adapters.

markstuart-oai (commented):
Reviewed `33bd3a3d9e6f3a1c8f8adc3eb464016a80296062`. The transport, connection lifecycle, protocol, and error extraction address most of the earlier structural duplication, and the prior correctness issues are resolved. One maintainability concern remains in handshake construction; details inline. Exact-head CI is green; local runtime tests were unavailable because this leased box has neither Ruby nor mise.

lib/openai/helpers/websocket/client_request.rb
jbeckwith-oai: Fixed in 06ba8541. OpenAI::WebSocket::ClientRequest now owns the complete shared handshake sequence: request-option validation, safe base-URL parsing/rebasing and ws conversion, request building/preparation, workload-identity deadline and token replacement, proxy-authorization filtering, remaining-timeout calculation, and timeout handoff. Realtime and Responses now only supply product guards, route/query, their existing user-facing messages, and product error construction. I preserved Realtime timeout cause chaining and Responses payload-safe suppression, with a focused Responses timeout regression. Focused Responses/Realtime suites and full lint/type/RBS checks pass.

apcha-oai (changes requested):
The runtime changes address much of my earlier feedback: permissive parsing and forwarding, plus shared WebSocket infrastructure, are the right direction. However, **the public typing remains blocking. We cannot ship this API with `T.anything` replacing its request and event types.** Providing consumers with useful, accurate types is a core purpose of this SDK, and Sorbet is a first-class typechecker we support.

**P0 — Restore the public request/event typing contract.**

`ServerEvent` and `ClientEvent` are currently `T.anything`, with corresponding RBS `top` aliases. This removes the typing contract from the new API. Decoding events into generated classes at runtime does not substitute for static typing.

This has concrete consequences: Sorbet accepts `connection.send_event(123)` and `connection.send_event(Object.new)`, while a README-style event loop fails at `event.type` because the received value is `T.anything`.

We need a properly typed surface before shipping. Either generate it, which I would prefer, or handwrite complete declarations referencing the generated types and maintain them accordingly. **Erasing the public types is not an acceptable workaround for incorrect generated declarations.**

**P1 — Accurately type the accepted request parameters and support normal Sorbet usage.**

`response.create` needs to describe the parameters it accepts—`model`, `input`, `reasoning`, `tools`, `stream_id`, etc.—with their schema-defined types and optionality. The distinction is whether those parameters are accurately typed, not whether the language represents them as keywords or an object.

The existing [Ruby Responses#create RBI](https://github.com/openai/openai-ruby/blob/180f75fb4f1843159668183b7cf349999e232bed/rbi/openai/resources/responses.rbi#L28) demonstrates the intended pattern: individually typed keyword parameters referencing generated types. [Python](https://github.com/openai/openai-python/blob/83267847a0219ea8b584c9d60f92c7a4dffd392a/src/openai/resources/responses/responses.py#L134) similarly exposes typed keywords; [Node](https://github.com/openai/openai-node/blob/f83de8fe4c4a309cabae861ab552a3ce6d4a2c65/src/resources/responses/responses.ts#L10575) exposes an object with explicitly typed properties. Please apply that pattern using the WebSocket schema.

The current `create(**params)` RBI instead declares each keyword value as `T::Hash[Symbol, T.anything]`. Ordinary string-valued `model` and `input` fail Sorbet, while invalid hash-valued versions pass.

Replacing that declaration with `T.anything` would remove the immediate error without providing the required parameter contract. Please add consumer typechecks demonstrating that normal calls succeed, incorrect known parameter types fail, and generated nested inputs work correctly.

**Clarification on `stream_id` — resolved, not blocking.**

I am withdrawing the earlier P2 concern. This is the known Sorbet hash-shape limitation around omitted keys, not a Ruby runtime or WebSocket-specific model regression. The direct `stream_id` accessor is correctly typed as `T.nilable(String)`, and the RBS declarations already mark the hash key optional. Sorbet cannot express that same optional-key contract precisely in an RBI heterogeneous hash shape; in the supported version, indexing the shape with `to_hash[:stream_id]` is inferred as `T.untyped`. The earlier passing hash-indexing example therefore did not demonstrate that Sorbet had established a non-null `String`.

I consider this concern resolved within the SDK’s existing, documented Sorbet limitation. It is not an additional prerequisite for this PR, and it does not justify replacing the generated event types with `T.anything`; the P0 and P1 requirements above remain.

The live tests provide useful evidence that the runtime works—including function-call events, generated-model replay, and sequential requests—but green runtime tests and CI do not establish a correctly typed public API.

jbeckwith-oai:
Addressed the remaining review feedback in 80ee5954:

- P0: `Connection::ServerEvent` and `ClientEvent` now expose the generated Responses WebSocket unions, with `UnknownServerEvent` / raw hash fallbacks instead of `T.anything` / `top`.
- P1: `response.create` now has schema-shaped known keyword types in RBI and RBS while retaining `**params` at runtime for forward-compatible fields. Added consumer Sorbet coverage proving a normal generated nested input typechecks and a wrong known `model` type fails.
- Breaking-change audit: restored the pre-existing Realtime `AsyncWebSocket::Socket` RBI/RBS callback contract and a real runtime compatibility wrapper, including its legacy constructor and yielded class identity.

The earlier optional `stream_id` concern remains non-blocking as clarified in the review; the separate generator fix is tracked in openai/openai#1425293 and openai/openai-ruby-internal#83.

Verification: Responses WebSocket runtime/errors/Sorbet tests, Realtime connection/transport/auth/sideband/abort/network-invariant tests, Sorbet examples, RBS validation (1,285 files), focused RuboCop, and two consecutive clean fresh-context adversarial-review rounds.

apcha-oai (changes requested):
This is much closer and addresses the immediate typing issues. My remaining concern is that **the pieces that evolve with the API are still handwritten**.

For example, suppose we add a `foo` parameter to `response.create`, backed by a generated `FooParam` type. Codegen will update the request model, but nothing ensures `foo:` is added to these handwritten RBI/RBS signatures with the correct type and optionality. The runtime may forward it through `**params`, while the typed interface falls behind.

Similarly, adding a tool variant can update the generated tool union while leaving the manually enumerated `tools:` signature stale. The existing consumer tests can continue passing because they only exercise older parameters and tools.

We could adopt a “keep the signatures in sync” pattern, but I consider that an antipattern here. It creates a second copy of the API contract that a person, Codex, or a separate synchronization script must continually reconcile. A check that detects drift would help catch mistakes, but we should avoid introducing that duplication in the first place.

Please generate those schema-dependent pieces through the existing SDK generation pipeline, so adding a parameter or tool updates the WebSocket declarations in the same run.

The transport and connection lifecycle can remain handwritten. **The parts that change when the API changes should be generated.**

sylvesterkaczmarek (commented):
The transport/lifecycle refactor is much cleaner now, but one API-evolution boundary still looks fragile on the current head: `connection_resources.rbi` hand-enumerates the `response.create` keyword set and tool union while the runtime accepts forward-compatible `**params`. A schema/codegen update can therefore make runtime accept a new field or tool while Sorbet/RBS remains stale with no automatic coupling. I would generate the schema-dependent WebSocket request/event declarations from the same API model pipeline and keep only the transport/lifecycle layer handwritten; otherwise this becomes a second API contract that has to be manually synchronized.

lib/openai/helpers/websocket/connection.rb
```diff
@@ -0,0 +1,96 @@
...
+      def receive_raw
+        message = @socket.read
+        message&.to_str
+      end
+
+      # Send an already encoded text message.
+      #
+      # @api private
+      def send_raw(data)
+        raise connection_error(closed_send_message) if closed?
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d80f0d4d. Responses now overrides the public @api private send_raw boundary, enforcing the owner thread and poisoned/closed terminal state before delegating to shared text encoding. The shared writer routes through a private write_text primitive, and Responses poisons on both encoded and direct raw write failures. Added regressions for foreign-thread raw sends, direct raw-write poisoning, and poisoned retries not consulting the socket again.

rbi/openai/helpers/responses_websocket/connection.rbi
```diff
@@ -0,0 +1,82 @@
...
+  module Models
+    module Responses
+      class Connection
+        include Enumerable
+
+        ServerEvent = T.type_alias do
+          T.any(
+            OpenAI::Responses::ResponsesServerEvent::Variants,
+            OpenAI::Responses::UnknownServerEvent
+          )
```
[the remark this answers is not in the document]
jbeckwith-oai: Resolved by the reviewed Castiron integration in d80f0d4d. We intentionally retain best-effort known-event decoding to match existing HTTP/SSE Responses behavior, as Alex requested, rather than making a persistent WebSocket fatal on generated-model shape drift. The generated event RBI bug that previously required T.anything/top is now fixed at the generator layer for inherited and nested optional fields, so the concrete union is restored without reintroducing that defect. A server payload omitting a schema-required field remains outside the static event contract; unknown/new fields and events stay forward-compatible.

apcha-oai (approved):
Thanks for following through on the fixes, Justin! Moving the schema-dependent WebSocket signatures into Castiron and removing the handwritten duplicate addresses my remaining typing and maintainability concerns. The generated event unions and Sorbet consumer checks look good, and the runtime tests and live WebSocket tool-call/replay test passed on this head. LGTM.

PR #573, opened 2026-09-02
fix: emit done events for terminal chat tool calls

jbeckwith-oai:
## Summary

Fix chat-completion helper streams so a function tool call first appearing in the terminating choice chunk emits its expected `tool_calls.function.arguments.done` event.

Applications that react to this done event can otherwise miss a completed tool call even though `get_final_completion` contains the correct call and arguments. This is deterministic synthetic protocol-fixture evidence, not a claim about production frequency or incidents. The SDK does not execute tools itself.

The trigger is a valid streamed chat-completion choice whose first function tool call arrives complete in the same chunk that has `finish_reason: "tool_calls"`. It also occurs after a role-only prefix. Callers affected are consumers of:

```ruby
require "openai"

  model: "gpt-4o-mini",
  messages: [{role: :user, content: "Use the tool"}],
  tools: [
    {
      type: :function,
      function: {
        name: "lookup_value",
        strict: true,
        parameters: {
          type: "object",
          properties: {value: {type: "integer"}},
          required: ["value"],
          additionalProperties: false
        }
      }
    }
  ]
)

stream.each do |event|
  next unless event.type == :"tool_calls.function.arguments.done"

  puts [event.index, event.name, event.arguments, event.parsed].inspect
end
```

## Deterministic evidence

Using a network-disabled WebMock SSE fixture with a complete strict function call carrying `{"value":1}`:

| Prefix before terminal tool chunk | Before | After | Final arguments |
| --- | ---: | ---: | --- |
| none | 0 done events | 1 done event | `{"value":1}` |
| role only | 0 done events | 1 done event | `{"value":1}` |
| earlier tool header | 1 done event | 1 done event | `{"value":1}` |

The new focused public-entrypoint regression also covers multiple tools in one terminal frame, strict parsed arguments, non-strict and unknown-input compatibility, full argument deltas, final completion arguments, and the existing split header/delta sequence without duplicates.

`ChoiceEventState#get_done_events` handled the terminal finish reason before tracking tool indices present in that same chunk. A newly introduced terminal-frame tool therefore was not current when the terminal done event was emitted.

The fix keeps the existing completion-state flow and deduplication set, but performs the terminal tool flush after the current chunk's tool indices are processed. No new event type, public field, parser, index model, or abstraction is introduced.

This is a backward-compatible bug fix in the handwritten chat streaming helper. It preserves event shapes, final completions, argument delta payloads, strict parsing semantics, and supported sequential-tool behavior. Production changes are limited to `lib/openai/helpers/streaming/chat_completion_stream.rb`; coverage is a new focused test.

Non-goals include interleaved-tool protocol redesign, changing nonterminal early-completion semantics, adding choice IDs or event fields, changing argument parsing or tool dispatch, handling initial-chunk `length`/`content_filter`, service-tier/fingerprint/logprobs changes, stream transports/lifecycle, generated code, signatures, dependencies, or unrelated cleanup.

## Verification

```sh
mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_terminal_tool_test.rb
mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_completion_stream_test.rb
mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/resources/chat/completions/streaming_test.rb
mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/resources/chat/completions/streaming_snapshot_test.rb
git diff --check
```

Results:

- Focused regression: 3 runs, 16 assertions, 0 failures.
- Chat helper suite: 10 runs, 47 assertions, 0 failures.
- Public chat streaming suite: 22 runs, 107 assertions, 0 failures.
- Streaming snapshot suite: 15 runs, 73 assertions, 0 failures.
- Full lint/type/format task: passed; RuboCop inspected 2,796 files with no offenses, Sorbet had no errors, and 1,250 RBS files validated.
- Diff whitespace check: passed.

All fixtures are synthetic and network-disabled; no live API calls were made.

## Public custom-code budget

The trusted public gate passed against fresh `origin/main` and the committed candidate: 3,063 custom lines against a 4,000-line limit, with 937 lines of headroom. No budget file, reporting code, generation metadata, or exclusion changed.

## Security assessment

The fix improves event completeness while reusing the existing strict-argument parsing path unchanged. It adds no deserialization surface, validation rule, logging, credential handling, transport behavior, dependency, or sensitive fixture. No suspected vulnerability was identified.

## Known limitations

No live API request or full repository/Bedrock test suite was run locally. The focused public-entrypoint tests and complete chat-streaming helper, public-streaming, and snapshot suites listed above were run. The fixture proves the chunk-ordering case deterministically; it does not assert how often a live service emits that shape.

### DIFF
--- a/lib/openai/helpers/streaming/chat_completion_stream.rb
+++ b/lib/openai/helpers/streaming/chat_completion_stream.rb
@@ -648,11 +648,6 @@ module OpenAI
 
           if choice_snapshot.finish_reason
             events.concat(content_done_events(choice_snapshot, response_format))
-
-            if @current_tool_call_index && !@done_tool_calls.include?(@current_tool_call_index)
-              event = tool_done_event(tool_calls_by_index, @current_tool_call_index)
-              events << event if event
-            end
           end
 
           Array(choice_chunk.delta.tool_calls).each do |tool_call|
@@ -666,6 +661,13 @@ module OpenAI
             end
 
             @current_tool_call_index = tool_call.index
+          end
+
+          if choice_snapshot.finish_reason &&
+              @current_tool_call_index &&
+              !@done_tool_calls.include?(@current_tool_call_index)
+            event = tool_done_event(tool_calls_by_index, @current_tool_call_index)
+            events << event if event
           end
 
           events
--- a/test/openai/helpers/chat_stream_terminal_tool_test.rb
+++ b/test/openai/helpers/chat_stream_terminal_tool_test.rb
@@ -1,0 +1,161 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
+class OpenAI::Test::ChatStreamTerminalToolTest < Minitest::Test
+  extend Minitest::Serial
+  include WebMock::API
+
+  def before_all
+    super
+    WebMock.enable!
+  end
+
+  def after_all
+    WebMock.disable!
+    super
+  end
+
+  def setup
+    super
+    @client = OpenAI::Client.new(base_url: "http://localhost", api_key: "test-key")
+  end
+
+  def teardown
+    WebMock.reset!
+    super
+  end
+
+  def test_terminal_tool_call_emits_done_without_a_tool_header
+    [[], [role_choice]].each do |prefix|
+      stream, events = stream_events(prefix + [terminal_choice(tool_call(0, "strict_tool", "{\"value\":1}"))])
+
+      assert_equal([0], done_events(events).map(&:index))
+      assert_equal(["{\"value\":1}"], argument_deltas(events).map(&:arguments_delta))
+      assert_equal(
+        "{\"value\":1}",
+        stream.get_final_completion.choices.first.message.tool_calls.first.function.arguments
+      )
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
Checked the ordering change with multiple tools. Index transitions still flush the previous completed call, then the finish-reason flush after the loop catches the final or newly introduced terminal call. `@done_tool_calls` keeps the split header/delta path from emitting twice. The added cases cover all three shapes. Looks right.

HAYDEN-OAI (approved):
Reviewed the full diff at `f96f8ea71a05bad3c9d5a7961fb0154982b39a37`; no substantive issues found.

Flushing after the current chunk's tool indices are processed reaches the final newly introduced terminal call while preserving existing completion order, per-index deduplication, accumulated arguments, and strict-argument parsing. I also checked the initial-terminal, role-prefix, multi-tool, and split-header paths in source.

PR #580, opened 2026-09-02
fix: preserve initial chat stream system fingerprint

jbeckwith-oai:
## Summary

Preserve a `system_fingerprint` supplied on the first Chat Completions streaming chunk in the helper's current snapshot and final completion.

The public streaming helper previously exposed the first fingerprint on the raw chunk but initialized the accumulated completion with `nil`. Users inspecting `event.snapshot.system_fingerprint` or `stream.get_final_completion.system_fingerprint` therefore lost valid server metadata whenever the fingerprint appeared only on the first chunk, which is a normal streamed-response shape.

## Public SDK behavior

```ruby
  model: "gpt-4o-mini",
)

stream.each do |event|
  next unless event.type == :chunk

  puts event.snapshot.system_fingerprint
end

puts completion.system_fingerprint
```

When the server sends `system_fingerprint: "fp_initial"` on the first chunk and `nil` afterward, both the delivered snapshot progression and final completion now retain `"fp_initial"`.

## Deterministic synthetic evidence

Before:

```text
raw:       ["fp_initial", nil, nil]
snapshots: [nil, nil, nil]
final:     nil
```

After:

```text
raw:       ["fp_initial", nil, nil]
snapshots: ["fp_initial", "fp_initial", "fp_initial"]
final:     "fp_initial"
```

The focused offline regression also verifies that a later non-`nil` fingerprint still replaces the initial value, a later `nil` does not erase a value, and an initial `nil` remains `nil` unless a later chunk supplies a fingerprint. It checks that raw chunk values, text (`"Hello world"`), usage, choice index, and event sequence remain unchanged.

## Root cause and fix

`convert_initial_chunk_into_snapshot` hardcoded `system_fingerprint: nil` while later chunk accumulation already assigned only non-`nil` fingerprints. The fix initializes the existing snapshot field from `data[:system_fingerprint]`, allowing the existing later-chunk rule to preserve or replace it correctly.

## Compatibility and boundaries

This is a backward-compatible bug fix. It does not change the public API, parsing or strictness, tools, event construction, snapshot lifetime/isolation, transport lifecycle, other metadata fields, dependencies, or generated models/resources/signatures. The implementation stays in the handwritten helper boundary and adds one focused synthetic WebMock test.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_fingerprint_test.rb` — 1 run, 32 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_completion_stream_test.rb` — 10 runs, 47 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_initial_finish_test.rb` — 3 runs, 48 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/resources/chat/completions/streaming_test.rb` — 22 runs, 107 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/resources/chat/completions/streaming_snapshot_test.rb` — 15 runs, 73 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec rake test` — 1,554 runs, 13,943 assertions, 0 failures, 0 errors, 1 existing skip.
- `mise exec ruby@4.0.6 -- bundle exec rake format` — passed.
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — passed; 2,800 files inspected with no offenses, plus format, RBS, directive, Sorbet, and type checks.
- Public committed-candidate custom-code budget — passed at 3,311/4,000 lines with 689 lines headroom.

## Limitations

No live API call was made; evidence uses deterministic synthetic SSE chunks through the public client streaming interface. No dedicated security review was required because the change does not touch a security-sensitive surface.

### DIFF
--- a/lib/openai/helpers/streaming/chat_completion_stream.rb
+++ b/lib/openai/helpers/streaming/chat_completion_stream.rb
@@ -534,7 +534,7 @@ module OpenAI
               model: data[:model],
               choices: choices,
               usage: data[:usage],
-              system_fingerprint: nil,
+              system_fingerprint: data[:system_fingerprint],
               service_tier: data[:service_tier]
             }
           )
--- a/test/openai/helpers/chat_stream_fingerprint_test.rb
+++ b/test/openai/helpers/chat_stream_fingerprint_test.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
+class OpenAI::Test::ChatStreamFingerprintTest < Minitest::Test
+  extend Minitest::Serial
+  include WebMock::API
+
+  def before_all
+    super
+    WebMock.enable!
+  end
+
+  def after_all
+    WebMock.disable!
+    super
+  end
+
+  def setup
+    super
+    @client = OpenAI::Client.new(base_url: "http://localhost", api_key: "test-key")
+  end
+
+  def teardown
+    WebMock.reset!
+    super
+  end
+
+  def test_system_fingerprint_accumulates_without_changing_stream_output
+    cases = [
+      ["fp_initial", nil, "fp_initial"],
+      [nil, "fp_later", "fp_later"],
+      ["fp_initial", "fp_later", "fp_later"],
+      [nil, nil, nil]
+    ]
+
+    cases.each do |initial_fingerprint, later_fingerprint, expected_fingerprint|
+      stream = stream_for(initial_fingerprint:, later_fingerprint:)
+      event_types = []
+      raw_fingerprints = []
+      snapshot_fingerprints = []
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed `e43d4c19`. No substantive issues found. The initial `system_fingerprint` is preserved in stream snapshots and the final completion; later non-nil fingerprints still replace it, while nil chunks do not erase it.

PR #582, opened 2026-09-02
fix: preserve string-tagged structured output models

jbeckwith-oai:
Summary

Fix Chat Completions structured-output handling so supported string discriminators behave like their symbol counterparts. A nested `response_format` tagged with `"json_schema"` now serializes the model's JSON Schema object and hydrates the parsed message model; a streamed tool tagged with `"function"` now retains its typed argument model for both done events and the final completion.

Bug and impact

Users can pass SDK parameters with either Ruby symbols or JSON-style strings. Before this fix, replacing only `:json_schema` with `"json_schema"` in a converter-bearing Chat Completions response format sent the model class name as the wire `schema` value instead of its object schema, and the returned message had no typed `parsed` model. Replacing only `:function` with `"function"` for a streamed converter-bearing tool left done/final arguments as raw hashes instead of the requested model.

The failure is deterministic and offline:

- Before: `type: :json_schema` sent an object schema and returned `DiscriminatorText`; `type: "json_schema"` sent `"DiscriminatorText"` as `schema` and returned `NilClass`.
- Before: `type: :function` produced `DiscriminatorTool` for streamed done/final arguments; `type: "function"` produced `Hash`.
- After: both discriminator spellings send the same object schema/parameters and return the same typed models.

Root cause and minimal fix

The structured-output parser matched only the symbol `:json_schema` in its two nested response-format converter patterns, so string-tagged forms bypassed schema conversion and model capture. The stream input-tool binder similarly copied captured models only when `tool[:type] == :function`.

The fix broadens those exact checks only:

- the two existing converter-bearing response-format patterns accept `:json_schema | "json_schema"`;
- stream model binding accepts `:function` or `"function"`.

This uses the existing Chat Completions helper boundary and existing model conversion path; it does not add a new coercion framework or normalize arbitrary hashes.

Compatibility and scope

Explicit `name` and `strict` values remain intact. Direct converter forms, symbol controls, ordinary raw schema hashes, unknown tools, and custom/unrelated tool tags retain their existing behavior. The change is confined to the handwritten Chat Completions structured-output helper and its focused tests; generated code, transports, streaming helpers, signatures, dependencies, and public APIs are unchanged.

Verification

- Focused regression: `bundle exec ruby -Itest test/openai/helpers/chat_completion_parser_test.rb` — 11 runs, 186 assertions, 0 failures, 0 errors.
- Offline proof: symbol and string response formats both sent object schemas and hydrated `DiscriminatorText`; symbol and string streamed tools both produced `DiscriminatorTool` for done/final arguments.
- `bundle exec rake format` — passed and idempotent.
- `bundle exec rake lint` — passed; Sorbet clean, RuboCop directives valid, 1,250 RBS files valid, 2,799 files inspected with no offenses.
- `bundle exec rake test` — 1,555 runs, 13,974 assertions, 0 failures, 0 errors, 1 skip.
- `git diff --check` — clean.
- Two consecutive fresh adversarial review rounds, each with independent correctness/security and architecture/simplicity reviewers — clean.
- Committed public custom-code gate — passed: 3,311 / 4,000 custom lines, 689 lines headroom.

Security and limitations

The serialization/deserialization boundary was reviewed explicitly. This recognizes only two exact existing enum strings and adds no transport, authentication, logging, secret, dependency, file, or generated-code behavior. Tests use fake offline payloads; no live API request was made. The full suite retains one existing skip and existing Ruby/async warnings.

### DIFF
--- a/lib/openai/helpers/structured_output/chat_completion_parser.rb
+++ b/lib/openai/helpers/structured_output/chat_completion_parser.rb
@@ -67,7 +67,10 @@ module OpenAI
               }
             )
           in {
-              response_format: {type: :json_schema, json_schema: OpenAI::StructuredOutput::JsonSchemaConverter => model}
+              response_format: {
+                  type: :json_schema | "json_schema",
+                  json_schema: OpenAI::StructuredOutput::JsonSchemaConverter => model
+                }
             }
             parsed.fetch(:response_format).update(
               json_schema: {
@@ -78,7 +81,7 @@ module OpenAI
             )
           in {
               response_format: {
-                  type: :json_schema,
+                  type: :json_schema | "json_schema",
                   json_schema: {schema: OpenAI::StructuredOutput::JsonSchemaConverter => model}
                 }
             }
@@ -124,7 +127,7 @@ module OpenAI
           return [] if tools.nil?
 
           tools.map do |tool|
-            next tool unless tool[:type] == :function
+            next tool unless [:function, "function"].include?(tool[:type])
 
             function_name = tool.dig(:function, :name)
             model = tool_models[function_name]
--- a/test/openai/helpers/chat_completion_parser_test.rb
+++ b/test/openai/helpers/chat_completion_parser_test.rb
@@ -18,17 +18,18 @@ class OpenAI::Test::ChatCompletionParserTest < Minitest::Test
   class RecordingTransport < OpenAI::HTTPClient
     attr_reader :request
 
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed `df6b1b92`. No substantive issues found. String-tagged response formats and function tools now follow the existing symbol paths, preserving explicit schema settings and typed streamed arguments while keeping internal model metadata out of the request body.

PR #584, opened 2026-09-02
fix: preserve recursive structured output refs

jbeckwith-oai:
## Summary

Fix recursive structured-output schemas when a recursive `OpenAI::BaseModel` field is nullable or documented. Previously, `to_json_schema` could construct a cyclic Ruby hash, so serializing the schema raised `JSON::NestingError` before a request could be sent.

Affected users are callers defining recursive structured-output models such as linked lists or mutually recursive trees with `nil?: true` and/or `doc:` metadata.

## Public Ruby example

```ruby
require "json"
require "openai"

class Node < OpenAI::BaseModel
  required :name, String
  required :next, -> { Node }, nil?: true, doc: "Optional next node"
end

schema = Node.to_json_schema
puts JSON.generate(schema)
```

## Deterministic before/after

Before:

```text
JSON::NestingError: nesting of 100 is too deep. Did you try to serialize objects with circular references?
```

After:

```json
{"$defs":{"":{"type":"object","properties":{"name":{"type":"string"},"next":{"anyOf":[{"$ref":"#/$defs/"},{"type":"null"}],"description":"Optional next node"}},"required":["name","next"],"additionalProperties":false}},"$ref":"#/$defs/"}
```

## Root cause and fix

The converter already caches schema definitions and tracks every provisional `$ref`. For nullable or metadata-decorated references, it intentionally marks a pointer as non-reusable so an acyclic definition can be inlined with the local decoration. During recursive construction, that same inlining step inserted the completed definition back into one of its own descendants, producing an object cycle.

The fix marks only cache entries re-entered while their first reference path is still unresolved. Those recursive definitions stay in `$defs`, and their decorated sites remain finite `$ref`-based schemas. Ordinary acyclic reuse continues through the existing materialization path.

## Compatibility and scope

- Preserves plain recursive schemas.
- Preserves nullable and description metadata at recursive field sites.
- Preserves resolvable local `$ref` targets for self-recursive and mutually recursive optional links.
- Does not introduce `allOf` for recursive metadata; `allOf` is unsupported by the Structured Outputs strict subset.
- Does not change `ArrayOf`, `BaseModel`, `UnionOf`, parsers, Sorbet behavior, generated code, signatures, dependencies, or public API shape.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby /Users/jbeckwith/Documents/Codex/2026-08-31/do-this/work/batch2/verify_recursive_schema.rb /Users/jbeckwith/.codex/worktrees/5d25/openai-ruby`
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/structured_output_recursive_refs_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/structured_output_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/structured_output_api_names_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/structured_output_nullable_literal_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/structured_output_constant_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec rake format`
- `mise exec ruby@4.0.6 -- bundle exec rake test` — 1,555 runs, 13,952 assertions, 0 failures, 0 errors, 1 skip
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — 2,800 files inspected, no offenses; Sorbet/RBS/directive checks passed
- Two consecutive adversarial-review rounds, each with two new independent read-only reviewers, found no supported in-scope blockers.
- Trusted committed public custom-code budget: 3,311 / 4,000 lines.

## Limitations

Verification is deterministic and offline; no live API request was made. Sorbet recursive schemas remain intentionally unsupported and are outside this change.

### DIFF
--- a/lib/openai/helpers/structured_output/json_schema_converter.rb
+++ b/lib/openai/helpers/structured_output/json_schema_converter.rb
@@ -23,6 +23,15 @@ module OpenAI
             end
           end
           .freeze
+        # @api private
+        RECURSIVE = Object
+          .new
+          .tap do
+            _1.define_singleton_method(:inspect) do
+              "#<#{OpenAI::Helpers::StructuredOutput::JsonSchemaConverter}::RECURSIVE>"
+            end
+          end
+          .freeze
 
         # rubocop:disable Lint/UnusedMethodArgument
 
@@ -96,6 +105,10 @@ module OpenAI
             defs, path = state.fetch_values(:defs, :path)
             if (stored = defs[type])
               pointers = stored.fetch(OpenAI::Helpers::StructuredOutput::JsonSchemaConverter::POINTERS)
+              if pointers.first.fetch(:$ref).empty?
+                stored[OpenAI::Helpers::StructuredOutput::JsonSchemaConverter::RECURSIVE] = true
+              end
+
               pointers.first.except(OpenAI::Helpers::StructuredOutput::JsonSchemaConverter::NO_REF).tap do
                 pointers << _1
               end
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed `03f2fed2`. No substantive issues found. Recursive definitions remain in `$defs` without inlining their reference sites, preserving nullable and documented links without cyclic Ruby hashes. Ordinary acyclic reuse and the existing reference-escaping behavior remain unchanged.

PR #586, opened 2026-09-02
examples: handle structured response output variants

jbeckwith-oai:
## Summary

Structured Responses examples assumed every response output item was a message with content and every content part exposed parsed output. Valid Responses API results can place reasoning items or refusal parts before a later typed output-text part, so runnable examples could raise NoMethodError instead of printing the requested MathResponse or CalendarEvent.

This affects users copying:

- examples/responses/streaming_structured_outputs.rb
- examples/responses/streaming_previous_response.rb
- examples/structured_outputs_responses.rb

The examples now narrow output items to ResponseOutputMessage and content parts to ResponseOutputText before reading parsed, then print only the script's expected typed model.

Runnable public SDK shape:

    response = client.responses.create(
      model: "gpt-4o-2024-08-06",
      input: "Extract the event information.",
      text: CalendarEvent
    )

    response
      .output
      .grep(OpenAI::Models::Responses::ResponseOutputMessage)
      .flat_map(&:content)
      .grep(OpenAI::Models::Responses::ResponseOutputText)
      .each do |content|
        parsed = content.parsed
        next unless parsed.is_a?(CalendarEvent)

        pp(parsed)
      end

### Deterministic synthetic before/after

Before:

- output: [reasoning, message(output_text)] raised NoMethodError for nil.parsed.
- output: [message(refusal), message(output_text)] raised NoMethodError for ResponseOutputRefusal#parsed in both streaming examples.
- refusal-only streaming output raised NoMethodError instead of the examples' existing clear abort messages.

After:

- reasoning plus typed text prints an actual MathResponse or CalendarEvent.
- refusal plus typed text, including mixed refusal/output-text message content, prints the later actual typed model.
- refusal-only and JSON-null parsed output preserve script-specific behavior: both streaming examples keep their exact clear aborts, while the non-streaming example remains silent.

### Root cause and minimal fix

The extraction chain called content on every response.output union member and parsed on every message-content union member. The fix uses the SDK's existing concrete Responses union variants at the example boundary; it does not add helpers, duck typing, parser changes, or new exception policy.

### Compatibility and non-goals

This is a handwritten examples/test correction only. It does not change lib/, generated models/resources, signatures, dependencies, transport/stream parsing, public APIs, response creation/retrieval parameters, sequence handling, or output variants. Existing stream event printing, captured response IDs, cursor-zero resumption, and unknown-event behavior remain covered.

### Verification

- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/structured_response_example_variants_test.rb
  - 9 runs, 44 assertions, 0 failures, 0 errors
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/streaming_previous_response_example_test.rb
  - 9 runs, 74 assertions, 0 failures, 0 errors
- Supplied offline public-client/WebMock proofs for all three scripts
  - typed-only, reasoning plus typed, refusal plus typed succeed; refusal-only keeps expected behavior
- mise exec ruby@4.0.6 -- bundle exec rake test:examples:inventory
  - 24/35 exercised, 11 explicitly excluded
- mise exec ruby@4.0.6 -- bundle exec rake test
  - 1564 runs, 13989 assertions, 0 failures, 0 errors, 1 skip
- mise exec ruby@4.0.6 -- bundle exec rake format
  - examples Sorbet, RBS validation, RuboCop: 2802 files inspected, no offenses
- git diff --check
- Adversarial review: four rounds; round 2 found and fixed one WebMock teardown leak; rounds 3 and 4 were consecutive clean rounds
- Public custom-code budget against main 5baeac8131511141491fc41abdb2e4e4c596f3de and candidate e9cdee8ada2cd150a78a123b5a262e0a8c62c7fd
  - 3311 / 4000 custom lines, 689 headroom

### Limitations

No live API request was made; verification uses deterministic synthetic HTTP/SSE fixtures through the public client. This is not a security-sensitive change.

### DIFF
--- a/examples/responses/streaming_previous_response.rb
+++ b/examples/responses/streaming_previous_response.rb
@@ -169,7 +169,9 @@ begin
   parsed_output_received = false
   response
     .output
-    .flat_map { _1.content }
+    .grep(OpenAI::Models::Responses::ResponseOutputMessage)
+    .flat_map(&:content)
+    .grep(OpenAI::Models::Responses::ResponseOutputText)
     .each do |content|
       parsed = content.parsed
       next unless parsed.is_a?(MathResponse)
--- a/examples/responses/streaming_structured_outputs.rb
+++ b/examples/responses/streaming_structured_outputs.rb
@@ -40,7 +40,9 @@ puts("----- parsed outputs from final response -----")
 parsed_output_received = false
 response
   .output
-  .flat_map { _1.content }
+  .grep(OpenAI::Models::Responses::ResponseOutputMessage)
+  .flat_map(&:content)
+  .grep(OpenAI::Models::Responses::ResponseOutputText)
   .each do |content|
     # parsed is an instance of `MathResponse`
     parsed = content.parsed
--- a/examples/structured_outputs_responses.rb
+++ b/examples/structured_outputs_responses.rb
@@ -50,10 +50,13 @@ response = client.responses.create(
 
 response
   .output
-  .flat_map { _1.content }
-  # filter out refusal responses
-  .grep_v(OpenAI::Models::Responses::ResponseOutputRefusal)
+  .grep(OpenAI::Models::Responses::ResponseOutputMessage)
+  .flat_map(&:content)
+  .grep(OpenAI::Models::Responses::ResponseOutputText)
   .each do |content|
     # parsed is an instance of `CalendarEvent`
-    pp(content.parsed)
+    parsed = content.parsed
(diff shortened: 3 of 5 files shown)


HAYDEN-OAI (approved):
Reviewed `e9cdee8a`. No substantive issues found. Filtering the Responses output and content unions to message/text variants preserves later typed structured output after reasoning or refusal entries, while retaining the existing stream-resumption and missing-output behavior.

PR #591, opened 2026-09-03
feat(api): allow omitted function-call argument event names

markstuart-oai:
## Summary

Function-call argument completion events allow name to be absent.

## Changes

- Make name optional on stable and beta function-call argument completion events.
- Handle nil when reading name, or use the corresponding function-call item's name.


sylvesterkaczmarek (commented):
The generated RBI still types `name` as `String` in the writer/initializer/to_h shape while the model now allows it to be absent and the reader is `T.nilable(String)`. Sorbet callers cannot explicitly pass/set `nil`, and the declared hash shape is unsound for a missing name. Could these generated signatures use `T.nilable(String)` consistently?

PR #595, opened 2026-09-03
fix: preserve text logprobs in response streams

jbeckwith-oai:
## Problem
When customers request message.output_text.logprobs, responses.stream_raw retains server-provided token, logprob, and top_logprobs data, but responses.stream reconstructs enriched text delta and done events without that field. Normal streams lose both events; resumed streams lose the done event.

## Fix
At the existing handwritten enrichment boundary, forward the raw logprobs field only when it is present while rebuilding response.output_text.delta and response.output_text.done events. This preserves nonempty and empty arrays without eagerly coercing an omitted required generated field. The resumed no-snapshot delta path already passes through the raw event.

## Customer impact
Responses stream consumers now receive the same server-provided text logprob data available from raw streaming, including nested top_logprobs and numeric values, for normal and server-resumed streams.

## Synthetic proof
A WebMock public-entrypoint regression compares raw and enriched text events for normal and resumed streams. It covers nested nonempty arrays, explicit empty arrays, and omitted fields; verifies event types, IDs, indexes, sequence numbers, snapshot/no-snapshot behavior, and typed nested values. Before the fix, enriched normal delta/done values were nil and resumed done was nil. After the fix, provided arrays match raw events.

## Compatibility and scope
No logprob aggregation or recalculation. No parsing, structured output, snapshot, final-response, get_output_text, resume filtering, generated model, signature, dependency, CI, budget, or public API changes. Omitted ordinary text-only fields remain tolerated. The change only returns existing API data through existing event objects and adds no logging, storage, transport, deserialization, or new exposure.

## Verification
- New regression: 2 runs, 54 assertions, 0 failures, 0 errors
- Adjacent isolation and unknown-event tests: 3 runs, 24 assertions, 0 failures, 0 errors
- Existing Responses streaming suite: 33 runs, 122 assertions, 0 failures, 0 errors
- Full rake test: 1567 runs, 14157 assertions, 0 failures, 0 errors, 1 skip
- rake lint: passed, including RuboCop, RBS validation, directives, rubyfmt, and Sorbet example typecheck
- Supplied public-path reproduction: reproduced before and passes after
- Trusted public custom-code budget: 3311 / 4000 lines, 689 headroom

## Review
Four adversarial-review rounds used two fresh independent read-only reviewers each. Round 2 found and prompted the omitted-logprobs compatibility correction. Rounds 3 and 4 were consecutive clean rounds on unchanged final code.

## Limitations
Validation uses safe synthetic SSE only; no live API calls were made.

### DIFF
--- a/lib/openai/helpers/streaming/response_stream.rb
+++ b/lib/openai/helpers/streaming/response_stream.rb
@@ -114,6 +114,7 @@ module OpenAI
                   output_index: event.output_index,
                   sequence_number: event.sequence_number,
                   type: event.type,
+                  **event.to_h.slice(:logprobs),
                   snapshot: content.text
                 )
             else
@@ -146,6 +147,7 @@ module OpenAI
                 sequence_number: event.sequence_number,
                 text: event.text,
                 type: event.type,
+                **event.to_h.slice(:logprobs),
                 parsed: parsed
               )
 
--- a/test/openai/helpers/response_stream_text_logprobs_test.rb
+++ b/test/openai/helpers/response_stream_text_logprobs_test.rb
@@ -1,0 +1,173 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
+class OpenAI::Test::ResponseStreamTextLogprobsTest < Minitest::Test
+  extend Minitest::Serial
+  include WebMock::API
+
+  def before_all
+    super
+    WebMock.enable!
+  end
+
+  def after_all
+    WebMock.disable!
+    super
+  end
+
+  def setup
+    super
+    @client = OpenAI::Client.new(base_url: "http://localhost", api_key: "test-key")
+  end
+
+  def teardown
+    WebMock.reset!
+    super
+  end
+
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed exact head `2ee6aeab6f3524b0087ae51a0864473d34b1bf96` against base `0daedb2878373da899dac8b029fdb0523a26c6ef`. The two handwritten enrichment points now preserve a raw text event's `logprobs` key without synthesizing it when absent: this fixes normal delta/done events and resumed done events while leaving the already-preserving resumed delta path unchanged. I read the complete two-file diff and traced normal/raw/resumed entry points, SSE decoding, event unions and nested logprob models, BaseModel presence/coercion semantics, snapshot/final-response handling, stream lifecycle, signatures, and the repository's compatibility and security boundaries. I found no substantive correctness, compatibility, security, reliability, performance, or maintainability issues in this snapshot. Source-only review; tests and CI were intentionally not run or assessed.

PR #604, opened 2026-09-03
feat(api): add cache diagnostics and fix optional RBS keys

apcha-oai:
## Summary

Adds prompt-cache diagnostics and request options, and corrects Ruby RBS records to distinguish omitted keys from explicit null values.

## Changes

- Add prompt-cache diagnostics and comparison-response options to Responses and Beta Responses models
- Mark omittable RBS model and request-option keys as optional while preserving required and nullable fields
- Preserve optional RBS fields through formatting, with focused type and serialization regressions
- Refresh the API reference with current service-tier and rate-limit documentation

## Validation

- Fixed the RBS formatter to preserve optional record keys and regenerated the signatures. Compared parsed records across 973 changed RBS files: all 19,268 existing field entries retain their names and value types, with no declarations removed.
- The 282 removed `& OpenAI::Internal::Type::request_parameters` intersections are replaced by inline `?request_options: OpenAI::request_opts` fields. The intended change is key optionality; nullable value types remain unchanged.
- Focused formatter, serialization, and type-contract tests passed. The Ruby 3.3, 3.4, and 4.0 suites passed before promotion.


HAYDEN-OAI (approved):
Reviewed the complete 999-file base-to-head diff, including structured reconciliation of all 973 regenerated RBS files. The formatter preserves all 19,268 existing record fields and correctly distinguishes omitted keys from nullable values; all 282 request-parameter intersections are replaced by optional inline request options. Prompt-cache diagnostics and comparison options are consistent across Responses and Beta Responses, with matching runtime, RBS/RBI, serialization, and focused regression coverage. The service-tier and rate-limit reference updates also match the transformed API surface.

PR #607, opened 2026-09-04
fix: expose response metadata on binary results

jbeckwith-oai:
## API contract

For endpoints that directly return a binary `StringIO`, the SDK still returns that same writable `StringIO` object. The object now also exposes `last_response` and `_request_id`, so callers can read headers and request IDs without losing the familiar IO API.

With `include_raw_body: true`, `last_response.raw_body` is a separate frozen copy of the bytes; reading or writing the returned IO does not mutate the retained raw body. Binary endpoints stay binary even when the server labels the response as JSON. Streaming responses expose `_request_id` on the stream object. Union-backed text fallbacks and user-created `StringIO` values are unchanged.

## Blocked by

- https://github.com/openai/openai/pull/1396337 — Castiron source support
- https://github.com/openai/openai-ruby-internal/pull/80 — generated loader, carrier, and signatures

This draft contains only the SDK-owned runtime, documentation, and tests. It intentionally does not duplicate the generated `ResponseCarrier` files from the internal generator artifact.

## What changed

- Attach response metadata to direct binary `StringIO` results.
- Preserve a frozen raw-body copy when requested.
- Expose request IDs on streaming results.
- Document the behavior and add focused regressions for binary, streaming, YAML, and fallback cases.

## Validation

- `bundle exec ruby -Ilib -ropenai/response_carrier -Itest test/openai/client_test.rb` (with the generated carrier from #80 on `RUBYLIB`)
- `bundle exec ruby -Ilib -ropenai/response_carrier -Itest test/openai/resources/responses/streaming_test.rb` (with the generated carrier from #80 on `RUBYLIB`)
- `mise exec ruby@4.0.6 -- bundle exec rubocop lib/openai/internal/transport/base_client.rb lib/openai/internal/type/base_stream.rb lib/openai/request_options.rb test/openai/client_test.rb test/openai/resources/responses/streaming_test.rb`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -936,8 +936,9 @@ a supplied logger are isolated and never replace an API result or API error.
 ### Response metadata and request IDs
 
 OpenAI recommends logging request IDs in production so requests can be traced
-during troubleshooting. Top-level models and pages returned by the client expose
-immutable HTTP response metadata through `last_response`:
+during troubleshooting. Top-level models, pages, streams, and binary StringIO
+values returned by the client expose immutable HTTP response metadata through
+`last_response`:
 
 ```ruby
 response = openai.responses.create(model: "gpt-5.2", input: "Say 'this is a test'.")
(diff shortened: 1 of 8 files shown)


jbeckwith-oai:
Superseded by #608, which uses Castiron's public promotion path to include the exact generated snapshot together with this SDK-owned runtime patch. Keeping #608 as the single blocked draft avoids a duplicate candidate and makes the public checkout runnable.

PR #608, opened 2026-09-04
fix(api): expose metadata on binary responses

jbeckwith-oai:
## API contract

For endpoints that directly return a binary `StringIO`, the SDK still returns that same writable `StringIO` object. The object now also exposes `last_response` and `_request_id`, so callers can inspect headers and request IDs without losing the normal IO API.

With `include_raw_body: true`, `last_response.raw_body` is a separate frozen copy of the bytes; reading or writing the returned IO does not mutate the retained body. Binary endpoints stay binary even when a server labels the response as JSON. Streams expose `_request_id`. Union-backed plain-text fallbacks and user-created `StringIO` values are unchanged.

## Promotion provenance

- Castiron source support landed in https://github.com/openai/openai/pull/1396337.
- The generated loader/carrier/RBI/RBS output was promoted from https://github.com/openai/openai-ruby-internal/pull/80, which was closed after promotion.

This remains a draft for SDK review.

## Changes

- Add the generated `ResponseCarrier` loader, runtime module, and RBI/RBS intersections for direct binary endpoints.
- Attach response metadata only to direct binary `StringIO` results and preserve raw-body isolation.
- Expose `_request_id` on streams.
- Document the behavior and add focused regressions for binary, streaming, YAML, and fallback cases.

## Validation

- `bundle exec ruby -Ilib -Itest test/openai/client_test.rb`
- `bundle exec ruby -Ilib -Itest test/openai/resources/responses/streaming_test.rb`
- `bundle exec rake typecheck`
- `bundle exec rubocop lib/openai/internal/transport/base_client.rb lib/openai/internal/type/base_stream.rb lib/openai/request_options.rb test/openai/client_test.rb test/openai/resources/responses/streaming_test.rb`
- `git diff --check`


jbeckwith-oai:
@openai/sdks-team review requested. This draft was promoted from the Castiron-generated artifact in openai-ruby-internal#80; the generated loader/carrier/RBI/RBS output is preserved, with the SDK-owned runtime/docs/tests layered on top. It remains blocked on openai/openai#1396337 landing first.

markstuart-oai (commented):
Reviewed the current head across binary response decoding, response metadata, generated RBI/RBS surfaces, stream accessors, documentation, and focused coverage. The existing inline discussion captures the remaining issue; I found no additional distinct concerns.

lib/openai/internal/transport/base_client.rb
```diff
@@ -912,6 +918,8 @@ def request(req)
             OpenAI::Internal::Type::Converter.coerce(model, unwrapped).tap do |result|
               if result.is_a?(OpenAI::Internal::Type::BaseModel)
                 result._set_last_response(response_metadata)
+              elsif result.is_a?(StringIO) && direct_binary_response
+                result.extend(OpenAI::ResponseCarrier)._set_last_response(response_metadata)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 49a4ee1c. Direct binary responses now also extend a private duplication helper: `dup` delegates to StringIO, reattaches ResponseCarrier plus the helper to the copy, and reapplies the same immutable response metadata. The regression covers both `content.dup` and `content.dup.dup`, including `_request_id`, `last_response`, and YAML suppression.

lib/openai/internal/transport/base_client.rb
```diff
@@ -7,6 +7,22 @@
 module OpenAI
   module Internal
     module Transport
+      # Preserve response metadata behavior when callers duplicate a direct
+      # binary response.
+      #
+      # @api private
+      module DuplicableResponseCarrier
```
markstuart-oai: This helper is cohesive, but it pushes `base_client.rb` from 981 to 1,007 lines and gives the transport file ownership of `ResponseCarrier` duplication. Could the duplication behavior live alongside `ResponseCarrier` (or in a small dedicated internal type) so `parse_response` only attaches it? That would keep this already-large transport class focused without changing behavior.

markstuart-oai (approved):
Reviewed the replacement head across direct binary decoding, response metadata retention, StringIO duplication and YAML behavior, stream request IDs, generated RBI/RBS surfaces, documentation, and focused coverage. The previous duplication concern is resolved, and the exact-head test, type, lint, package, security, and Castiron checks are green.

lib/openai/internal/transport/base_client.rb
jbeckwith-oai: Fixed in 8536df2f. I moved the duplication behavior into the new private `OpenAI::Internal::Type::DuplicableResponseCarrier` helper; `base_client.rb` now only requires and attaches that helper at the direct-binary boundary. The existing focused regression still covers first/repeated `dup`, metadata identity, and YAML suppression.

PR #611, opened 2026-09-04
fix: parse quoted response charsets

### LINKED ISSUE #610: Quoted response charset parameters are ignored
## Summary

Raw/binary response decoding ignores a quoted `charset` parameter in `Content-Type`. For example, `text/plain; charset="UTF-8"` returns a `StringIO` whose bytes remain tagged `ASCII-8BIT`, while the equivalent unquoted `charset=UTF-8` is correctly tagged `UTF-8`.

Quoted parameter values are valid media-type syntax, so a proxy or origin that quotes the charset changes SDK behavior even though the declared encoding is the same.

## Reproduction

On current `main` at `32989f4b2d45edb4b2104a40f7561fb06c3341ee`:

```ruby
stub_request(:get, "http://localhost/files/file_123/content")
  .to_return(
    status: 200,
    headers: {"Content-Type" => 'text/plain; charset="UTF-8"'},
    body: "\xC3\xA9".b
  )

client = OpenAI::Client.new(base_url: "http://localhost", api_key: "test")
content = client.files.content("file_123")

content.read.encoding
# => ASCII-8BIT
```

The same response with `Content-Type: text/plain; charset=UTF-8` returns `Encoding::UTF_8`.

## Root cause

`OpenAI::Internal::Util.force_charset!` extracts the charset with:

```ruby
/charset=([^;\s]+)/
```

For a quoted parameter the captured value is `"UTF-8"`, including the quotes. `Encoding.find` rejects that name, the `ArgumentError` is intentionally suppressed, and the body keeps its binary encoding.

## Expected behavior

Quoted and unquoted charset parameters should resolve to the same Ruby `Encoding` without changing response contents or the existing `StringIO` return type.

sylvesterkaczmarek:
## Summary

Handle quoted `charset` parameters when decoding raw response bodies.

`Content-Type: text/plain; charset="UTF-8"` currently leaves the returned `StringIO` tagged `ASCII-8BIT` because the quotes are included in the value passed to `Encoding.find`. The decoder now accepts quoted and unquoted charset values consistently.

Fixes #610.

## Testing

- public `files.content` regression: passed on Ruby 4.0.6
- `test/openai/client_test.rb`: 48 passed, 232 assertions
- `test/openai/internal/util_test.rb`: 62 passed, 286 assertions
- `bundle exec rake lint:rubocop`: 2,869 files inspected, no offenses
- `git diff --check`: passed

No response contents or return types change; only the declared Ruby string encoding is corrected when the charset value is quoted.

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -654,7 +654,8 @@ module OpenAI
         # @param content_type [String]
         # @param text [String]
         def force_charset!(content_type, text:)
-          charset = /charset=([^;\s]+)/.match(content_type)&.captures&.first
+          match = /(?:\A|;)\s*charset\s*=\s*(?:"([^"]+)"|([^;\s]+))/i.match(content_type)
+          charset = match&.captures&.compact&.first
 
           return unless charset
 
--- a/test/openai/client_test.rb
+++ b/test/openai/client_test.rb
@@ -1147,6 +1147,20 @@ class OpenAITest < Minitest::Test
     end
   end
 
+  def test_binary_content_honors_quoted_response_charset
+    stub_request(:get, "http://localhost/files/file_quoted_charset/content")
+      .to_return(
+        status: 200,
+        headers: {"Content-Type" => "text/plain; charset=\"UTF-8\""},
+        body: "\xC3\xA9".b
+      )
+
+    openai = OpenAI::Client.new(base_url: "http://localhost", api_key: "My API Key")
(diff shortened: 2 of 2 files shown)


jbeckwith-oai:
Apologies — we had already addressed this separately in #613, which has now merged. Thank you for identifying the issue and putting together a thoughtful fix and reproduction; we appreciate the contribution. Closing this as superseded by #613.

PR #616, opened 2026-09-04
fix: preserve chat stream moderation results

jbeckwith-oai:
## Problem

Moderated Chat Completions can deliver the existing moderation field on streamed raw chunks, including a final moderation-only chunk with empty choices. The chat streaming helper accumulated content and usage but dropped moderation, so current snapshots and get_final_completion returned nil even though stream_raw exposed the result.

## User impact

Applications using client.chat.completions.stream for moderated completions now receive the same moderation result already available through client.chat.completions.stream_raw. Unmoderated streams remain nil.

## Fix

The helper now carries moderation into the initial parsed snapshot and, for later non-nil chunks, deep-dumps and coerces the field into the existing final ChatCompletion moderation type. Omitted or explicit nil later values do not erase a previously received moderation result. Raw chunks are not mutated.

## Synthetic regression proof

The focused WebMock SSE test uses only fake local data through the public stream_raw and stream APIs. Before this fix, a final empty-choices moderation chunk retained nested moderation data in stream_raw while helper snapshots and final completion returned nil. After this fix, snapshots and final completion retain the same nested input/output success and error variants, boolean flags, and numeric scores. The test also covers initial moderation surviving later omitted and nil fields, and the unmoderated nil control.

## Compatibility and scope

No public signatures, generated models, dependencies, transport behavior, parsing timing, event types, safety policy, tool-call accumulation, logprob accumulation, or payload limits change. Content, choices, usage, event ordering, and raw chunk values remain covered. The diff is limited to the handwritten stream helper and one focused regression test.

## Verification

- Focused regression: 3 runs, 21 assertions, 0 failures, 0 errors
- Affected chat streaming checks: 36 runs, 234 assertions, 0 failures, 0 errors
- Full offline suite against the current-spec mock: 1626 runs, 14450 assertions, 0 failures, 0 errors, 1 skip
- rake lint: RuboCop inspected 2870 files with no offenses; Sorbet and 1278 RBS validations passed
- Trusted public custom-code budget on committed candidate: 3363 of 4000 custom lines, 637 headroom
- Dedicated bounded response deserialization and isolation security review: no new sensitive sink, logging path, policy decision, or raw-chunk mutation
- Adversarial review: two consecutive clean rounds, each with two new independent read-only reviewers

## Limitations

This is an offline synthetic regression proof. It does not claim live server observation and does not introduce moderation enforcement.

### DIFF
--- a/lib/openai/helpers/streaming/chat_completion_stream.rb
+++ b/lib/openai/helpers/streaming/chat_completion_stream.rb
@@ -154,6 +154,12 @@ module OpenAI
           completion_snapshot.usage = chunk.usage if chunk.usage
           completion_snapshot.system_fingerprint = chunk.system_fingerprint if chunk.system_fingerprint
           completion_snapshot.service_tier = chunk.service_tier if chunk.service_tier
+          if chunk.moderation
+            completion_snapshot.moderation = OpenAI::Internal::Type::Converter.coerce(
+              OpenAI::Chat::ChatCompletion::Moderation,
+              model_dump(chunk.moderation)
+            )
+          end
 
           completion_snapshot
         end
@@ -537,6 +543,7 @@ module OpenAI
               model: data[:model],
               choices: choices,
               usage: data[:usage],
+              moderation: data[:moderation],
               system_fingerprint: data[:system_fingerprint],
               service_tier: data[:service_tier]
             }
--- a/test/openai/helpers/chat_stream_moderation_test.rb
+++ b/test/openai/helpers/chat_stream_moderation_test.rb
@@ -1,0 +1,208 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
Treating moderation as sticky stream state is the correct accumulation rule: a moderation-only terminal chunk must update the snapshot, while later omitted or nil values must not erase an earlier result. The raw-stream parity assertions cover that boundary well.

markstuart-oai (approved):
The stream accumulator now carries moderation into the initial snapshot and updates it only from later non-nil chunks, preserving the existing sticky stream-state semantics. The focused public-API coverage exercises final moderation-only chunks, initial values, nil/omitted updates, and raw-stream isolation. Exact-head CI is green across the Ruby test matrix, lint, types, packaging, CodeQL, and Castiron. I could not rerun the focused test locally because the required Ruby 4.0.6 toolchain is not installed on this review box.

PR #622, opened 2026-09-04
feat(api): correct function argument completion event fields (openapi-545)

markstuart-oai:
## Summary

Removes the name model attribute from response.function_call_arguments.done to match the API response.

## Changes

- Remove the name attribute and constructor keyword from stable and beta Responses function argument completion models.
- Update code using the removed attribute or keyword to read name from the function-call item matched by item_id.


sylvesterkaczmarek (commented):
The generated Ruby surface is internally consistent: the runtime models, websocket/server-event wrappers, RBI signatures, and RBS signatures all drop `name` together. That matters because leaving the field in only the static or wrapper layer would still make a valid terminal arguments event unusable for some consumers. `item_id` remains sufficient to recover the corresponding function-call identity.

jbeckwith-oai (approved):
Reviewed the exact head. The stable and beta Responses models, websocket unions, RBI, and RBS stay consistent with the corrected contract, while function-call item and Realtime names remain unchanged. CI is green.

PR #623, opened 2026-09-04
fix: preserve UTF-8 SSE stream boundaries

jbeckwith-oai:
## Problem

Valid UTF-8 SSE streams could lose data at the decoder boundary when a supported custom `OpenAI::HTTPClient` returned ordinary Ruby `String` chunks instead of binary strings. A single leading UTF-8 BOM caused the first event to be ignored, and UTF-8-labeled fragmentation could drop later non-ASCII events and the terminal finish reason.

## User impact

Applications using custom HTTP transports could receive incomplete raw or helper streams even when the wire bytes were valid SSE. For example, a stream containing `你好世界 end` could return only `你好世界` with no `stop` finish reason, and a leading BOM could remove `start ` from `start end`.

## Fix

- Reassemble line fragments through a binary internal buffer so delimiter scanning does not depend on each Ruby fragment's encoding label.
- Preserve the transport's emitted line encoding and allow an initial `US-ASCII` prefix label to yield to later UTF-8 content.
- Reuse already-binary transport chunks to preserve existing large-payload memory behavior.
- Ignore exactly one UTF-8 BOM only at the start of SSE decoding; embedded and non-leading BOM content remains data.

## Verification

- Coordinator public probes, before: leading BOM lost `start `; UTF-8 fragments lost ` end` and `stop`.
- Coordinator public probes, after: plain/whole-BOM/split-BOM raw and helper results match; binary and UTF-8 fragments both preserve `你好世界 end` and `stop`.
- `mise exec ruby@4.0.6 -- bundle exec rake test TEST=test/openai/sse_utf8_boundary_test.rb` — 5 runs, 8 assertions, 0 failures, 0 errors.
- Selected util, streaming, helper, unknown-event, and large-payload tests — 156 runs, 758 assertions, 0 failures, 0 errors.
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — passed.
- `mise exec ruby@4.0.6 -- bundle exec rake typecheck` — passed.
- Serialized current-spec full suite with offline Steady mock — 1656 runs, 14670 assertions, 0 failures, 0 errors, 1 skip.
- Dedicated bounded streaming/deserialization security scan — 0 findings.
- Adversarial review — six fresh-context paired rounds; supported mixed-label and binary-copy findings were fixed; rounds 5 and 6 were consecutively clean on final code.
- Trusted public custom-code budget — 3363 / 4000, 637 lines headroom.

## Compatibility and scope

The change is limited to `decode_lines`/`decode_sse` and one focused regression test. It preserves LF/CRLF handling, ASCII/no-BOM streams, embedded BOM content, caller-owned fragment bytes and encodings, JSONL behavior, one-shot closing, and large-payload semantics. It does not change public APIs, charset policy, transport behavior, dependencies, generated files, SSE retry/event-ID semantics, EOF dispatch, `[DONE]`, or reconnection behavior.

## Limitations

Verification uses deterministic offline synthetic responses through the public custom transport path; no live API examples were run.

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -781,32 +781,37 @@ module OpenAI
       class << self
         # @api private
         #
-        # Assumes Strings have been forced into having `Encoding::BINARY`.
-        #
-        # This decoder is responsible for reassembling lines split across multiple
-        # fragments.
+        # This decoder is responsible for reassembling bytes split across multiple
+        # String fragments without depending on each fragment's encoding label
+        # while preserving the String encoding callers supplied.
         #
         # @param enum [Enumerable<String>]
         #
         # @return [Enumerable<String>]
         def decode_lines(enum)
           re = /(\r\n|\r|\n)/
-          buffer = String.new
+          buffer = String.new(encoding: Encoding::BINARY)
+          encoding = nil
           cr_seen = nil
 
           chain_fused(enum) do |y|
             enum.each do |row|
+              encoding = row.encoding if !row.empty? && (encoding.nil? || encoding == Encoding::US_ASCII)
               offset = buffer.bytesize
-              buffer << row
+              buffer << (row.encoding == Encoding::BINARY ? row : row.b)
               while (match = re.match(buffer, cr_seen&.to_i || offset))
                 case [match.captures.first, cr_seen]
                 in ["\r", nil]
                   cr_seen = match.end(1)
                   next
                 in ["\r" | "\r\n", Integer]
-                  y << buffer.slice!(..(cr_seen.pred))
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed both changed files and traced the SSE/JSONL decoders through the raw and helper streaming paths. The binary buffer keeps delimiter offsets consistent across UTF-8 fragments, and stripping one BOM after line reassembly preserves embedded and later BOMs. I found no substantive issues in the encoding-label handling, source-string ownership, CR/LF handling, cleanup, or existing binary large-payload path. Validation was static: Ruby is unavailable in this environment, so I could not independently run the regression suite.

PR #624, opened 2026-09-04
fix: parse SSE retry metadata as decimal

jbeckwith-oai:
## Problem

WHATWG SSE retry fields containing ASCII digits are decimal integers, but Ruby Integer(value) infers prefixes. On a valid Chat event stream, retry: 08 and retry: 09 raised ArgumentError before the following data event, while retry: 010 was interpreted as 8 instead of 10.

## User-visible behavior

A public Chat stream_raw call now accepts legal leading-zero retry metadata and still yields the following chunk:

```ruby
stream = client.chat.completions.stream_raw(
  model: "gpt-4o-mini",
)
stream.each { |chunk| puts chunk.choices.first.delta.content }
```

With the same valid synthetic data event after it, retry: 08 previously raised ArgumentError; now the chunk is delivered and the retry metadata is 8. Likewise, retry: 010 is stored as 10, not 8. Required digit-only values 0, 8, 08, 09, and 010 map to 0, 8, 8, 9, and 10.

## Fix

At the existing digit-only retry conversion in decode_sse, pass explicit base 10 to Integer. The existing ASCII-digit guard is unchanged, so malformed, signed, fractional, and non-ASCII retry values remain ignored.

## Scope and compatibility

The diff is limited to that one numeric conversion and a focused regression test. It does not add reconnection, retry policy, parser state, payload limits, API changes, dependencies, generated changes, or alter event data, event IDs, UTF-8/BOM handling, fragments, line endings, EOF dispatch, or stream closing behavior.

## Verification

- Red proof: the new focused test raised ArgumentError for 08 in both public stream and metadata paths before the fix.
- Focused regression: 2 runs, 10 assertions, 0 failures.
- Existing util SSE tests: 63 runs, 287 assertions, 0 failures.
- UTF-8 boundary tests: 5 runs, 8 assertions, 0 failures.
- Chat completion streaming tests: 22 runs, 107 assertions, 0 failures.
- Ruby formatting, RuboCop (2,880 files), and typecheck passed.
- Serialized full suite against the current-spec mock: 1,658 runs, 14,680 assertions, 0 failures, 0 errors, 1 skip.
- Trusted current-main public custom-code budget: 3,363 / 4,000 lines, 637 headroom.
- Two consecutive fresh adversarial-review rounds with two independent read-only reviewers each were clean.
- Bounded streaming/deserialization security review was clean; no new trust boundary, logging, allocation limit, state, or sink was introduced.

## Limitations

The full suite retains one existing skipped test and emitted existing dependency/runtime warnings. No live API examples were run.

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -871,7 +871,7 @@ module OpenAI
                 in "id" unless value.include?("\0")
                   current.merge!(id: value)
                 in "retry" if /^\d+$/ =~ value
-                  current.merge!(retry: Integer(value))
+                  current.merge!(retry: Integer(value, 10))
                 else
                 end
               else
--- a/test/openai/sse_decimal_retry_test.rb
+++ b/test/openai/sse_decimal_retry_test.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require_relative "test_helper"
+
+class OpenAI::Test::SseDecimalRetryTest < Minitest::Test
+  extend Minitest::Serial
+
+  def test_digit_only_retry_metadata_does_not_interrupt_raw_chat_stream
+    %w[0 8 08 09 010].each do |retry_value|
+      stream = raw_stream(retry_value)
+
+      assert_equal(1, stream.to_a.length, "retry=#{retry_value}")
+    ensure
+      stream&.close
+    end
+  end
+
+  def test_digit_only_retry_metadata_is_parsed_as_decimal
+    {
+      "0" => 0,
+      "8" => 8,
+      "08" => 8,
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed both changed files and the surrounding SSE decoding, raw Chat stream, and cleanup paths. Explicit base 10 fixes leading-zero retry values while preserving the existing digit-only guard. The regression tests cover decimal metadata and continued delivery through stream_raw for 0, 8, 08, 09, and 010. No substantive findings.

Static review only: Ruby is unavailable in this environment, so I did not run tests or independently reproduce the reported suite results.

PR #626, opened 2026-09-05
feat(api): type Responses WebSocket create events

jbeckwith-oai:
## Summary

Adds type signatures for creating Responses over WebSocket connections.

## Changes

- Add schema-derived keyword typing for Responses WebSocket create events
- Preserve compatibility with additional event fields


jbeckwith-oai:
The reviewed generated snapshot has been integrated into #562, which remains the single public review surface. Closing this temporary Castiron promotion PR to avoid duplicate review.

PR #627, opened 2026-09-05
fix: strip upload filenames without platform path rules

thegoodengineer:
## Problem

`FilePart.new(content, filename:)` passes the caller-supplied name through `::File.basename`, which applies the **host platform's** path rules:

https://github.com/openai/openai-ruby/blob/8e6abfd57a33ae7c7360f496f8baf15e1cf74db4/lib/openai/file_part.rb#L113-L115

On Windows that means `\` is a separator and `:` is a drive separator, so a filename is truncated at the first colon:

```ruby
OpenAI::FilePart.new(io, filename: "report: final.md").filename
# Linux/macOS => "report: final.md"
# Windows     => "report"

OpenAI::FilePart.new(io, filename: "notes:v2.txt").filename
# Windows     => "notes"
```

The name is uploaded to the API rather than resolved against the local filesystem, so the same code sends a different filename depending on where it runs, and on Windows it silently drops most of the name including the extension.

## Fix

Strip directory components directly against both separators instead of delegating to the platform:

```ruby
private def strip_directories(filename)
  filename.to_s.split(SEPARATORS).last.to_s
end
```

Content-derived names are deliberately left alone. `Pathname` content and path-backed `IO` still use `::File.basename`, because those really are local paths, and `test_path_backed_io_filename_preserves_platform_path_semantics` already pins that behavior.

## Behavior change

| input | before (Linux) | before (Windows) | after |
| --- | --- | --- | --- |
| `"report: final.md"` | `"report: final.md"` | `"report"` | `"report: final.md"` |
| `"nested/report.md"` | `"report.md"` | `"report.md"` | `"report.md"` |
| `"nested\report.md"` | `"nested\report.md"` | `"report.md"` | `"report.md"` |

Backslash now strips on every platform rather than only on Windows, which is what makes the result uniform.

## Tests

`test_explicit_filename_strips_directories_without_platform_path_rules` covers colon-bearing names, both separators, and the `Pathname` form. It fails on `main` on both platforms: on Windows via the colon cases, on Linux via the backslash case.


lib/openai/file_part.rb
```diff
@@ -121,5 +129,22 @@ def initialize(content, filename: nil, content_type: nil)
...
+    # locally, so drop separators directly and keep every other character, no
+    # matter which platform the SDK runs on.
+    #
+    # @api private
+    #
+    # @param filename [Pathname, String]
+    #
+    # @return [String]
+    private def strip_directories(filename)
+      filename.to_s.split(SEPARATORS).last.to_s
```
jbeckwith-oai: Please preserve the prior rejection of NUL bytes in explicit filenames. `File.basename("nested/safe\0name.txt")` raises `ArgumentError`, but `filename.to_s.split(SEPARATORS).last.to_s` accepts the value, so this changes existing validation behavior. Please reject NUL before splitting and add a focused regression test.

jbeckwith-oai (changes requested):
Thanks for the focused fix—this is a real cross-platform behavior problem, and the overall approach is appropriately scoped. One compatibility regression needs to be addressed before merging.

lib/openai/file_part.rb
jbeckwith-oai: Fixed in be6829e0: explicit filenames now reject NUL before separator stripping, preserving the prior `File.basename` validation behavior, and `test_explicit_filename_rejects_null_bytes` covers the regression.

lib/openai/file_part.rb
```diff
@@ -121,5 +129,25 @@ def initialize(content, filename: nil, content_type: nil)
...
+    # @api private
+    #
+    # @param filename [Pathname, String]
+    #
+    # @return [String]
+    private def strip_directories(filename)
+      name = filename.to_s
+      raise ArgumentError, "path name contains null byte" if name.include?("\0")
+
+      name.split(SEPARATORS).last.to_s
```
markstuart-oai: [P2] Preserve valid basenames when discarded directories contain invalid encoding bytes. `filename: "bad\xFF/safe.txt"` (UTF-8 or US-ASCII, also wrapped in `Pathname`) previously becomes `safe.txt` and successfully passes generated file-request dumping and multipart serialization. This regexp split validates the entire input and now raises `ArgumentError`, even though the invalid byte belongs only to a directory being discarded. I reproduced this against the actual base and this head. This is a concrete remaining case of the earlier encoding feedback; an invalid byte in the final basename already fails in the unchanged multipart escaper, so that broader upload claim is not needed. Please make this canonical stripping helper tolerate invalid bytes in discarded components while retaining the pre-strip NUL rejection, and add a regression through multipart serialization. Preserve valid multibyte characters too (e.g. Windows-31J `表`, whose encoding contains byte 0x5c), rather than unconditionally treating every 0x5c byte as a separator.

markstuart-oai (commented):
Reviewed all four changed files and the filename boundary through generated request dumping, multipart serialization, and vector-store staging. The focused NUL fix preserves rejection before stripping; the private helper and matching RBI/RBS declarations are appropriately scoped, with no material structural simplification or decomposition needed. One compatibility regression remains in the regexp split (inline).

Validation on Linux/Ruby 4.0.6 using available cached gems: FilePart, serialization-cursor, and multipart-content-type tests passed (29 tests/717 assertions); vector-store uploader tests passed (22/142). Independent base/head probes verified the reported invalid-directory regression, separator/colon/Unicode cases, NUL rejection, cursor preservation, and input mutation independence. Ruby/RBI syntax, RBS parsing, and diff whitespace checks passed. The broader util test run had nine CGI fixture errors; the locked bundle is unavailable, and Windows/full-suite validation was not run.

lib/openai/file_part.rb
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 2ff77a45. Explicit filename normalization now searches a byte-length-preserving scrubbed copy, then slices the original bytes, so invalid UTF-8 and mixed valid/invalid Shift_JIS names preserve their bytes and original encoding without mistaking multibyte trail bytes for separators. Added focused coverage for invalid UTF-8, Shift_JIS multibyte characters, mixed valid/invalid Shift_JIS, empty names, and root-only separators.

markstuart-oai (approved):
Re-reviewed 2ff77a45 after response 3944739700, including all four files in the base diff and the generated request/multipart/vector-store callers. My prior P2 (3944608475, review 5125978658) is fixed: invalid UTF-8/US-ASCII bytes in discarded directories now yield safe.txt through both file and upload-part serialization. Original bytes/encoding and valid Shift_JIS trail bytes are preserved; NUL rejection remains intact. The logic stays in the canonical FilePart boundary, with matching RBI/RBS contracts and no material structural regression or missed simplification.

Validated on Linux/Ruby 4.0.6 with cached gems: 56 focused tests, 876 assertions, all passing; 6,019 independent encoding/filename/serialization checks passed, plus explicit separator, drive/UNC, dot, trailing/root/empty, mutation and cursor probes. Ruby/RBI syntax, RBS parsing and diff checks passed. Windows, the locked bundle, full-suite/typechecker runs and live uploads were not run. No remaining findings.

lib/openai/file_part.rb
jbeckwith-oai: Fixed in 2ff77a45. The helper now searches a byte-length-preserving scrubbed copy and slices the original bytes, so invalid bytes in discarded directories no longer raise, while valid multibyte Shift_JIS characters, original encoding, and NUL rejection are preserved. Focused regression coverage was added.

jbeckwith-oai (approved):

PR #629, opened 2026-09-05
fix: match response charset parameters case-insensitively

thegoodengineer:
## Problem

`force_charset!` looks for the charset parameter with a case-sensitive pattern:

https://github.com/openai/openai-ruby/blob/8e6abfd57a33ae7c7360f496f8baf15e1cf74db4/lib/openai/internal/util.rb#L659-L660

Media type parameter names are case-insensitive ([RFC 9110 8.3.1](https://www.rfc-editor.org/rfc/rfc9110#section-8.3.1)), so `Charset=` names the same parameter as `charset=`. The pattern only matched the lowercase spelling:

```ruby
text = String.new.force_encoding(Encoding::BINARY)
OpenAI::Internal::Util.force_charset!("application/json; Charset=utf-8", text: text)
text.encoding  # => #<Encoding:BINARY (ASCII-8BIT)>
```

The charset is ignored and the body is handed back still tagged `Encoding::BINARY`, so callers get bytes where they expected decoded text.

## Fix

Add the `i` flag.

Every other media type pattern in this file already ignores case, so this one was the outlier:

- `JSON_CONTENT` and `JSONL_CONTENT` are both defined with `%r{...}i`
- the `text/event-stream` branch of `decode_content` uses `%r{...}i`

Charset *values* were already matched case-insensitively, since `Encoding.find` handles that; `test_charset` covers it.

## Tests

`test_charset_parameter_name_is_case_insensitive` covers `Charset=`, `CHARSET=`, and a mixed-case quoted form. It fails on `main` with `Encoding::BINARY` where `Encoding::UTF_8` is expected.

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -659,10 +659,13 @@ module OpenAI
         #
         # https://www.iana.org/assignments/character-sets/character-sets.xhtml
         #
+        # Media type parameter names are case-insensitive, so `Charset=` and
+        # `CHARSET=` name the same parameter as `charset=`.
+        #
         # @param content_type [String]
         # @param text [String]
         def force_charset!(content_type, text:)
-          charset = /charset=([^;\s]+)/.match(content_type)&.captures&.first
+          charset = /charset=([^;\s]+)/i.match(content_type)&.captures&.first
           charset = charset[1...-1] if charset&.start_with?("\"") && charset.end_with?("\"")
 
           return unless charset
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -691,7 +691,8 @@ module OpenAI
         #
         # Read a media type parameter value: either a token that ends at the first
         # whitespace, or a quoted string in which `\` escapes the next character.
-        # A quoted string that is never closed has no value.
+        # A quoted string that is never closed, or has non-whitespace after its
+        # closing quote, has no value.
         #
         # @param value [String]
         #
@@ -702,14 +703,16 @@ module OpenAI
           unquoted = String.new
           escaped = false
 
-          value[1..].to_s.each_char do |char|
+          value[1..].to_s.each_char.with_index do |char, index|
             if escaped
               unquoted << char
               escaped = false
             elsif char == "\\"
               escaped = true
             elsif char == "\""
-              return unquoted
+              return unquoted if value[(index + 2)..].to_s.match?(/\A\s*\z/)
+
+              return nil
             else
               unquoted << char
             end
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -657,18 +657,99 @@ module OpenAI
 
         # @api private
         #
-        # https://www.iana.org/assignments/character-sets/character-sets.xhtml
+        # Split a media type on its `;` delimiters, ignoring any that appear inside a
+        # quoted string so that a parameter value may contain one.
         #
-        # Media type parameter names are case-insensitive, so `Charset=` and
-        # `CHARSET=` name the same parameter as `charset=`.
+        # @param content_type [String]
+        #
+        # @return [Array<String>]
+        private def split_media_type_segments(content_type)
+          segments = [String.new]
+          quoted = escaped = false
+
+          content_type.to_s.each_char do |char|
+            if escaped
+              segments.last << char
+              escaped = false
+            elsif quoted && char == "\\"
+              segments.last << char
+              escaped = true
+            elsif char == "\""
+              segments.last << char
+              quoted = !quoted
+            elsif char == ";" && !quoted
+              segments << String.new
+            else
+              segments.last << char
+            end
+          end
+
+          segments
+        end
+
+        # @api private
+        #
+        # Read a media type parameter value: either a token that ends at the first
+        # whitespace, or a quoted string in which `\` escapes the next character.
+        # A quoted string that is never closed has no value.
+        #
+        # @param value [String]
+        #
+        # @return [String, nil]
+        private def read_media_type_value(value)
+          return value[/\A[^;\s]+/] unless value.start_with?("\"")
+
+          unquoted = String.new
+          escaped = false
+
+          value[1..].to_s.each_char do |char|
+            if escaped
+              unquoted << char
+              escaped = false
+            elsif char == "\\"
+              escaped = true
+            elsif char == "\""
+              return unquoted
+            else
+              unquoted << char
+            end
+          end
+
+          nil
+        end
+
+        # @api private
+        #
+        # Find the value of the `charset` media type parameter. Parameter names are
+        # case-insensitive, so `Charset` and `CHARSET` name the same parameter, but the
+        # name has to match in full: a `charset` suffix of another parameter's name, or
+        # a `charset=` inside a quoted value, belongs to a different parameter.
+        #
+        # @param content_type [String]
+        #
+        # @return [String, nil]
+        private def media_type_charset(content_type)
+          split_media_type_segments(content_type).each do |segment|
+            name, separator, value = segment.partition("=")
+            next if separator.empty?
+            next unless name.strip.casecmp?("charset")
+
+            return read_media_type_value(value.lstrip)
+          end
+
+          nil
+        end
+
+        # @api private
+        #
+        # https://www.iana.org/assignments/character-sets/character-sets.xhtml
         #
         # @param content_type [String]
         # @param text [String]
         def force_charset!(content_type, text:)
-          charset = /charset=([^;\s]+)/i.match(content_type)&.captures&.first
-          charset = charset[1...-1] if charset&.start_with?("\"") && charset.end_with?("\"")
+          charset = media_type_charset(content_type)
 
-          return unless charset
+          return if charset.nil? || charset.empty?
 
           begin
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Making the charset parameter-name match case-insensitive is the standards-correct change, and it is nicely scoped: the value handling was already case-insensitive through `Encoding.find`, while this was the remaining case-sensitive parameter-name path.

The regression covering `Charset=`, `CHARSET=`, and a mixed-case quoted value is useful because it tests both name normalization and preservation of the existing value parsing. This looks complete for the reported issue.

lib/openai/internal/util.rb
```diff
@@ -654,10 +654,13 @@ def encode_content(headers, body)
...
         # https://www.iana.org/assignments/character-sets/character-sets.xhtml
         #
+        # Media type parameter names are case-insensitive, so `Charset=` and
+        # `CHARSET=` name the same parameter as `charset=`.
+        #
         # @param content_type [String]
         # @param text [String]
         def force_charset!(content_type, text:)
-          charset = /charset=([^;\s]+)/.match(content_type)&.captures&.first
+          charset = /charset=([^;\s]+)/i.match(content_type)&.captures&.first
```
jbeckwith-oai: Making this regex case-insensitive also broadens its existing substring match. For example, Content-Type: text/plain; X-CHARSET=ISO-8859-1; charset=UTF-8 now matches the suffix inside X-CHARSET first and forces the wrong encoding; a quoted value such as title="note; Charset=ISO-8859-1;" has the same issue. Please match only a top-level media-type parameter named charset (respecting quoted strings and escapes) before comparing the name case-insensitively.

test/openai/internal/util_test.rb
```diff
@@ -851,6 +851,23 @@ def test_charset
...
+    # `Charset=` has to decode the same as one labelled `charset=`. Matching only
+    # the lowercase spelling left the body at `Encoding::BINARY`.
+    [
+      "application/json; Charset=utf-8",
+      "application/json; CHARSET=utf-8",
+      "application/json; ChArSeT=\"utf-8\""
+    ].each do |content_type|
+      text = String.new.force_encoding(Encoding::BINARY)
+
+      OpenAI::Internal::Util.force_charset!(content_type, text: text)
```
jbeckwith-oai: This regression test calls the private helper directly with application/json, but decode_content routes application/json through JSON.parse and never calls force_charset!. The caller-visible path is raw non-JSON and error bodies. Please exercise decode_content with text/plain; Charset=... and assert the returned StringIO body encoding so the test proves the behavior users observe.

jbeckwith-oai (changes requested):
Requesting changes for the two inline findings: the case-insensitive matcher can select charset-like text from unrelated parameters or quoted values, and the regression test does not exercise the caller-visible decode path. Please address both before merge.

lib/openai/internal/util.rb
thegoodengineer: Good catch, and the quoted-value case was broken before this PR too. Both are fixed in 03e48b8.

The header is now parsed rather than substring-matched: `split_media_type_segments` splits on `;` while tracking quoted strings and backslash escapes, and `media_type_charset` takes the first parameter whose name is exactly `charset` after `strip` and `casecmp?`. Values are read by `read_media_type_value` as either a token ending at whitespace or a quoted string, which also replaces the ad-hoc unquoting that just stripped a leading and trailing `"`.

Your two examples, and an escaped quote inside a value, now resolve as expected:

| `Content-Type` | before | after |
| --- | --- | --- |
| `text/plain; X-CHARSET=ISO-8859-1; charset=UTF-8` | ISO-8859-1 | UTF-8 |
| `text/plain; title="note; charset=ISO-8859-1;"` | ISO-8859-1 | BINARY |
| `text/plain; charset="UTF-8"; x="\"charset=ISO-8859-1\""` | ISO-8859-1 | UTF-8 |

Worth noting for review: the existing `test_charset` cases put `charset=` in the first segment, before the media type (`"charset=uTf-8 application/json; "`), so the scan deliberately considers every segment rather than skipping segment 0. Those cases still pass unchanged.

test/openai/internal/util_test.rb
thegoodengineer: You're right, and I confirmed it: `decode_content` with `application/json` returns a parsed Hash and never reaches `force_charset!`, so the old test proved nothing about what a caller observes.

Both regression tests now go through `decode_content` with a raw `text/plain` body and assert the encoding of the returned `StringIO`, via a small `decoded_body_encoding` helper.

I checked the tests fail for the right reasons rather than just passing on the new code:

| lib under test | result |
| --- | --- |
| `main` (no fix) | both tests fail |
| the earlier `/i`-only commit | `test_charset_matches_only_a_parameter_named_charset` still fails |
| this branch | both pass |

That middle row is your first finding reproduced as a test, so the substring behavior can't come back silently.

thegoodengineer:
Both findings addressed in 03e48b8, and the branch is rebased onto current `main`.

**Parameter selection.** The header is parsed instead of substring-matched: split on `;` while tracking quoted strings and backslash escapes, then take the first parameter whose name is exactly `charset`, compared case-insensitively. Values are read as a token ending at whitespace or as a quoted string, which also replaces the ad-hoc unquoting that only stripped a leading and trailing `"`. This fixes the `X-CHARSET` case introduced by making the match case-insensitive, and the quoted-value case that predates this PR.

**Test path.** Both regression tests now go through `decode_content` with a raw `text/plain` body and assert the encoding of the returned `StringIO`. `application/json` is parsed to a Hash and never reaches `force_charset!`, so the previous test could not observe the caller-visible behavior.

Verification: against `main` both tests fail; against the earlier `/i`-only commit the parameter-selection test still fails; on this branch all pass. The pre-existing `test_charset` and `test_quoted_charset` are unchanged and still pass, including the cases that place `charset=` before the media type.

sylvesterkaczmarek (approved):
Re-reviewed 03e48b8. The parser now selects an exact charset parameter instead of substring-matching, while respecting quoted strings and escapes. The regressions also go through decode_content with text/plain, so they now exercise the caller-visible path. The X-CHARSET and quoted-value cases that motivated the requested changes are covered. No blocker from me.

jbeckwith-oai:
@sylvesterkaczmarek save your tokens!  We are happy to run the automated reviews on these :)

sylvesterkaczmarek:
Fair point — thanks. I’ll leave the automated review loop to the repo tooling and only jump in when there’s a specific human question or something substantive to add.

jbeckwith-oai (approved):
Re-reviewed the final rebased head 3277273. The charset parser now matches only exact case-insensitive top-level parameters, handles quoted/escaped values, rejects malformed quoted suffixes, and the regressions exercise decode_content. CI, CodeQL, Castiron, and automated code/security reviews are all green; no outstanding review threads remain.

lib/openai/internal/util.rb
```diff
@@ -655,17 +655,104 @@ def encode_content(headers, body)
...
+        #
+        # Read a media type parameter value: either a token that ends at the first
+        # whitespace, or a quoted string in which `\` escapes the next character.
+        # A quoted string that is never closed, or has non-whitespace after its
+        # closing quote, has no value.
+        #
+        # @param value [String]
+        #
+        # @return [String, nil]
+        private def read_media_type_value(value)
```
markstuart-oai, line 700: [P2] Decompose the media-type parser and remove duplicate quote-state logic

This pushes the file past 1k lines. Can we decompose this first? `util.rb` grows from 985 lines at the merge base to 1,072, with a new three-helper parser embedded alongside query encoding, multipart writing, stream decoding, and other unrelated utilities. Both `split_media_type_segments` and this reader now maintain quote/escape traversal logic; this reader additionally manages slice offsets and closing-quote suffix validation.

Please put the response media-type parsing contract in a focused internal module and simplify the value reader rather than merely moving these state machines. Once the splitter has isolated a value, an anchored quoted-string match followed by quoted-pair unescaping can replace this reader's loop, escape flag, index arithmetic, and nested returns. I checked that model against this reader on 10,000 deterministic inputs without a mismatch. Keep exact case-insensitive names, first-match behavior, legacy first-segment/token handling, and rejection of unterminated or suffixed quoted values covered through `decode_content`.

I checked the existing alternatives: `Net::HTTPHeader#type_params` splits blindly on semicolons, and `FilePart.validate_content_type` validates strict outbound headers rather than extracting a permissive response parameter. Neither is a drop-in replacement.

markstuart-oai (commented):
Reviewed all changes at 327727397fa28d1905319f018b599f7321e4bde4 against merge base 55ae0137740f05f30173e3dc73d546a500b29d2a, including response/error callers and existing parsing contracts. The earlier exact-parameter and caller-visible test findings are addressed; I independently reproduced both new regressions against the base implementation. One structural issue remains below, so I am not approving this implementation yet.

Validation on Ruby 4.0.6: 4 charset tests/16 assertions, 9 content-type tests/45 assertions, and 3 client contract tests/28 assertions passed. Additional temporary probes passed for 21 charset inputs, byte preservation, public success/error decoding, JSON and direct-binary compatibility, and 10,000 deterministic quoted-reader equivalence cases. The broader util/content-type run had 77 tests/333 assertions with 9 errors in multipart FakeCGI tests (missing params/[] in the available CGI); no assertion failures. Full suite and lint/typecheck were not run.

PR #630, opened 2026-09-05
fix: upload pathless IO streams instead of raising

thegoodengineer:
## Problem

`write_multipart_chunk` derives a filename for a bare `IO` from its path:

https://github.com/openai/openai-ruby/blob/8e6abfd57a33ae7c7360f496f8baf15e1cf74db4/lib/openai/internal/util.rb#L557-L559

`IO#to_path` is `nil` for a stream with no backing path, such as a pipe or a socket, and `::File.basename(nil)` raises:

```ruby
reader, writer = IO.pipe
writer.write("hello")
writer.close

OpenAI::Internal::Util.encode_content(
  {"content-type" => "multipart/form-data"},
  {file: reader}
).last.to_a
# TypeError: no implicit conversion of nil into String
```

The failure happens while the body is being encoded, before any request goes out, and the message points at neither the stream nor the field.

## Fix

Guard the branch and send the part without a filename, which [RFC 7578 §4.2](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) makes optional:

```ruby
in Pathname | IO unless val.to_path.nil?
```

`Pathname#to_path` always returns a `String`, so only the `IO` case is affected.

The rest of the SDK already treats a nil `to_path` as "no filename", which is what makes this branch the outlier:

- `FilePart#initialize` guards it with `content.to_path&.then { ::File.basename(_1) }`
- `VectorStoreFileUploader#stage_file` falls back to `"upload"`

A caller who wants a name for such a stream can still pass one with `OpenAI::FilePart.new(io, filename: "...")`, which is matched by the earlier branch.

## Tests

`test_multipart_pathless_io_is_sent_without_a_filename` encodes an `IO.pipe` reader and asserts the part carries the field name and body but no `filename=`. It errors on `main` with the `TypeError` above.


sylvesterkaczmarek (commented):
This fits the existing multipart model well: a pathless stream is still valid form-data, it simply has no filename, and callers that need one can already wrap the stream in `OpenAI::FilePart` with an explicit filename.

The guard fixes the failure at the encoder boundary without inventing a synthetic path or changing behavior for path-backed `IO` and `Pathname`. The `IO.pipe` regression also exercises cleanup and proves the request body still carries the field and payload. I do not see a correctness issue with this change.

jbeckwith-oai (approved):
Reviewed the pathless IO multipart hardening. The change is narrowly scoped, preserves existing public FileInput normalization and path-backed filename behavior, and passed two clean adversarial-review rounds.

sylvesterkaczmarek:
One follow-up after merge: I opened #664 for the generic `Client#request` multipart path, where a raw pathless IO can still be serialized without a filename even though `FileInput.dump` uses the `upload` fallback. Linking it here so the edge case does not get lost.

PR #631, opened 2026-09-05
fix: parse SSE field lines that carry no value

thegoodengineer:
## Problem

`decode_sse` matches field lines with a pattern that requires a colon:

https://github.com/openai/openai-ruby/blob/8e6abfd57a33ae7c7360f496f8baf15e1cf74db4/lib/openai/internal/util.rb#L845-L846

The [event stream format](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) this function already cites says a non-empty line without a colon is a field whose name is the whole line and whose value is the empty string. Those lines fell through to the `else` and were dropped.

The visible effect is on `data`, where a valueless line contributes a blank line to the buffer:

```
data: line 1
data
data: line 2
```

```ruby
# main  => {data: "line 1\nline 2\n"}
# here  => {data: "line 1\n\nline 2\n"}
```

A blank line is silently collapsed out of the middle of the value, so a payload that encodes paragraph breaks that way comes back subtly wrong.

## Fix

Make the colon and value optional and treat a missing value as the empty string:

```ruby
in /^([^:]+)(?::\s?(.*))?$/
```

- `data` appends a blank line, as above
- `event` and `id` set empty buffers, per the spec
- `retry` still ignores the value, since `""` is not a run of digits
- comment lines are unaffected; the earlier `/^:/` branch still matches them first

Appending to the data buffer changes from `<< (value << "\n")` to `<< value << "\n"`. The old form mutated the captured value, which now raises, because a valueless line yields the frozen empty string from `nil.to_s`. The buffer is built the same way either way.

## Tests

Three cases added to `test_decode_sse`, covering a valueless `data` line, valueless `event`/`id` lines, and a valueless `retry` line. The `data` case fails on `main`.


sylvesterkaczmarek (commented):
I checked the parser change against the SSE field-line rules, and the handling here is internally consistent. A non-empty line without a colon should be treated as a field name with an empty value, so `value.to_s` gives the right representation without adding special cases for each field.

The change to the `data` append is also important rather than cosmetic: `<< value << "\n"` preserves the buffer construction without mutating the captured value, which matters now that the no-colon branch can produce the empty string from `nil`. The earlier comment-line branch still takes precedence, an empty `event` or `id` is retained, and an empty `retry` correctly fails the numeric guard. The three regressions cover the behaviours introduced by the broader matcher rather than only the motivating `data` case.

This looks like a clean standards-compliance fix with appropriately narrow scope.

jbeckwith-oai (commented):
Reviewed the SSE parser change against the WHATWG field-line rules and the surrounding stream decoder behavior. This is a real, narrowly scoped standards-compliance fix: a non-empty line without a colon must be processed as the whole field name with an empty value, so bare data/event/id/retry lines now behave as expected while comments and colon-bearing fields retain their existing paths.\n\nI found no blocking issues after adversarial review. The change stays confined to decode_sse and focused regression coverage, preserves literal field matching and the existing stream/hash contract, and introduces no new security concern.\n\nVerified locally with the repo-compatible Ruby and locked bundle:\n- bundle exec ruby -Itest test/openai/internal/util_test.rb: 63 runs, 290 assertions, 0 failures\n- bundle exec rubocop lib/openai/internal/util.rb test/openai/internal/util_test.rb: no offenses\n- git diff --check: passed

jbeckwith-oai (approved):
Approved after scope-bounded adversarial review and focused verification. No blocking findings.

PR #634, opened 2026-09-05
fix: type chat completion streams for Sorbet

jbeckwith-oai:
## Problem

The documented high-level Chat Completions streaming helper works at runtime, but the shipped RBI omitted `client.chat.completions.stream`, `ChatCompletionStream#text`, `#get_final_completion`, and other stream methods. It also advertised `parsed_arguments` on function-tool argument events even though runtime exposes `parsed`.

This is a Sorbet typing fix, not a runtime behavior change. A typed application using the supported high-level workflow failed compilation with five missing-method errors; the equivalent `stream_raw` call type-checked.

```ruby
  model: "gpt-4o-mini",
)

stream.text.each { |text| puts(text) }
```

## Change

- Add a handwritten Chat stream RBI that mirrors the existing `stream_raw` request keyword contract and returns the public `OpenAI::Streaming::ChatCompletionStream`.
- Type high-level stream event callbacks, eager text enumeration, final completion, current snapshot, output text, and lifecycle/response metadata inherited from the existing stream base.
- Correct the two function-tool argument event readers from nonexistent `parsed_arguments` to runtime-backed `parsed`.
- Add focused real-Sorbet positive and negative consumer checks plus an offline runtime shape check.

Synthetic before/after:

```text
before: Method stream does not exist on OpenAI::Resources::Chat::Completions
        Method parsed does not exist on ChatFunctionToolCallArgumentsDoneEvent
after:  No errors! Great job.
        parsed_arguments remains a Sorbet error
```

## Compatibility and scope

Runtime code, request behavior, transport models, parser behavior, generated RBI files, RBS, and Responses streaming are unchanged. Existing parsed values remain `T.untyped`, matching the established contract. `text` is declared as the ordinary eager `Enumerator` returned at runtime, not a lazy enumerator.

Customer incidence is not measured; the deterministic impact is a compile-time failure for Sorbet consumers using this public workflow.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_sorbet_test.rb` — 3 runs, 11 assertions, 0 failures
- Original positive consumer probe via `SRB_SKIP_GEM_RBIS=1 ... srb typecheck` — no errors
- Existing `stream_raw` control probe via Sorbet — no errors
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_completion_stream_test.rb` — 10 runs, 47 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_fingerprint_test.rb` — 1 run, 32 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_terminal_tool_test.rb` — 3 runs, 16 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec rake format:rb`
- `mise exec ruby@4.0.6 -- bundle exec rake typecheck:sorbet`
- `mise exec ruby@4.0.6 -- bundle exec rake lint:rubyfmt`
- Serialized `TEST_API_BASE_URL=... mise exec ruby@4.0.6 -- bundle exec rake test` — 1,711 runs, 14,919 assertions, 0 failures, 0 errors, 1 skip
- Required adversarial review — two consecutive clean rounds, two fresh independent read-only reviewers per round
- Trusted committed-candidate public custom-code budget check — success, 3,365 / 4,000 lines

## Limitations

No live API request was made; verification uses synthetic offline data and the repository mock. This does not measure how many downstream Sorbet applications use Chat streaming.

### DIFF
--- a/rbi/openai/helpers/streaming/chat_completion_stream.rbi
+++ b/rbi/openai/helpers/streaming/chat_completion_stream.rbi
@@ -1,0 +1,164 @@
+# typed: strong
+
+module OpenAI
+  module Helpers
+    module Streaming
+      class ChatCompletionStream
+        include OpenAI::Internal::Type::BaseStream
+
+        ChatCompletionStreamEvent = T.type_alias do
+          T.any(
+            ChatChunkEvent,
+            ChatContentDeltaEvent,
+            ChatContentDoneEvent,
+            ChatRefusalDeltaEvent,
+            ChatRefusalDoneEvent,
+            ChatFunctionToolCallArgumentsDeltaEvent,
+            ChatFunctionToolCallArgumentsDoneEvent,
+            ChatLogprobsContentDeltaEvent,
+            ChatLogprobsContentDoneEvent,
+            ChatLogprobsRefusalDeltaEvent,
+            ChatLogprobsRefusalDoneEvent
+          )
+        end
+
+        Message = type_member { {fixed: ChatCompletionStreamEvent} }
+        Elem = type_member { {fixed: ChatCompletionStreamEvent} }
+
+        sig { returns(OpenAI::Chat::ChatCompletion) }
+        def get_final_completion
+        end
+
+        sig { returns(String) }
+        def get_output_text
+        end
+
+        sig { returns(T.self_type) }
+        def until_done
+        end
+
+        sig { returns(T.nilable(OpenAI::Chat::ChatCompletion)) }
+        def current_completion_snapshot
+        end
+
+        sig { returns(T::Enumerator[String]) }
+        def text
+        end
+      end
+    end
+  end
+
+  module Resources
+    class Chat
+      class Completions
+        sig do
+          params(
+            messages: T::Array[
+              T.any(
+                OpenAI::Chat::ChatCompletionDeveloperMessageParam::OrHash,
+                OpenAI::Chat::ChatCompletionSystemMessageParam::OrHash,
+                OpenAI::Chat::ChatCompletionUserMessageParam::OrHash,
+                OpenAI::Chat::ChatCompletionAssistantMessageParam::OrHash,
+                OpenAI::Chat::ChatCompletionToolMessageParam::OrHash,
+                OpenAI::Chat::ChatCompletionFunctionMessageParam::OrHash
+              )
+            ],
+            model: T.any(String, OpenAI::ChatModel::OrSymbol),
+            audio: T.nilable(OpenAI::Chat::ChatCompletionAudioParam::OrHash),
+            frequency_penalty: T.nilable(Float),
+            function_call: T.any(
+              OpenAI::Chat::CompletionCreateParams::FunctionCall::FunctionCallMode::OrSymbol,
+              OpenAI::Chat::ChatCompletionFunctionCallOption::OrHash
+            ),
+            functions: T::Array[OpenAI::Chat::CompletionCreateParams::Function::OrHash],
+            logit_bias: T.nilable(T::Hash[Symbol, Integer]),
+            logprobs: T.nilable(T::Boolean),
+            max_completion_tokens: T.nilable(Integer),
+            max_tokens: T.nilable(Integer),
+            metadata: T.nilable(T::Hash[Symbol, String]),
+            modalities: T.nilable(T::Array[OpenAI::Chat::CompletionCreateParams::Modality::OrSymbol]),
+            moderation: T.nilable(OpenAI::Chat::CompletionCreateParams::Moderation::OrHash),
+            n: T.nilable(Integer),
+            parallel_tool_calls: T::Boolean,
+            prediction: T.nilable(OpenAI::Chat::ChatCompletionPredictionContent::OrHash),
+            presence_penalty: T.nilable(Float),
+            prompt_cache_key: T.nilable(String),
+            prompt_cache_options: OpenAI::Chat::CompletionCreateParams::PromptCacheOptions::OrHash,
+            prompt_cache_retention: T.nilable(OpenAI::Chat::CompletionCreateParams::PromptCacheRetention::OrSymbol),
+            reasoning_effort: T.nilable(OpenAI::ReasoningEffort::OrSymbol),
+            response_format: T.any(
+              OpenAI::ResponseFormatText::OrHash,
+              OpenAI::ResponseFormatJSONSchema::OrHash,
+              OpenAI::ResponseFormatJSONObject::OrHash
+            ),
+            safety_identifier: T.nilable(String),
+            seed: T.nilable(Integer),
+            service_tier: T.nilable(OpenAI::Chat::CompletionCreateParams::ServiceTier::OrSymbol),
+            stop: T.nilable(OpenAI::Chat::CompletionCreateParams::Stop::Variants),
+            store: T.nilable(T::Boolean),
+            stream_options: T.nilable(OpenAI::Chat::ChatCompletionStreamOptions::OrHash),
+            temperature: T.nilable(Float),
+            tool_choice: T.any(
+              OpenAI::Chat::ChatCompletionToolChoiceOption::Auto::OrSymbol,
+              OpenAI::Chat::ChatCompletionAllowedToolChoice::OrHash,
+              OpenAI::Chat::ChatCompletionNamedToolChoice::OrHash,
+              OpenAI::Chat::ChatCompletionNamedToolChoiceCustom::OrHash
+            ),
+            tools: T::Array[
+              T.any(OpenAI::Chat::ChatCompletionFunctionTool::OrHash, OpenAI::Chat::ChatCompletionCustomTool::OrHash)
+            ],
+            top_logprobs: T.nilable(Integer),
+            top_p: T.nilable(Float),
+            user: String,
+            verbosity: T.nilable(OpenAI::Chat::CompletionCreateParams::Verbosity::OrSymbol),
+            web_search_options: OpenAI::Chat::CompletionCreateParams::WebSearchOptions::OrHash,
+            stream: T.noreturn,
+            request_options: OpenAI::RequestOptions::OrHash
+          )
+            .returns(OpenAI::Streaming::ChatCompletionStream)
+        end
+        def stream(
+          messages:,
+          model:,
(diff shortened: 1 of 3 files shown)


rbi/openai/helpers/streaming/chat_completion_stream.rbi
```diff
@@ -0,0 +1,164 @@
...
+
+        sig { returns(String) }
+        def get_output_text
+        end
+
+        sig { returns(T.self_type) }
+        def until_done
+        end
+
+        sig { returns(T.nilable(OpenAI::Chat::ChatCompletion)) }
```
markstuart-oai: [P2] Preserve the partial snapshot's nullable finish reason

`current_completion_snapshot` returns `ParsedChatCompletion`, whose `ParsedChoice` deliberately permits `finish_reason: nil` before a terminal chunk. Advertising it as `ChatCompletion` makes Sorbet infer the ordinary choice's non-null `FinishReason::TaggedSymbol`. I verified that `T.must(stream.current_completion_snapshot).choices.fetch(0).finish_reason.to_sym` type-checks, while the same access on a snapshot after a normal nonterminal chunk raises `NoMethodError` on nil. Please describe the partial completion/choice contract explicitly in the RBI, preserving nullable finish reasons, and add a consumer check plus a nonterminal-chunk runtime assertion; making only the outer snapshot nullable does not model this.

rbi/openai/helpers/streaming/events.rbi
```diff
@@ -154,8 +154,13 @@ module OpenAI
         def last_response
         end
 
-        sig { returns(T.untyped) }
-        def each
+        sig do
+          params(
+            blk: T.proc.params(event: ChatCompletionStream::ChatCompletionStreamEvent).void
```
markstuart-oai, line 211: [P2] Complete the event interface used by the typed callback

The callback now yields the concrete event union, but none of these chat event RBIs declares `type`, and the content/refusal logprob delta events omit `snapshot`. Those are real runtime fields documented in `helpers.md`. A consumer's `stream.each { |event| event.type }` now reports a missing method for all 11 alternatives; narrowing to `ChatLogprobsContentDeltaEvent` still rejects `event.snapshot`. Please add the missing field signatures alongside the typed callback and test these documented accesses (including both logprob snapshot variants), so consumers can use the event contract without casts or `T.unsafe`.

markstuart-oai (commented):
Reviewed the full diff and the runtime stream, event, parsed-model, and request-conversion contracts. The focused chat-stream suite passes (76 tests, 569 assertions), as does the shipped-RBI Sorbet check. Two additional contract gaps are detailed inline: partial snapshots are advertised with a non-null finish reason, and the newly typed event callbacks cannot access documented event fields. I also reproduced the existing converter finding at https://github.com/openai/openai-ruby/pull/634#discussion_r3941266595: both `response_format: MyModel` and `tools: [MyModel]` are supported at runtime but rejected by the new signature. Please cover these consumer workflows in the regression tests.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
```diff
@@ -0,0 +1,164 @@
...
+            presence_penalty: T.nilable(Float),
+            prompt_cache_key: T.nilable(String),
+            prompt_cache_options: OpenAI::Chat::CompletionCreateParams::PromptCacheOptions::OrHash,
+            prompt_cache_retention: T.nilable(OpenAI::Chat::CompletionCreateParams::PromptCacheRetention::OrSymbol),
+            reasoning_effort: T.nilable(OpenAI::ReasoningEffort::OrSymbol),
+            response_format: T.any(
+              OpenAI::ResponseFormatText::OrHash,
+              OpenAI::ResponseFormatJSONSchema::OrHash,
+              OpenAI::ResponseFormatJSONObject::OrHash
+            ),
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 094c7ff4: the stream RBI now accepts OpenAI::StructuredOutput::JsonSchemaConverter for both response_format and tools, matching create. The Sorbet consumer test covers MathResponse and MathTool.

rbi/openai/helpers/streaming/events.rbi
jbeckwith-oai: Fixed in 094c7ff4: all 11 concrete chat event alternatives now expose type, and both logprob delta events expose snapshot. Consumer checks cover event.type and both snapshot variants.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
jbeckwith-oai: Fixed in 094c7ff4: the RBI now models the runtime ParsedChatCompletion and ParsedChoice snapshot contract, including nullable finish_reason. Consumer typing and a nonterminal runtime assertion cover the partial state.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
```diff
@@ -0,0 +1,183 @@
...
+            ChatLogprobsContentDoneEvent,
+            ChatLogprobsRefusalDeltaEvent,
+            ChatLogprobsRefusalDoneEvent
+          )
+        end
+
+        Message = type_member { {fixed: ChatCompletionStreamEvent} }
+        Elem = type_member { {fixed: ChatCompletionStreamEvent} }
+
+        sig { returns(OpenAI::Chat::ChatCompletion) }
```
markstuart-oai: [P2] Preserve the parsed choice type in the final result too

The snapshot fix works, but `get_final_completion` still discards the same nullable contract. Using the new nonterminal `[DONE]` fixture through the public client, I verified that this method returns `ParsedChatCompletion` with a `ParsedChoice` whose `finish_reason` is nil. Sorbet accepts `stream.get_final_completion.choices.fetch(0).finish_reason.to_sym`, and that expression raises `NoMethodError` at runtime. Please return `OpenAI::Chat::ParsedChatCompletion` here as well, update the positive assertion, and add a negative consumer check for an unguarded final finish reason. This corroborates discussion_r3941437889; no runtime behavior change is needed.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
```diff
@@ -0,0 +1,183 @@
...
+        sig { returns(T::Array[OpenAI::Models::Chat::ParsedChoice]) }
+        attr_accessor :choices
+      end
+    end
+  end
+
+  module Helpers
+    module Streaming
+      class ChatCompletionStream
+        include OpenAI::Internal::Type::BaseStream
```
markstuart-oai, line 41: [P2] Override the inherited initializer with the actual stream constructor

Including `BaseStream` without declaring this class's initializer makes Sorbet use `initialize(model:, url:, response_metadata:, response:, unwrap:, stream:)`. I verified that `raw = client.chat.completions.stream_raw(...); OpenAI::Streaming::ChatCompletionStream.new(raw_stream: raw)` works at runtime, but Sorbet rejects `raw_stream:` and requests the six unrelated base keywords. Please declare `initialize(raw_stream:, response_format: nil, input_tools: nil)` here, following the existing ResponseStream RBI pattern, and cover construction from a typed raw stream. This independently confirms discussion_r3941437896.

markstuart-oai (commented):
Re-reviewed the full updated diff and runtime/type contracts. The earlier converter, event accessor, and direct snapshot findings are addressed, and the focused suite passes (77 tests, 571 assertions), along with the shipped-RBI Sorbet check. I independently reproduced the two remaining issues already identified by the bot: the final completion still hides nullable finish reasons, and the stream inherits an initializer signature that does not match its runtime constructor. The inline comments give concrete consumer reproductions and fixes.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
jbeckwith-oai: Fixed in 6e145216: ChatCompletionStream now declares the runtime initialize(raw_stream:, response_format: nil, input_tools: nil) signature. The consumer typecheck covers direct construction, so the inherited BaseStream keyword shape is no longer exposed.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 6e145216: get_final_completion now returns ParsedChatCompletion, and ParsedChoice preserves nullable finish_reason through its accessor and constructor while sparse to_hash remains broad. Consumer and runtime coverage exercise the nonterminal DONE case.

jbeckwith-oai: Fixed in 6e145216: get_final_completion now returns ParsedChatCompletion, and ParsedChoice preserves nullable finish_reason through its accessor and constructor while sparse to_hash remains broad. Consumer and runtime coverage exercise the nonterminal DONE case.

rbi/openai/helpers/streaming/chat_completion_stream.rbi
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 6e145216: ChatCompletionStream now declares the runtime initialize(raw_stream:, response_format: nil, input_tools: nil) signature. The consumer typecheck covers direct construction, so the inherited BaseStream keyword shape is no longer exposed.

markstuart-oai (approved):
Re-reviewed the full current diff and the incremental fixes against the runtime streaming and parsed-model contracts. The previous findings are addressed: final completions preserve nullable parsed-choice finish reasons, the concrete raw-stream constructor is declared, and converter inputs plus event accessors remain covered. I verified that construction from a typed raw stream now type-checks and the previously unsafe final finish-reason access is rejected. Parsed-choice construction and hash declarations also account for omitted finish reasons without changing runtime behavior. The focused suite passes (77 tests, 574 assertions), along with the shipped-RBI Sorbet check and additional positive/negative consumer probes. No remaining substantive correctness or maintainability findings. Validation was offline; I did not run the full SDK suite, live API tests, Steep, or full lint.

PR #635, opened 2026-09-05
fix: merge JSON-keyed extra body overrides

jbeckwith-oai:
## Problem

`request_options[:extra_body]` is documented as a deep merge into the generated request body, but the merge previously compared Ruby keys directly. A JSON-derived override uses string keys, while generated request bodies use symbols. When both represented the same JSON member, serialization emitted duplicate members and the later override could discard generated siblings after decode.

This is deterministic data loss in the supported request-building path for users who load overrides from JSON configuration. It is not a measured customer-incidence claim and does not rely on a live API call.

## User-facing example

```ruby
require "json"
require "openai"

client = OpenAI::Client.new
override = JSON.parse('{"text":{"format":{"strict":true}}}')

client.responses.create(
  model: "gpt-4o-mini",
  input: "Return JSON",
  text: {
    format: {
      type: :json_schema,
      name: "Answer",
      schema: {
        type: "object",
        properties: {answer: {type: "string"}},
        required: ["answer"],
        additionalProperties: false
      }
    }
  },
  request_options: {extra_body: override}
)
```

With safe synthetic transport capture, before this change the JSON-derived override encoded two `text` members and decoding retained only:

```ruby
{"format" => {"strict" => true}}
```

After this change it encodes one `text` member and preserves the generated schema siblings:

```ruby
{
  "format" => {
    "type" => "json_schema",
    "name" => "Answer",
    "schema" => {"type" => "object", ...},
    "strict" => true
  }
}
```

## Fix

At the existing request-body merge boundary in `BaseClient#build_request`, reconcile only exact string/symbol counterparts while merging hash layers. The body-side key is retained, overrides still win for scalar leaves, nested hashes still merge, arrays still replace, and case-sensitive names remain distinct.

The lookup uses per-level string/symbol indexes, so it stays linear without interning arbitrary JSON strings as symbols. Missing keys are checked before fetch, avoiding caller Hash default-proc execution. Leaf values are not traversed or consumed.

## Compatibility and scope

- No public API changes.
- No changes to shared `Internal::Util.deep_merge`, generated files, models, types, dependencies, providers, retries, redirects, authentication, headers, URLs, or response parsing.
- Caller-owned and frozen containers remain unmodified.
- Replayability remains determined from the effective merged body.

## Testing

- Added `test/openai/extra_body_merge_test.rb` covering:
  - generated public Responses calls with JSON-derived keys;
  - ordinary Hash and `OpenAI::RequestOptions` callers;
  - top-level and nested collisions in both key directions;
  - scalar override, array replacement, case-sensitive keys, frozen input immutability;
  - enumerable leaf non-consumption/replayability and Hash default-proc safety.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/extra_body_merge_test.rb` — 5 runs, 15 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/internal/transport/base_client_origin_test.rb` — 24 runs, 412 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/http_client_test.rb` — 36 runs, 158 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/client_options_test.rb` — 13 runs, 87 assertions, 0 failures.
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — RuboCop, directives, Sorbet, RBS validation all passed.
- Serialized full suite against the safe current-spec mock: `mise exec ruby@4.0.6 -- bundle exec rake test` — 1,713 runs, 14,923 assertions, 0 failures, 0 errors, 1 skip.
- Trusted current-main public custom-code budget check — passed: 3,365 / 4,000 lines.

## Review and security

Four paired fresh-context adversarial-review rounds were run; the final two consecutive rounds were clean. The review identified and fixed two in-scope issues before publication: avoiding quadratic counterpart scans and avoiding missing-key default-proc reads.

The bounded serialization/replayability security review found no symbol interning, leaf consumption, credential/logging change, or change to authentication, transport destinations, redirects, retries, or response parsing. Because this touches the request serialization boundary, `@openai/sdks-team` review is requested.

## Limitations

This intentionally handles only string/symbol equivalents during `extra_body` request-body merging. It does not define a general key-coercion policy or change unrelated duplicate-key behavior elsewhere in the SDK.


lib/openai/internal/transport/base_client.rb
```diff
@@ -494,6 +494,72 @@ def initialize(
           }
         end
 
+        # @api private
+        #
+        # Keep extra_body's deep-merge contract when one layer uses Ruby symbol
+        # keys and the other uses JSON-derived string keys for the same wire field.
+        # Do not intern caller-provided strings as symbols while finding matches.
+        private def merge_request_body(body, extra_body)
```
markstuart-oai: [P2] Extract the request-body merge policy from BaseClient

this pushes the file past 1k lines. can we decompose this first? BaseClient grows from 992 to 1,058 lines. These four new methods implement a self-contained, pure key-reconciliation policy, but live alongside URL validation, retries, redirects, encoding, and response handling. Keep the call at the existing build_request boundary and move the policy into a focused internal request-body merge module. Please also simplify the implementation there: after the exact-key check, a single string-name-to-original-key index can handle both string/symbol directions, so the two separate indexes and their indexing/matching helper pair are unnecessary. Preserve exact-key precedence, body-side key identity, top-level nil omission versus nested nil replacement, and the existing non-consuming leaf behavior. The wire-level tests should remain at the public request boundary.

markstuart-oai (commented):
The string/symbol reconciliation preserves generated siblings and override precedence in the serialized requests I checked, including nested nulls, array replacement, API-name aliases, frozen inputs, and non-consuming enumerable leaves. Focused request/transport tests passed (78 runs, 672 assertions), merge utility tests passed (10/10), and additional transport-capture probes passed; Sorbet, RBS validation, and RuboCop on the changed files also passed. I have one structural concern below: this independent merge policy adds four methods to BaseClient and pushes it past 1,000 lines. Please decompose that policy before approval. A broader utility-test run also encountered nine existing multipart FakeCGI fixture errors; those are not the basis for this review.

lib/openai/internal/transport/base_client.rb
jbeckwith-oai: Addressed in d402e99f. I extracted the pure request-body policy into `OpenAI::Internal::Transport::RequestBodyMerge`, leaving `BaseClient` with only the existing build-request delegation. The helper now uses one per-layer string-name index built from the original body-side keys, preserves exact-key precedence and body-key identity, and carries original subtree provenance through recursion so aliases introduced only by `extra_body` are not newly reconciled. I also added a public request-boundary regression for that nested case; focused tests and full lint/type/signature validation pass.

markstuart-oai (approved):
Re-reviewed d402e99f and the response to my earlier finding. The extraction addresses it: BaseClient is back to 992 lines, and the focused RequestBodyMerge module replaces four helpers and two directional indexes with two methods and one original-body key index. Carrying the original subtree through recursion is justified by the explicitly bounded alias policy and its request-boundary regression.

Fresh validation passed: 79 request/transport tests (677 assertions), 10 merge utility tests, 30 additional/inherited probe tests (99 assertions), Sorbet, 1,286 RBS files, and RuboCop on all four changed files. The wire probes cover generated siblings, aliases, null/array precedence, exact-key identity, frozen inputs, default-proc safety, and effective-body replayability. No remaining substantive findings.

Limits: I did not run the full SDK suite or full lint. The intentional override-only duplicate-key regression emits a JSON deprecation warning. The current OkTest report shows 233/237 passing; those CI failures still need attention before merge.

PR #638, opened 2026-09-05
fix: preserve complete interleaved tool arguments

jbeckwith-oai:
## Problem

When parallel Chat function calls stream their argument fragments in an interleaved order, the high-level stream could finalize a strict call as soon as another tool index appeared. That emitted an empty, missing, or partial `tool_calls.function.arguments.done` event even though later fragments completed the same call and the final completion contained the correct JSON.

## User-facing behavior

```ruby
  model: "gpt-4o-mini",
  messages: [{role: :user, content: "Look up both records"}],
  tools: [lookup_tool],
  parallel_tool_calls: true
)

stream.each do |event|
  next unless event.type == :"tool_calls.function.arguments.done"

  puts [event.index, event.arguments, event.parsed].inspect
end
```

For a synthetic two-call stream with simultaneous headers and alternating argument fragments:

```
before: [[0, "", nil], [1, "", nil]]
after:  [[0, "{\"x\":1}", {x: 1}], [1, "{\"x\":2}", {x: 2}]]
```

The final completion was already correct before and remains correct after.

## Fix

The per-choice event state now keeps a bounded set of observed pending tool indices. On an index transition, it emits early exactly as before unless a known strict call has missing arguments or is still unparsed in the already-hydrated snapshot. Deferred calls are flushed once at the terminal choice boundary, preserving terminal incomplete, non-strict, and unknown-call events with `parsed: nil`.

This reuses existing strict argument hydration rather than adding a second JSON parse, so ordinary deltas do not rescan historical calls. The existing done set still enforces exactly-once completion.

## Scope and compatibility

- Changed only `ChoiceEventState` completion bookkeeping in `lib/openai/helpers/streaming/chat_completion_stream.rb`.
- Added one public-stream regression file, `test/openai/helpers/chat_stream_interleaved_tool_done_test.rb`.
- Preserves sequential early done events, terminal/single-call behavior, final completion assembly, delta contents, choice isolation, finish errors, and raw-chunk handling.
- No public API, parser tolerance, tool execution, generated code, dependency, transport, logging, auth, schema, budget, or provenance changes.

## Verification

- Reproduced on base `b68ce906106cdaa165cbb6ebdd5ec34b2cc7ba2b`: sequential done values were complete/parsed; interleaved done values were empty/unparsed while final completion was complete.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/helpers/chat_stream_interleaved_tool_done_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/helpers/chat_stream_terminal_tool_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/helpers/chat_stream_initial_tool_parsing_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/helpers/chat_stream_late_choice_tool_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/helpers/chat_completion_parser_test.rb`
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/resources/chat/completions/streaming_test.rb`
- Serialized full suite: `TEST_API_BASE_URL=http://127.0.0.1:49137 mise exec ruby@4.0.6 -- bundle exec rake test` — 1,714 runs, 14,936 assertions, 0 failures, 0 errors, 1 skip.

## Review and security

Eight paired fresh-context adversarial rounds were run. Review found and removed repeated historical rescans, preserved syntactically complete `null` timing, and kept only unresolved pending calls. Rounds 7 and 8 were consecutive clean rounds on unchanged final code.

After review feedback, two additional paired fresh-context rounds were clean for the narrow omitted-arguments fix.

The change touches parsed tool-argument completion timing but adds no parser, evaluator, transport, logging, or credential behavior. Synthetic data only; no live API requests or tool execution.

## Reach and limitations

This deterministically proves the public SDK path for the synthetic interleaved ordering. It does not measure customer incidence or claim that a production server currently emits this exact ordering.

### DIFF
--- a/lib/openai/helpers/streaming/chat_completion_stream.rb
+++ b/lib/openai/helpers/streaming/chat_completion_stream.rb
@@ -656,6 +656,8 @@ module OpenAI
           @logprobs_content_done = false
           @logprobs_refusal_done = false
           @done_tool_calls = Set.new
+          @pending_tool_calls = Set.new
+          @deferred_tool_calls = Set.new
           @current_tool_call_index = nil
         end
 
@@ -671,19 +673,27 @@ module OpenAI
               events.concat(content_done_events(choice_snapshot, response_format))
 
               if @current_tool_call_index
-                event = tool_done_event(tool_calls_by_index, @current_tool_call_index)
-                events << event if event
+                if !@deferred_tool_calls.include?(@current_tool_call_index) &&
+                    incomplete_strict_tool_call?(tool_calls_by_index, @current_tool_call_index)
+                  @deferred_tool_calls.add(@current_tool_call_index)
+                end
+
+                unless @deferred_tool_calls.include?(@current_tool_call_index)
+                  event = tool_done_event(tool_calls_by_index, @current_tool_call_index)
+                  events << event if event
+                end
               end
             end
 
+            @pending_tool_calls.add(tool_call.index) unless @done_tool_calls.include?(tool_call.index)
             @current_tool_call_index = tool_call.index
           end
 
-          if choice_snapshot.finish_reason &&
-              @current_tool_call_index &&
-              !@done_tool_calls.include?(@current_tool_call_index)
-            event = tool_done_event(tool_calls_by_index, @current_tool_call_index)
-            events << event if event
+          if choice_snapshot.finish_reason
+            @pending_tool_calls.to_a.each do |tool_index|
+              event = tool_done_event(tool_calls_by_index, tool_index)
+              events << event if event
+            end
           end
 
           events
@@ -749,6 +759,8 @@ module OpenAI
           return nil if @done_tool_calls.include?(tool_index)
 
           @done_tool_calls.add(tool_index)
+          @pending_tool_calls.delete(tool_index)
+          @deferred_tool_calls.delete(tool_index)
 
           tool_call = tool_calls_by_index[tool_index]
           return nil unless tool_call&.type == :function
@@ -766,6 +778,22 @@ module OpenAI
             arguments: tool_call.function.arguments,
             parsed: parsed_args
           )
+        end
+
+        def incomplete_strict_tool_call?(tool_calls_by_index, tool_index)
+          tool_call = tool_calls_by_index[tool_index]
+          return false unless tool_call&.type == :function
+
+          function = tool_call.function
+          tool = find_input_tool(function.name)
+          return false unless tool&.dig(:function, :strict)
+          return false unless function.respond_to?(:parsed) && function.parsed.nil?
+          return true unless function.arguments
+
+          JSON.parse(function.arguments)
+          false
+        rescue JSON::ParserError
+          true
         end
 
         def parse_content(message, response_format)
--- a/test/openai/helpers/chat_stream_interleaved_tool_done_test.rb
+++ b/test/openai/helpers/chat_stream_interleaved_tool_done_test.rb
@@ -1,0 +1,182 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
+class OpenAI::Test::ChatStreamInterleavedToolDoneTest < Minitest::Test
+  extend Minitest::Serial
+  include WebMock::API
+
+  def before_all
+    super
+    WebMock.enable!
+  end
+
+  def after_all
+    WebMock.disable!
+    super
+  end
+
+  def setup
+    super
+    @client = OpenAI::Client.new(base_url: "http://localhost", api_key: "test-key")
+  end
+
+  def teardown
+    WebMock.reset!
+    super
+  end
+
+  def test_interleaved_strict_tool_calls_wait_for_complete_arguments
+    sequential = stream_result(
+      [
+        [header(0)],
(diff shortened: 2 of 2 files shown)


lib/openai/helpers/streaming/chat_completion_stream.rb
```diff
@@ -768,6 +780,22 @@ def tool_done_event(tool_calls_by_index, tool_index)
...
+          tool_call = tool_calls_by_index[tool_index]
+          return false unless tool_call&.type == :function
+
+          function = tool_call.function
+          tool = find_input_tool(function.name)
+          return false unless tool&.dig(:function, :strict)
+          return false unless function.respond_to?(:parsed) && function.parsed.nil?
+          return false unless function.arguments
+
+          JSON.parse(function.arguments)
```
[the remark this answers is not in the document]
jbeckwith-oai: I rechecked this on the exact current head ed5b4d4f with a deterministic public `client.chat.completions.stream` reproduction using two strict calls, simultaneous empty-string headers, and 20 alternating fragment updates. The current guard at the transition checks `@deferred_tool_calls` before calling `incomplete_strict_tool_call?`; once each call is observed incomplete, it is sticky-deferred until terminal flush. The measured counts were `snapshot_hydration: 24` (the existing `parse_tool_calls!` work) and `done_boundary_probe: 2` total, with complete done arguments for both calls. So the done boundary adds one probe per call, not one full-buffer parse per switch. I am leaving this thread unresolved for reviewer agreement; no code change is warranted for the current head.

lib/openai/helpers/streaming/chat_completion_stream.rb
```diff
@@ -768,6 +780,22 @@ def tool_done_event(tool_calls_by_index, tool_index)
...
 
+        def incomplete_strict_tool_call?(tool_calls_by_index, tool_index)
+          tool_call = tool_calls_by_index[tool_index]
+          return false unless tool_call&.type == :function
+
+          function = tool_call.function
+          tool = find_input_tool(function.name)
+          return false unless tool&.dig(:function, :strict)
+          return false unless function.respond_to?(:parsed) && function.parsed.nil?
+          return false unless function.arguments
```
markstuart-oai: [P2] Keep strict calls pending when the header omits arguments

`ChatCompletionChunk::Choice::Delta::ToolCall::Function.arguments` is optional, so a known strict call can arrive with just its name before its first argument fragment. This guard treats that state as complete. With headers for indices 0 and 1 in one chunk, omitting index 0's `arguments`, followed by `{"x":1}` and `{"x":2}` for the respective calls, index 0 emits done with nil arguments and enters `@done_tool_calls`. Later fragments produce a correct final completion but never repair the done event; reading `event.arguments` raises `OpenAI::Errors::ConversionError`. I reproduced this through `client.chat.completions.stream`. Please classify missing arguments as incomplete for known strict calls, preserve the terminal flush, and add the omitted-field variant to the public-stream regression.

markstuart-oai (commented):
Reviewed the complete diff and the accumulation, parsing, and done-event lifecycle. The added regressions and seven related test files pass locally (56 tests, 444 assertions). Additional synthetic public-stream probes confirm sparse indices, choice isolation, and repeated terminal handling, but a strict tool header with an omitted arguments field still emits an unusable done event before its arguments arrive; details inline. I also checked the existing performance suggestion: the deferred set prevents reparsing on every subsequent switch, so that claimed quadratic regression did not reproduce. Validation was local and synthetic; I did not run live API tests, the full suite, or lint/typechecks.

lib/openai/helpers/streaming/chat_completion_stream.rb
jbeckwith-oai: Fixed in 5d8ad435. For a known strict function, missing `function.arguments` is now classified as incomplete at the ChoiceEventState boundary, so an interleaved header cannot emit a nil done event before later fragments arrive. The call stays sticky-deferred and terminal flush emits the complete arguments exactly once. I added the public-stream omitted-header regression while preserving the existing syntactically complete `null` early-done case, terminal incomplete/non-strict/unknown behavior, and the one-probe bound. Focused stream tests, `rake lint`, the serialized full suite (1714 runs, 14936 assertions, 0 failures/errors, 1 skip), and two clean fresh review rounds passed.

markstuart-oai (approved):
Re-reviewed the complete diff at 5d8ad435db5bbd69a66d28302a8362c983e9364b. The omitted-arguments issue from my previous review is fixed: known strict calls remain pending until their arguments arrive, and the terminal flush retains exactly-once completion. The bookkeeping stays within ChoiceEventState, preserves sequential early completion and complete JSON null timing, and does not add historical-call rescans on ordinary deltas. I found no remaining substantive correctness or maintainability issues.

Validation: six focused streaming/parser test files passed (44 tests, 351 assertions). Restoring the previous guard reproduces the reported ConversionError in the new regression; restoring the fix passes it. Diff whitespace checks pass. I did not run the full suite, live API tests, or lint/typechecks.

PR #639, opened 2026-09-05
fix: populate nested API error metadata

jbeckwith-oai:
## Problem

Standard HTTP JSON API failures use an `error` envelope, but `OpenAI::Errors::APIStatusError#code`, `#type`, and `#param` only read top-level fields. A synthetic 429 Responses request therefore raised the right `RateLimitError` and preserved the raw body/request ID while leaving classification metadata unavailable.

## User-facing behavior

Applications can now classify standard nested API errors directly:

```ruby
begin
  client.responses.create(model: "gpt-4o-mini", input: "Hello")
rescue OpenAI::Errors::APIStatusError => error
  case error.code
  when "insufficient_quota"
    # Prompt the user to check quota or billing.
  when "rate_limit_exceeded"
    # Back off and retry later.
  end
end
```

Synthetic public-path result for the same 429 body:

```ruby
# body: {error: {code: "insufficient_quota", type: "insufficient_quota", param: "model"}}
# before: [error.code, error.type, error.param] # => [nil, nil, nil]
# after:  [error.code, error.type, error.param] # => ["insufficient_quota", "insufficient_quota", "model"]
```

## Fix

`APIStatusError` now keeps explicitly present top-level metadata authoritative and falls back to `body[:error]` only when each corresponding top-level key is absent. The existing converter still handles the selected scalar value, and malformed/non-Hash envelopes remain safe through the existing lookup semantics.

The new public HTTP-boundary regression covers:

- a synthetic 429 quota/rate-limit envelope;
- a concise synthetic 400 bad-parameter envelope;
- top-level precedence, explicit null/scalar behavior, and malformed envelope safety.

## Compatibility and scope

The raw outer `body`, exception subclass, status, URL, headers, request ID, cause, message formatting, redaction, and logging behavior are unchanged. Flat/provider bodies continue to work. This is limited to normal HTTP JSON responses; it does not parse raw strings or change SSE, OAuth, provider, transport, retry, generated model, RBI/RBS, dependency, or architecture behavior.

Reach confidence: deterministic supported public-path behavior is proven with synthetic data. This does not claim measured customer incidence or live API observations.

## Verification

- Red before fix: `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/api_error_metadata_test.rb` failed the two nested metadata assertions.
- Green after fix: `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/api_error_metadata_test.rb` — 3 runs, 20 assertions.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/errors_test.rb` — 36 runs, 3,196 assertions.
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/logging_security_test.rb` — 28 runs, 849 assertions.
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — clean (2,910 files, no RuboCop offenses; RBS validation clean).
- `mise exec ruby@4.0.6 -- bundle exec rake typecheck` — clean.
- Serialized `TEST_API_BASE_URL=http://127.0.0.1:49137 mise exec ruby@4.0.6 -- bundle exec rake test` — 1,714 runs, 14,954 assertions, 0 failures, 0 errors, 1 skip.
- Trusted current-main public custom-code budget on committed candidate `419fa8d96949268131def631efd3525bd9216adc` — success, 3,365 / 4,000 lines.
- Adversarial review: two consecutive clean rounds, each with two fresh independent read-only reviewers; no in-scope findings.
- Scoped response-deserialization/error-privacy review: no new sink or disclosure path; existing error and logging security tests remain green.

## Limitations

No live API call was made; all evidence is deterministic and synthetic.

### DIFF
--- a/lib/openai/errors.rb
+++ b/lib/openai/errors.rb
@@ -264,9 +264,9 @@ module OpenAI
           safe_status_message(url: url, status: status, body: body)
         end
 
-        @code = OpenAI::Internal::Type::Converter.coerce(String, OpenAI::Internal::Util.dig(body, :code))
-        @param = OpenAI::Internal::Type::Converter.coerce(String, OpenAI::Internal::Util.dig(body, :param))
-        @type = OpenAI::Internal::Type::Converter.coerce(String, OpenAI::Internal::Util.dig(body, :type))
+        @code = OpenAI::Internal::Type::Converter.coerce(String, error_metadata(body, :code))
+        @param = OpenAI::Internal::Type::Converter.coerce(String, error_metadata(body, :param))
+        @type = OpenAI::Internal::Type::Converter.coerce(String, error_metadata(body, :type))
         super(
           url: url,
           status: status,
@@ -276,6 +276,14 @@ module OpenAI
           response: response,
           message: message&.to_s
         )
+      end
+
+      private def error_metadata(body, key)
+        missing = Object.new
+        top_level = OpenAI::Internal::Util.dig(body, key) { missing }
+        return top_level unless top_level.equal?(missing)
+
+        OpenAI::Internal::Util.dig(body, [:error, key])
       end
 
       private def safe_status_message(url:, status:, body:)
--- a/lib/openai/errors.rb
+++ b/lib/openai/errors.rb
@@ -279,11 +279,7 @@ module OpenAI
       end
 
       private def error_metadata(body, key)
-        missing = Object.new
-        top_level = OpenAI::Internal::Util.dig(body, key) { missing }
-        return top_level unless top_level.equal?(missing)
-
-        OpenAI::Internal::Util.dig(body, [:error, key])
+        OpenAI::Internal::Util.dig(body, key) { OpenAI::Internal::Util.dig(body, [:error, key]) }
       end
 
       private def safe_status_message(url:, status:, body:)
--- a/test/openai/api_error_metadata_test.rb
+++ b/test/openai/api_error_metadata_test.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require_relative "test_helper"
+
+class OpenAI::Test::APIErrorMetadataTest < Minitest::Test
+  def test_standard_rate_limit_envelope_populates_metadata
+    body = {
+      error: {
+        message: "Synthetic quota error",
+        code: "insufficient_quota",
+        type: "insufficient_quota",
+        param: "model"
+      }
+    }
+
+    error = response_error(status: 429, body: body)
+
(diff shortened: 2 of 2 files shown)


lib/openai/errors.rb
```diff
@@ -278,6 +278,14 @@ def initialize(url:, status:, headers:, body:, request:, response:, message: nil
         )
       end
 
+      private def error_metadata(body, key)
+        missing = Object.new
+        top_level = OpenAI::Internal::Util.dig(body, key) { missing }
+        return top_level unless top_level.equal?(missing)
```
markstuart-oai: [P3] Use the existing missing-value fallback directly

`Util.dig` already invokes its block only when lookup fails, preserving explicitly present `nil` and `false`. Can we replace the sentinel/identity branch with `OpenAI::Internal::Util.dig(body, key) { OpenAI::Internal::Util.dig(body, [:error, key]) }`? This keeps the metadata policy in one helper while deleting the extra object and missing-value protocol. I verified equivalent results through the public HTTP path for absent, null, scalar, malformed, and mixed top-level/nested payloads.

markstuart-oai (approved):
Reviewed the full diff and traced the HTTP decoding → APIStatusError construction path, metadata declarations, and streaming/OAuth callers. The nested fallback preserves explicit top-level values and the raw response/status/header path. Independently verified 67 focused test runs (4,065 assertions), the new regression failing against the base implementation, and 102 synthetic public-path cases covering mixed, null, scalar, and malformed shapes. One small simplification below; no substantive correctness or architectural blockers found. No live API testing. Validation limit: the broader utility test file had nine FakeCGI errors in multipart tests; these are not review findings.

lib/openai/errors.rb
jbeckwith-oai: Replaced the sentinel/identity branch with the existing missing-only `Util.dig` block fallback. I also verified the public HTTP path for nested, mixed top-level/nested, explicit `nil`/`false`, absent, malformed, and scalar bodies; the focused regression now keeps absent and scalar cases covered.

markstuart-oai (approved):
Re-reviewed the complete diff and the update at cad89043. The canonical dig fallback now directly expresses absent-key fallback without the per-call sentinel, while preserving authoritative top-level nil/false values. Nested metadata stays in the existing error boundary and retains the established coercion and malformed-body behavior. The rebase and additional mixed-envelope regression are also reviewed. The prior simplification comment is addressed; no substantive findings.

Validation: the new public-request metadata tests and existing error tests passed on Ruby 4.0.6 (39 tests, 3,228 assertions); diff whitespace checks passed. I did not run the full suite, lint/typechecks, or live API calls.

PR #643, opened 2026-09-05
fix: redact malformed streamed JSON errors

jbeckwith-oai:
## Summary
- remove raw structured model output and parser text from malformed streaming JSON errors
- suppress the rescued parser exception cause to prevent cause-chain disclosure
- add a focused sensitive-marker regression for the public structured streaming path

## Validation
- GEM_HOME=/tmp/openai-ruby-min GEM_PATH=/tmp/openai-ruby-min:/opt/homebrew/lib/ruby/gems/3.2.0 ruby -Itest test/openai/resources/responses/streaming_test.rb --name test_structured_output_streaming_with_malformed_json
- GEM_HOME=/tmp/openai-ruby-min GEM_PATH=/tmp/openai-ruby-min:/opt/homebrew/lib/ruby/gems/3.2.0 ruby -Itest test/openai/resources/responses/streaming_test.rb --name test_structured_output_streaming
- GEM_HOME=/tmp/openai-ruby-min GEM_PATH=/tmp/openai-ruby-min:/opt/homebrew/lib/ruby/gems/3.2.0 ruby -Itest test/openai/resources/responses/streaming_test.rb
- GEM_HOME=/tmp/openai-ruby-min GEM_PATH=/tmp/openai-ruby-min:/opt/homebrew/lib/ruby/gems/3.2.0 ruby -S rubocop --force-exclusion lib/openai/helpers/streaming/response_stream.rb test/openai/resources/responses/streaming_test.rb
- ./scripts/rubyfmt --check -- lib/openai/helpers/streaming/response_stream.rb test/openai/resources/responses/streaming_test.rb
- ruby -c for both changed files
- git diff --check

## Security review
The malformed response text is network/model-derived and can contain customer content. The fix preserves RuntimeError and successful structured parsing while ensuring the propagated error message and cause chain contain no model output or parser excerpt. Two required adversarial-review rounds completed cleanly before push.

### DIFF
--- a/lib/openai/helpers/streaming/response_stream.rb
+++ b/lib/openai/helpers/streaming/response_stream.rb
@@ -267,10 +267,10 @@ module OpenAI
           begin
             parsed = JSON.parse(text, symbolize_names: true)
             OpenAI::Internal::Type::Converter.coerce(@text_format, parsed)
-          rescue JSON::ParserError => e
+          rescue JSON::ParserError
             raise(
-              "Failed to parse structured text as JSON for #{@text_format}: #{e.message}. " \
-                "Raw text: #{text.inspect}"
+              "Failed to parse structured text as JSON for #{@text_format}",
+              cause: nil
             )
           end
         end
--- a/test/openai/resources/responses/streaming_test.rb
+++ b/test/openai/resources/responses/streaming_test.rb
@@ -276,7 +276,9 @@ class OpenAI::Test::Resources::Responses::StreamingTest < Minitest::Test
     end
 
     assert_match(/Failed to parse structured text as JSON/, error.message)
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed the full two-file diff and the streaming/parser callers. The direct raise with cause: nil removes payload-bearing diagnostics while preserving RuntimeError, the existing rescue boundary, and successful structured parsing; no extra abstraction is needed here.

Validated on Ruby 4.0.6: all 33 streaming tests pass (126 assertions). Additional offline probes pass (334 assertions) across malformed payloads, normal/resumed streams, message/inspect/full_message, nested causes, and unrelated exception propagation. The probes detect the old implementation's behavior. No substantive findings.

PR #644, opened 2026-09-05
fix: suppress Responses WebSocket exception causes

jbeckwith-oai:
## Summary
- explicitly apply each WebSocket product error cause policy at shared raise boundaries
- prevent Responses errors from retaining suppressed implicit Ruby causes in full_message
- preserve Realtime cause propagation

## Validation
- direct trigger: Responses cause nil and full_message leak false
- Responses WebSocket errors: 2 runs, 11 assertions
- Responses WebSocket connection: 47 runs, 213 assertions
- Realtime async WebSocket transport: 13 runs, 63 assertions
- rake lint:rubocop: 2,910 files, no offenses
- security diff review: complete, no findings
- adversarial review: two consecutive clean rounds, four fresh-context reviewers

## Note
- Full rake test reached 1,715 runs / 14,947 assertions and failed only on the unrelated existing Safety Alerts fixture 404; the same failure reproduced in isolation.

### DIFF
--- a/lib/openai/helpers/websocket/async_websocket_transport.rb
+++ b/lib/openai/helpers/websocket/async_websocket_transport.rb
@@ -170,20 +170,23 @@ module OpenAI
         def read
           @connection.read
         rescue StandardError => e
-          raise @error_factory.call(url: @url, cause: e)
+          error = @error_factory.call(url: @url, cause: e)
+          raise error, cause: error.cause
         end
 
         def write(message)
           @connection.write(message)
           @connection.flush
         rescue StandardError => e
-          raise @error_factory.call(url: @url, cause: e)
+          error = @error_factory.call(url: @url, cause: e)
+          raise error, cause: error.cause
         end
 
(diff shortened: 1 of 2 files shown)


markstuart-oai (approved):
Reviewed the six shared raise boundaries, product error factories, and surrounding parse/lifecycle paths. Explicitly passing the error's cause follows the existing shared request pattern and keeps product policy in the error abstraction; no extra product branches or wrapper are needed. No substantive correctness or maintainability findings.

Validated on Ruby 4.0.6: 98 focused tests / 529 assertions passed, plus synthetic boundary and public-path probes (2 tests / 100 assertions) covering read, write/flush, close, abort, opening, cleanup, missing dependencies, Responses parsing, native Ruby causes, nested rendering, and Realtime cause identity. The same probe detects the retained implicit cause with the merge-base transport. Full-suite and full-lint claims were not independently reproduced; no live API calls were used.

PR #645, opened 2026-09-05
fix: preserve unknown discriminated union values

jbeckwith-oai:
## Summary
- preserve explicit unregistered discriminators as raw values instead of structurally coercing them into known union variants
- retain existing known-discriminator dispatch and missing-discriminator shorthand matching
- add focused regression coverage for symbol/string discriminator forms, the reported Responses tool-call misclassification, and unknown stream-event serialization

## Validation
- focused tests: 27 runs, 206 assertions, 0 failures
- compatibility probe: unknown tags stay Hash; known computer_call stays ResponseComputerToolCall; absent request tag keeps shorthand conversion
- bundle exec rake lint: passed (RuboCop, directive validation, Sorbet, RBS)
- scoped Codex Security diff review: complete, 0 findings
- adversarial review: two consecutive clean rounds, two fresh independent reviewers per round

## Full suite note
- bundle exec rake test reached 1,716 runs / 14,961 assertions with one unrelated existing fixture-server 404 in OpenAI::Test::Resources::Safety::AlertsTest#test_retrieve; that failure reproduces when run alone.

### DIFF
--- a/lib/openai/internal/type/union.rb
+++ b/lib/openai/internal/type/union.rb
@@ -100,8 +100,25 @@ module OpenAI
             return nil if key == OpenAI::Internal::OMIT
 
             key = key.to_sym if key.is_a?(String)
-            _, found = known_variants.find { |k,| k == key }
-            found&.call
+            _, found = known_variants.find { |k,| !k.nil? && k == key }
+            return found.call if found
+
+            known_variants.each do |variant_key, variant_fn|
+              next unless variant_key.nil?
+
+              target = variant_fn.call
+              next unless target.is_a?(Class) && target <= OpenAI::Internal::Type::BaseModel
+
+              field = target.known_fields[@discriminator]
+              next unless field
+
+              type = field.fetch(:type_fn).call
+              next unless OpenAI::Internal::Type::Enum === type
+
+              return target if type === key
+            end
+
+            OpenAI::Internal::Type::Unknown
           else
             nil
           end
--- a/lib/openai/helpers/streaming/chat_completion_stream.rb
+++ b/lib/openai/helpers/streaming/chat_completion_stream.rb
@@ -265,7 +265,10 @@ module OpenAI
 
           choice.delta.tool_calls.each do |tool_call_delta|
             tool_call = tool_calls_by_index[tool_call_delta.index]
-            next unless tool_call&.type == :function && tool_call_delta.function
+            unless tool_call.is_a?(OpenAI::Chat::ChatCompletionMessageFunctionToolCall) &&
+                tool_call_delta.function
+              next
+            end
 
             parsed_args = if tool_call.function.respond_to?(:parsed)
               tool_call.function.parsed
@@ -326,7 +329,7 @@ module OpenAI
 
           delta_tool_calls.each do |tool_call_chunk|
             tool_call_snapshot = tool_calls_by_index[tool_call_chunk.index]
-            next unless tool_call_snapshot&.type == :function
+            next unless tool_call_snapshot.is_a?(OpenAI::Chat::ChatCompletionMessageFunctionToolCall)
 
             input_tool = find_input_tool(tool_call_snapshot.function.name)
             next unless input_tool&.dig(:function, :strict)
@@ -751,7 +754,7 @@ module OpenAI
           @done_tool_calls.add(tool_index)
 
           tool_call = tool_calls_by_index[tool_index]
-          return nil unless tool_call&.type == :function
+          return nil unless tool_call.is_a?(OpenAI::Chat::ChatCompletionMessageFunctionToolCall)
 
           parsed_args = parse_function_tool_arguments(tool_call.function)
 
--- a/lib/openai/helpers/streaming/response_stream.rb
+++ b/lib/openai/helpers/streaming/response_stream.rb
@@ -46,10 +46,11 @@ module OpenAI
           text_parts = []
 
           response.output.each do |output|
-            next unless output.type == :message
+            next unless output.is_a?(OpenAI::Models::Responses::ResponseOutputMessage)
 
             output.content.each do |content|
-              next unless content.type == :output_text
+              next unless content.is_a?(OpenAI::Models::Responses::ResponseOutputText)
+
               text_parts << content.text
             end
           end
@@ -100,39 +101,27 @@ module OpenAI
 
           case event
           when OpenAI::Models::Responses::ResponseTextDeltaEvent
+            snapshot = nil
             if @current_snapshot
               output = @current_snapshot.output[event.output_index]
-              assert_type(output, :message)
-
-              content = output.content[event.content_index]
-              assert_type(content, :output_text)
-              events_to_yield <<
-                OpenAI::Streaming::ResponseTextDeltaEvent.new(
-                  content_index: event.content_index,
-                  delta: event.delta,
-                  item_id: event.item_id,
-                  output_index: event.output_index,
-                  sequence_number: event.sequence_number,
-                  type: event.type,
-                  **event.to_h.slice(:logprobs),
-                  snapshot: content.text
-                )
-            else
-              # A server-directed resumed stream may begin after response.created.
-              # Without the omitted prefix, a snapshot would be incomplete and
-              # materializing every partial prefix would make streaming quadratic.
-              events_to_yield <<
-                OpenAI::Streaming::ResponseTextDeltaEvent.new(event.to_h.merge(snapshot: nil))
+              if output.is_a?(OpenAI::Models::Responses::ResponseOutputMessage)
+                content = output.content[event.content_index]
+                snapshot = content.text if content.is_a?(OpenAI::Models::Responses::ResponseOutputText)
+              end
             end
+
+            # A server-directed resumed stream or an unknown future snapshot value
+            # has no complete prefix from which to build a truthful snapshot.
+            events_to_yield <<
+              OpenAI::Streaming::ResponseTextDeltaEvent.new(event.to_h.merge(snapshot: snapshot))
 
           when OpenAI::Models::Responses::ResponseTextDoneEvent
             text = if @current_snapshot
               output = @current_snapshot.output[event.output_index]
-              assert_type(output, :message)
-
-              content = output.content[event.content_index]
-              assert_type(content, :output_text)
-              content.text
+              if output.is_a?(OpenAI::Models::Responses::ResponseOutputMessage)
+                content = output.content[event.content_index]
+                content.text if content.is_a?(OpenAI::Models::Responses::ResponseOutputText)
+              end
             else
               event.text
             end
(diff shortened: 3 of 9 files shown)


lib/openai/internal/type/union.rb
```diff
@@ -101,7 +101,7 @@ def variants = derefed_variants.map { _2 }
 
             key = key.to_sym if key.is_a?(String)
             _, found = known_variants.find { |k,| k == key }
-            found&.call
+            found ? found.call : OpenAI::Internal::Type::Unknown
```
markstuart-oai: [P2] Update text extraction consumers for the newly raw variants

Returning Unknown here exposes a caller assumption in `Response#output_text` and `ResponseStream#get_output_text`: both call `.type` on every output item and content block. With a synthetic response containing `output: [{type: "future_tool"}, <a valid message containing "hello">]`, both `client.responses.create(...).output_text` and `client.responses.stream(...).get_output_text` now raise `NoMethodError: undefined method 'type' for an instance of Hash`. An unknown content block preceding a valid `output_text` block causes the same failure. I reproduced all four through WebMock; replacing only this union implementation with the merge-base version returns `"hello"` in each case.

Please keep the raw-preservation behavior and update both text extractors to select known message/text models before accessing model attributes, with public-entrypoint regression tests covering unknown output items and nested content blocks. Otherwise a future API variant makes existing text extraction crash even when the response includes valid text.

markstuart-oai (commented):
Two correctness issues need addressing before approval.

1. Confirmed the existing inline report at https://github.com/openai/openai-ruby/pull/645#discussion_r3941994954: known web-search tags are represented by unkeyed variants whose type fields are enums. All four supported tags (web_search, web_search_2025_08_26, web_search_preview, web_search_preview_2025_03_11) now coerce to Hash in both Responses::Tool and Beta::BetaTool; the merge-base implementation returns the corresponding WebSearchTool/WebSearchPreviewTool models in all eight cases. Explicitly keyed known variants retain dispatch, but these supported tags do not. Recognition must account for the canonical enum-valued discriminator representation before treating a value as unknown; blindly restoring structural fallback for arbitrary tags would undo the intended fix.

2. My inline comment documents a separate public-caller regression: Response#output_text and ResponseStream#get_output_text call model methods on the newly raw output/content hashes. Four WebMock public-entrypoint cases raise at head and return the known text with the base union implementation.

Using the existing Unknown converter at the canonical dispatch point is a direct design, with no unnecessary wrapper or file-size growth. However, it currently equates an absent keyed registration with an unknown discriminator, which misses the enum-valued variants. Keep the central invariant while recognizing all declared discriminator values and updating downstream text consumers.

Validation: local type tests passed (67 tests, 567 assertions), selected stream tests passed (18 tests, 174 assertions), and 856 truly unknown symbol/string cases across all 214 discriminated unions preserve raw identity and serialization without structural fallback. These checks do not establish that all known tags preserve behavior; the eight web-search cases above demonstrate the gap. CI showed 18 successful checks and two skipped checks. Full suite/lint and the author's adversarial-review claims were not independently rerun.

lib/openai/internal/type/union.rb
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 3406d81d. Explicit unregistered tags now first probe only unkeyed BaseModel discriminator fields in isolated state; a recognized multi-tag web-search discriminator then gets the normal full-model coercion path, while a genuinely unknown tag still remains raw. Added stable and Beta regressions for all web-search tag variants, including unrelated inexact fields.

jbeckwith-oai: Fixed in 3406d81d. Response#output_text and ResponseStream#get_output_text now select concrete known message/text models before attribute access, and streaming accumulation similarly guards message, text, and function-call mutations. Added public create/stream regressions with unknown top-level output items and nested content preceding valid output text.

lib/openai/internal/type/union.rb
```diff
@@ -101,12 +104,49 @@ def variants = derefed_variants.map { _2 }
...
+        # An explicit discriminator can still belong to an unkeyed variant whose model
+        # accepts several discriminator values. Match only that field before coercing the
+        # whole model so unrelated inexact fields do not affect discriminator dispatch.
+        #
+        # @param value [Object]
+        #
+        # @param state [Hash{Symbol=>Object}]
+        #
+        # @return [Object]
+        private def coerce_unknown_discriminator(value, state:)
```
markstuart-oai: [P2] Resolve unkeyed tags in the canonical resolver instead of adding a second dispatch path

This fixes the web-search cases, but adds an avoidable state machine to the shared union flow: `resolve_variant` returns a new sentinel, `coerce` interprets it by re-reading the discriminator and resolving/coercing another target, and `dump` interprets the same sentinel differently. The helper also runs a temporary coercion/exactness calculation just to test tag membership. All four current unkeyed BaseModel variants expose an enum discriminator, and `Enum#===` already provides that membership contract.

Please keep keyed and unkeyed tag recognition inside target resolution: extract/normalize the tag once, resolve a keyed target or a matching declared unkeyed target, otherwise return the existing `Type::Unknown`; reserve `nil` for structural fallback when the tag is absent. Then ordinary converter dispatch handles coercion/dumping. This removes the sentinel, both special-case caller branches, duplicate extraction, and the helper that mixes recognition with execution.

I tested a temporary resolver-only version using declared field membership: the same 89 tests and 856 unknown cases pass, and serialized output is identical for 32 stable/Beta web-search representations. Please retain wire-compatibility tests when making this simplification.

markstuart-oai (commented):
Both prior correctness findings are fixed on this head: the eight stable/Beta web-search cases return their canonical models, and all four public create/stream text-extraction reproductions now return the known text while skipping future variants. The concrete-model guards are appropriate.

One maintainability issue remains in the union resolver: the new sentinel and coercion-specific fallback split discriminator resolution across multiple paths. The inline comment gives a simpler target-resolution design using the existing enum contract. I validated a temporary resolver-only alternative outside the checkout against the same tests and wire fixtures before suggesting it.

Current-head validation: 89 tests / 753 assertions passed; 856 unknown symbol/string cases across 214 unions preserve identity and serialization without structural fallback; 32 known web-search representations, parent-state preservation and nested model/array cases passed. CI showed 18 successful checks and two skipped checks. No changed file crosses 1,000 lines. Full suite/lint were not independently rerun.

lib/openai/internal/type/union.rb
jbeckwith-oai: Fixed in 92ee0739. `resolve_variant` now owns the full discriminator decision: it extracts and normalizes the tag once, resolves keyed variants, then checks enum-backed unkeyed BaseModel discriminator membership with `Enum#===`, and otherwise returns `Type::Unknown` (with `nil` reserved for absent tags). I also added coverage for an explicit `type: nil` so it cannot collide with unkeyed registry entries.

lib/openai/internal/type/union.rb
```diff
@@ -100,8 +100,25 @@ def variants = derefed_variants.map { _2 }
...
+              field = target.known_fields[@discriminator]
+              next unless field
+
+              type = field.fetch(:type_fn).call
+              next unless OpenAI::Internal::Type::Enum === type
+
+              return target if type === key
+            end
+
+            OpenAI::Internal::Type::Unknown
```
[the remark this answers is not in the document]
markstuart-oai: Confirmed independently on 92ee0739 through `client.chat.completions.stream` with WebMock. A synthetic first chunk containing `delta.tool_calls: [{index: 0, id: "call_fake", type: "future_tool", function: {name: "fake_tool", arguments: "{}"}}]` raises `NoMethodError` at line 329 before the terminal chunk is consumed. The known `function` control completes; replacing only the union implementation with the merge-base version also completes the unknown case, albeit with the incorrect known-model coercion this PR is fixing.

Please retain raw preservation and guard all three typed-only consumers: `parse_tool_calls!` (329), `tool_call_delta_events` (268), and `ChoiceEventState#tool_done_event` (754). Each currently uses `tool_call&.type`, which still calls `.type` on a non-nil Hash. Selecting the known function-call model before accessing its attributes keeps unknown snapshots observable without emitting function-specific events. Add a public stream regression with unknown and known calls/content through terminal completion. I assess this as P2; the previous resolver-structure and Responses text-consumer findings are fixed.

markstuart-oai (commented):
The prior structural finding is fixed: keyed and enum-backed unkeyed recognition now lives in resolve_variant, the extra sentinel/helper/coercion state is gone, and coerce/dump use ordinary target dispatch. Both earlier Responses correctness fixes also remain intact. No further structural or file-size blocker found.

One P2 caller regression remains in Chat Completions streaming. I independently reproduced the existing bot finding through the public stream API and added the reproducer and all three affected consumer locations to its inline thread: https://github.com/openai/openai-ruby/pull/645#discussion_r3942214015 . A future tool-call discriminator becomes a raw Hash and then crashes a typed-only path; the known-function control completes. Preserve raw unknown values and guard the function-specific parsing/delta/done consumers before approval.

Validation on this exact head: 106 tests / 855 assertions passed; all four original public Responses reproductions pass; 32 enum-backed web-search representations retain canonical models and wire output. The 214-union unknown sweep passes 856 symbol/string cases; an expanded canonical-resolver sweep passes 4,932 keyed-known and 2,556 explicit nil/non-string/empty-tag cases. The Realtime resolver that additionally requires role was excluded from the generic known-tag fixture sweep. CI showed 18 successful checks and two skipped checks. Full suite/lint and the author's adversarial-review claims were not independently rerun.

lib/openai/internal/type/union.rb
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in ff0a4837. ChatCompletionStreamState now treats only concrete ChatCompletionMessageFunctionToolCall snapshots as function calls before reading `.type`/`.function`; raw future tool-call hashes are preserved and skipped by parse, delta-event, and done-event paths. Added focused stream coverage proving an unknown tool-call discriminator remains raw and emits no function helper events.

markstuart-oai (approved):
All prior findings are addressed on this head. Unknown Chat tool calls now remain raw through parsing, delta events, and completion without entering function-only paths. I verified the original public failure and mixed known/unknown streams in both tool orders through terminal completion. Known calls retain their arguments/events and ordinary text remains available.

The canonical resolver still cleanly distinguishes known keyed/enum-backed tags, absent tags, and genuinely unknown tags. The Responses changes also simplify event construction and preserve unknown incremental slots with nil snapshots rather than fabricated models or prefixes. No remaining correctness, security, or maintainability finding in the complete nine-file diff; no file crosses 1,000 lines.

Validation: 108 tests / 863 assertions passed; all four prior public Responses reproductions pass; 32 known web-search representations and nested/state-preservation checks pass. Sweeps passed 856 unknown symbol/string cases, 4,932 keyed-known representations, and 2,556 nil/non-string/empty-tag cases. Public incremental Responses coverage confirms unknown slots do not interfere with known text accumulation. CI showed 18 successful checks and two skipped checks. Full suite/lint were not independently rerun.

PR #647, opened 2026-09-05
docs: clarify dormant compatibility-check boundary

jbeckwith-oai:
## Summary
- document that dormant developer helpers are not security controls by name alone
- record that scripts/detect-breaking-changes is not invoked by tracked workflows or tracked required jobs at this revision
- preserve reassessment if the helper later becomes required or privileged

## Validation
- confirmed the reported helper restores historical tests but invokes lint rather than runtime assertions
- confirmed no tracked workflow or Rake entrypoint invokes the helper
- confirmed the finding is a release-quality compatibility-signal defect, not a demonstrated security boundary crossing

## Verification
- git diff --check 49ccb69cbe4fd0a738bbc03840b7bb117bb183f6
- targeted repository search for detect-breaking-changes workflow/Rake callers
- bash -n scripts/detect-breaking-changes
- mandatory adversarial review: two consecutive clean rounds with two fresh independent reviewers per round

### DIFF
--- a/docs/architecture/security-model.md
+++ b/docs/architecture/security-model.md
@@ -206,6 +206,19 @@ PR CI. A contributor who can modify those tracked files does not gain a new
 privilege merely because CI executes them. Do not report that execution alone
 as a security finding.
 
+Dormant developer helpers are not security controls merely because their names
+suggest compatibility or safety verification. At this revision,
+`scripts/detect-breaking-changes` is not invoked by any tracked workflow or
+tracked required job; when a contributor explicitly runs it, it replaces
+selected tests with historical versions and invokes lint rather than runtime
+assertions.
+That can affect contributor confidence or release-quality review, but without a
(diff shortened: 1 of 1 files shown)


markstuart-oai (approved):
Verified against scripts/detect-breaking-changes, the Rake task graph, and tracked workflows at 856b7981. The helper restores selected historical tests and invokes lint/type checking rather than runtime assertions; no tracked workflow or Rake entrypoint invokes it. The clarification correctly separates that compatibility-signal defect from a demonstrated security boundary and explicitly requires reassessment if the helper becomes required or privileged, while preserving independent runtime-security findings. This does not establish external branch-protection or environment configuration.

The wording belongs in the canonical security model and introduces no structural or maintainability regression (267 → 280 lines). No substantive findings. Validation: git diff --check and bash -n scripts/detect-breaking-changes passed; runtime tests were not run for this documentation-only change.

PR #648, opened 2026-09-05
fix: stream container file uploads as multipart

jbeckwith-oai:
## Summary
- route non-nil container file uploads through the existing streaming multipart encoder
- preserve JSON requests for file_id, including callers forwarding file: nil
- add focused public-entrypoint regressions for multipart bytes/path handling and JSON compatibility

## Validation
The finding is real. containers.files.create forwarded file inputs with the default application/json content type, so request encoding JSON-generated the body instead of sending multipart file bytes. The container-files API accepts raw file content as multipart and file_id as JSON.

## Tests
- ruby -c lib/openai/resources/containers/files.rb
- ruby -c test/openai/resources/containers/files_test.rb
- ruby -Itest test/openai/resources/containers/files_test.rb
- ruby -Itest test/openai/internal/multipart_content_type_test.rb
- ruby -Itest test/openai/file_part_test.rb
- ruby -Itest test/openai/internal/util_test.rb
- rubocop --ignore-unrecognized-cops lib/openai/resources/containers/files.rb test/openai/resources/containers/files_test.rb
- git diff --check

The exact locked RuboCop bundle could not activate because the pinned syntax_tree-rbs Git revision is no longer available; the installed fallback RuboCop reported no offenses.

## Security review
The file branch now reaches the existing multipart encoder, preserving basename reduction, multipart header escaping, content-type validation, configured-origin enforcement, and streaming behavior. The file_id JSON branch remains unchanged.

### DIFF
--- a/lib/openai/resources/containers/files.rb
+++ b/lib/openai/resources/containers/files.rb
@@ -37,6 +37,7 @@ module OpenAI
           @client.request(
             method: :post,
             path: ["containers/%1$s/files", container_id],
+            headers: parsed[:file].nil? ? nil : {"content-type" => "multipart/form-data"},
             body: parsed,
             model: OpenAI::Models::Containers::FileCreateResponse,
             security: {bearer_auth: true},
--- a/test/openai/resources/containers/files_test.rb
+++ b/test/openai/resources/containers/files_test.rb
@@ -122,6 +122,7 @@ class OpenAI::Test::Resources::Containers::FilesTest < OpenAI::Test::ResourceTes
       request = value
       true
     end
+
     client = OpenAI::Client.new(
       api_key: "test-key",
       base_url: "http://example.test/v1",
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed the full diff and traced the container-file entrypoint through parameter dumping, the existing multipart encoder, and the transport boundary. This keeps the format choice in the resource: non-nil files use streaming multipart, while file_id requests (including forwarded file: nil) retain JSON. Reusing the canonical encoder preserves file metadata, escaping, IO replay rules, and cleanup without adding a shared mode or duplicate upload logic. No substantive issues found.

Validation: the two new regressions passed (15 assertions); multipart-content-type and FilePart suites passed (25 tests, 674 assertions); an independent public-entrypoint matrix passed (3 tests, 72 assertions) for binary/empty strings, StringIO offsets, IO/pathless IO, Pathname, FilePart metadata, string keys, JSON variants, and mutated MIME rejection. Both changed files pass Ruby syntax checks and the diff passes whitespace checks. The broader util suite run encountered nine existing CGI-harness NoMethodErrors with the cached Ruby 4.0/CGI environment; full-suite and live API validation were not performed.

PR #649, opened 2026-09-05
docs: clarify Realtime transcription event boundary

jbeckwith-oai:
## Summary
- document the exact Realtime trust boundary for the informational client-secret finding
- clarify that the legacy transcription_session.updated model is outside the active WebSocket server-event union
- keep the disposition narrow without changing runtime behavior or public APIs

## Validation
- finding disposition: non-impacting, high confidence
- active helper parses session.updated; transcription_session.updated becomes UnknownServerEvent before the legacy model can require client_secret
- WebSocket authentication occurs at handshake time; post-handshake transcription updates carry configuration

## Verification
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/realtime/connection_test.rb
- mise exec ruby@4.0.6 -- bundle exec rake build:docs
- git diff --check
- adversarial review: 3 rounds; round 1 correction addressed; rounds 2 and 3 clean

## Scope
- docs-only canonical security-model clarification
- no tracking records, dependency changes, runtime changes, or public API changes

### DIFF
--- a/docs/architecture/security-model.md
+++ b/docs/architecture/security-model.md
@@ -126,8 +126,21 @@ Important boundaries:
 - Authenticated Realtime handshakes are a transport boundary distinct from
   event parsing. Bearer credentials, caller-selected WebSocket destinations,
   TLS, proxy credentials, and traces must remain bound to their intended
-  origin and diagnostic audience
-  (`lib/openai/helpers/realtime/client_extension.rb:167-280`,
+  origin and diagnostic audience. Bearer credentials, including ephemeral
+  client secrets passed to client apps, authenticate the handshake before any
+  typed event is sent; post-handshake transcription `session.update` events
+  carry configuration rather than a client secret. The legacy generated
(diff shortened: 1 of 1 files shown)


markstuart-oai (approved):
Verified the clarification against the active server-event union, conversion path, transcription session models, and shared WebSocket handshake/connection lifecycle. At this revision, `transcription_session.updated` returns `UnknownServerEvent` before model coercion; `session.updated` accepts transcription configuration without `client_secret`. Handshake authentication remains distinct from post-handshake `session.update` configuration. This conclusion is specific to the current parser/union and should be reassessed if that dispatch changes.

The explanation belongs in the canonical security model and introduces no structural complexity or material missed simplification (267 → 280 lines). No substantive findings.

Validation: Ruby 4.0.6 with existing local gems: connection tests 32/227 assertions, authentication-retry tests 8/44 assertions, and synthetic boundary probes 2/9 assertions, all passing; `git diff --check` passed. The full Bundler environment was incomplete, so I did not rebuild the docs or run the full suite. No live API calls.

PR #651, opened 2026-09-05
fix: preserve hash-form type metadata

jbeckwith-oai:
## Summary
- preserve metadata from hash-form type declarations while retaining separate-spec precedence
- keep structural type keys excluded from metadata
- add regressions for BaseModel, ArrayOf, HashOf, precedence, and public structured-output JSON Schema constraints

## Validation
- ruby -Ilib -Itest test/openai/internal/type/base_model_test.rb
- ruby -Ilib -Itest test/openai/helpers/structured_output_test.rb
- ruby -c on changed Ruby files
- git diff --check

## Security
Hash-form metadata can include JSON Schema constraints such as pattern. Before this fix, the empty default spec won over hash-form metadata, weakening the schema sent for lower-trust model output. The shared collector now preserves those constraints without changing structural type selection.

### DIFF
--- a/lib/openai/internal/type/converter.rb
+++ b/lib/openai/internal/type/converter.rb
@@ -122,7 +122,10 @@ module OpenAI
           #
           # @return [Hash{Symbol=>Object}]
           def meta_info(type_info, spec)
-            [spec, type_info].grep(Hash).first.to_h.except(:const, :enum, :union, :nil?)
+            [type_info, spec]
+              .grep(Hash)
+              .reduce({}, :merge)
+              .except(:const, :enum, :union, :nil?)
           end
 
           # @api private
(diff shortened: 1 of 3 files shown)


markstuart-oai (approved):
Thanks! Reviewed the complete diff and traced declaration metadata through BaseModel, ArrayOf/HashOf, Union callers, JSON Schema conversion, and the Chat Completions/Responses request paths. The shared collector is the right place for this fix: merging metadata retains separate-spec precedence without changing structural type selection or adding caller-specific branches.

Verified hash-form constraints, frozen inputs, explicit overrides, nullable/optional behavior, nested and recursive schemas, and mocked public requests carrying pattern. Both changed test files, adjacent regressions, and additional probes pass locally: 125 tests / 858 assertions on Ruby 4.0.6. The new schema regression fails against the old collector as expected; syntax and diff checks also pass. No substantive correctness or maintainability findings.

PR #652, opened 2026-09-05
fix: deserialize diarized transcriptions correctly

jbeckwith-oai:
Ensure transcription response union selection prefers structurally specific verbose and diarized models before the permissive basic fallback. Add deterministic public-client regressions for typed diarized segments, generated constant defaults, missing-speaker fallback, and verbose empty-segment compatibility. Verified with client/resource tests, RuboCop, Ruby formatting, RBS validation, Sorbet typechecking, focused security review, and the required adversarial-review convergence.

### DIFF
--- a/lib/openai/models/audio/transcription_create_response.rb
+++ b/lib/openai/models/audio/transcription_create_response.rb
@@ -12,17 +12,17 @@ module OpenAI
       module TranscriptionCreateResponse
         extend OpenAI::Internal::Type::Union
 
-        # Represents a transcription response returned by model, based on the provided input.
-        variant -> { OpenAI::Audio::Transcription }
+        # Represents a verbose json transcription response returned by model, based on the provided input.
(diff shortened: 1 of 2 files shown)


markstuart-oai (approved):
Looks good. I traced the public transcription call through the strict union/model coercion path: trying the specialized models before the broad text-only model preserves typed diarized speaker segments, while required-field validation retains the basic fallback for incomplete segments. Verbose responses with empty segments still select the verbose model, and the shipped RBI/RBS union types remain consistent.

This is a direct fix in the response union without introducing a custom discriminator or duplicating parsing. Validated locally: 52 client tests (274 assertions), 12 union tests (92 assertions), and six additional model-shape probes covering basic/logprobs, empty diarized, populated/word-only verbose, and explicit diarized constants. No substantive correctness or maintainability findings.

PR #653, opened 2026-09-06
fix: keep Realtime metadata out of nested payloads

jbeckwith-oai:
## Problem

Realtime resource helpers already lift known event metadata such as `event_id` to the event envelope. When callers combine JSON-derived string keys with an explicit Ruby symbol-key override, the helper selected the symbol value but left the string-key copy inside the nested payload.

The supported public path deterministically serialized this split:

```ruby
params = JSON.parse('{"event_id":"template-event","instructions":"synthetic instructions"}')

client.realtime.connect(model: "gpt-realtime-2.1", transport: transport) do |connection|
  connection.response.create(**params, event_id: "caller-event")
end
```

Before:

```json
{"type":"response.create","event_id":"caller-event","response":{"instructions":"synthetic instructions","event_id":"template-event"}}
```

After:

```json
{"type":"response.create","event_id":"caller-event","response":{"instructions":"synthetic instructions"}}
```

This is proven deterministic public-wire behavior for callers composing supported Realtime helpers with JSON-loaded configuration and explicit metadata overrides. Customer incidence and live API behavior are unmeasured.

## Fix

The existing `event_payload` helper now chooses the envelope value using the same symbol-key precedence, including explicit `nil`, then deletes both the symbol and string spellings of each known metadata key before serializing the nested payload.

The focused public-path regression covers:

- `session.update`, `response.create`, and `conversation.item.create`
- mixed JSON string keys plus explicit symbol overrides
- `event_id` and `previous_item_id`
- explicit `nil` precedence
- ordinary nested payload values and caller input immutability
- existing single-form string-key and symbol-key behavior

## Compatibility and scope

No public API, metadata precedence, generated model, transport, authentication, URL, WebSocket protocol, dependency, or broad key-normalization behavior changes. The diff is limited to the existing Realtime metadata extraction helper and one focused test file.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/realtime/connection_metadata_keys_test.rb` — 2 runs, 9 assertions
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/realtime/connection_test.rb` — 32 runs, 227 assertions
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — rubyfmt, RuboCop (2,920 files), Sorbet examples, and RBS validation passed
- `TEST_API_BASE_URL=http://127.0.0.1:49137 mise exec ruby@4.0.6 -- bundle exec rake test` — 1,760 runs, 15,176 assertions, 0 failures, 0 errors, 1 skip
- trusted current-main public custom-code budget — success, 3,450 / 4,000 lines

Adversarial review converged with two consecutive clean rounds, each using two fresh independent read-only reviewers. The serialization-boundary security assessment found no auth, transport, logging, deserialization, dependency, or CI/security-surface expansion.

## Limitations

No live API calls were made. The evidence is synthetic and uses the supported public fake-transport path.

### DIFF
--- a/lib/openai/helpers/realtime/connection_resources.rb
+++ b/lib/openai/helpers/realtime/connection_resources.rb
@@ -14,7 +14,7 @@ module OpenAI
         private def event_payload(params, *metadata_keys)
           payload = params.to_h.dup
           metadata = metadata_keys.to_h do |key|
-            value = payload.key?(key) ? payload[key] : payload[key.to_s]
+            value = payload.key?(key) ? payload.fetch(key) : payload.fetch(key.to_s, nil)
             payload.delete(key)
             payload.delete(key.to_s)
             [key, value]
--- a/lib/openai/helpers/realtime/connection_resources.rb
+++ b/lib/openai/helpers/realtime/connection_resources.rb
@@ -14,7 +14,9 @@ module OpenAI
         private def event_payload(params, *metadata_keys)
           payload = params.to_h.dup
           metadata = metadata_keys.to_h do |key|
-            value = payload.key?(key) ? payload.delete(key) : payload.delete(key.to_s)
+            value = payload.key?(key) ? payload[key] : payload[key.to_s]
+            payload.delete(key)
+            payload.delete(key.to_s)
             [key, value]
           end
 
--- a/test/openai/realtime/connection_metadata_keys_test.rb
+++ b/test/openai/realtime/connection_metadata_keys_test.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+
+require_relative "connection_test_support"
+
+class OpenAI::Test::RealtimeConnectionMetadataKeysTest < Minitest::Test
+  include OpenAI::Test::RealtimeConnectionTestSupport
+
+  def test_metadata_overrides_do_not_leave_json_keys_in_nested_payloads
+    session_params = JSON.parse(
+      "{\"event_id\":\"template-session\",\"instructions\":\"synthetic session instructions\"}"
+    )
+    response_params = JSON.parse(
+      "{\"event_id\":\"template-response\",\"instructions\":\"synthetic response instructions\"}"
+    )
+    item_params = JSON.parse(
+      "{\"event_id\":\"template-item\",\"previous_item_id\":\"template-previous\"," \
+        "\"type\":\"message\",\"role\":\"user\",\"content\":[{\"type\":\"input_text\",\"text\":\"Hello\"}]}"
+    )
+    original_session_params = session_params.dup
+    original_response_params = response_params.dup
+    original_item_params = Marshal.load(Marshal.dump(item_params))
+    socket = FakeSocket.new
+
+    client.realtime.connect(model: "gpt-realtime-2.1", transport: FakeTransport.new(socket)) do |connection|
+      connection.session.update(**session_params, event_id: nil)
+      connection.response.create(**response_params, event_id: "caller-response")
+      connection.conversation.items.create(
+        **item_params,
+        event_id: "caller-item",
+        previous_item_id: "caller-previous"
+      )
+    end
+
+    session_event, response_event, item_event = socket.writes.map { JSON.parse(_1) }
+    assert_equal(
+      {
+        "type" => "session.update",
+        "session" => {
+          "type" => "realtime",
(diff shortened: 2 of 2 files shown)


lib/openai/helpers/realtime/connection_resources.rb
```diff
@@ -14,7 +14,9 @@ def initialize(connection)
         private def event_payload(params, *metadata_keys)
           payload = params.to_h.dup
           metadata = metadata_keys.to_h do |key|
-            value = payload.key?(key) ? payload.delete(key) : payload.delete(key.to_s)
+            value = payload.key?(key) ? payload[key] : payload[key.to_s]
```
markstuart-oai: [P2] Preserve absence without invoking Hash defaults

I independently confirmed the existing bot feedback through `client.realtime.connect(..., transport: FakeTransport.new(socket))`. Passing `params = Hash.new("root").merge(type: :message, role: :user, content: [{type: :input_text, text: "Hello"}])` to `connection.conversation.items.create(**params)` now emits both `event_id: "root"` and `previous_item_id: "root"`, although neither key exists. The base helper omitted both. Per the generated contract, this changes insertion from append-to-end to beginning-of-conversation; default procs likewise fabricate IDs for all three helpers.

Keep the current shared extraction structure, but use a presence-only read, e.g. `payload.key?(key) ? payload.fetch(key) : payload.fetch(key.to_s, nil)`, followed by both deletions. That preserves explicit nil precedence without another branch or broader normalization. Add a public fake-transport regression for missing metadata with a default value/default proc, including previous_item_id.

markstuart-oai (commented):
Reviewed both changed files, all three event_payload callers, generated event contracts, serialization, shared fake transport, and existing connection tests. The mixed-key fix preserves symbol precedence (including nil), removes both forms, and stays in the correct shared helper. No structural duplication or decomposition issue; runtime file grows 130→132 lines and the new focused test is 101 lines.

One P2 regression remains: absent metadata now consults Hash defaults. I independently reproduced the prior bot finding through the public transport path, including a fabricated previous_item_id that changes insertion semantics. Please address the inline finding before approval.

Validation on Ruby 4.0.6: metadata tests 2 runs/9 assertions; connection tests 32 runs/227 assertions, all passing; syntax checks pass. Separate public-path probes compared all three helpers with default-value/default-proc hashes against the base helper, and checked explicit nil/frozen input. Full suite and lint reported by the author were not rerun; no live API calls.

lib/openai/helpers/realtime/connection_resources.rb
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d20684b2: the fallback now uses a presence-preserving `fetch(key.to_s, nil)` after the symbol-key presence check, so absent metadata does not invoke Hash defaults/default procs while explicit symbol `nil` still wins. Added public FakeTransport coverage across session, response, and conversation item helpers.

jbeckwith-oai: Fixed in d20684b2: absent metadata is now read with `fetch(key.to_s, nil)` rather than `[]`, preserving absence without invoking Hash defaults/default procs while retaining dual-form deletion and symbol precedence. The new public FakeTransport regression covers default values/default procs and verifies missing `previous_item_id` does not change conversation insertion semantics.

markstuart-oai (approved):
Re-reviewed the complete two-file diff at d20684b2 and the author replies to my prior review 5125963504. The P2 Hash-default regression is fixed: fetch preserves absent metadata without invoking defaults, symbol presence retains explicit nil precedence, and both spellings are removed from the nested payload.

No remaining correctness or structural findings. Extraction stays in the canonical helper without new branches, duplicate normalization, or unnecessary abstractions. Runtime file: 130→132 lines; focused test file: 152 lines.

Validated on Ruby 4.0.6 using cached gems: metadata tests 3 runs/14 assertions and connection tests 32 runs/227 assertions, all passing. An independent 30-case public FakeTransport matrix passed across all three callers, both default-value/default-proc hashes, absent/string/symbol/mixed/nil metadata, ordinary nested values, and frozen-input immutability. Syntax and diff whitespace checks passed. Relevant generated contracts, signatures, serialization and shared test support were also checked. Full suite/lint were not rerun; no live API calls.

PR #655, opened 2026-09-06
fix: isolate structured output field metadata

jbeckwith-oai:
## Problem

When the same nested structured-output model is used by more than one field, metadata attached to the first field could leak onto later fields. For example:

```ruby
class Address < OpenAI::BaseModel
  required :city, String
end

class Order < OpenAI::BaseModel
  required :billing, Address, doc: "Billing address only"
  required :shipping, Address
  required :pickup, Address
end

schema = Order.to_json_schema
```

Before this change, the synthetic public schema incorrectly contained:

```ruby
{
  shipping: {
    allOf: [
      {:$ref => "#/$defs/.billing"},
      {description: "Billing address only"}
    ]
  }
}
```

After this change, only the annotated field retains that metadata:

```ruby
{
  billing: {description: "Billing address only", ...},
  shipping: {:$ref => "#/$defs/.billing"},
  pickup: {:$ref => "#/$defs/.billing"}
}
```

This is deterministic public-method evidence for supported reused nested models. Real-world incidence is unmeasured.

## Fix

At the existing cached-reference reuse boundary, later uses now receive a fresh hash containing only the cached `$ref`. Field-local metadata and internal reference bookkeeping remain attached only to the field that added them.

The fix preserves the existing definition target, schema shape, nilability/recursion handling, repeated conversion behavior, and runtime coercion. It does not add schema keywords, APIs, abstractions, dependencies, or generated-file changes.

## Tests

- Added a focused public `to_json_schema` regression for one documented and two unannotated uses of the same nested model, including deterministic repeated conversion.
- Added a safe synthetic Chat Completions request regression that inspects the serialized request schema through WebMock.
- Proved the new regression red on the old reuse line: 2 runs, 3 assertions, 2 failures.
- Proved it green with the fix: 2 runs, 5 assertions, 0 failures.
- Focused structured-output/reference/nullable coverage: 56 runs, 248 assertions, 0 failures.
- Adjacent constant/Sorbet/Chat Completions parser/Responses parser coverage: 27 runs, 323 assertions, 0 failures.
- Full suite with the supported Ruby and local synthetic mock: 1760 runs, 15172 assertions, 0 failures, 0 errors, 1 skip.
- Trusted public custom-code budget against freshly fetched main: success, 3450 / 4000 custom lines.

## Review and security

Two consecutive adversarial-review rounds each used exactly two new independent read-only reviewers. Both rounds were clean with no in-scope findings and no code changes between rounds.

Security assessment: this is a local schema serialization/hash-reuse correction. It does not touch authentication, network transport, deserialization, logging, files, dependencies, CI, or release behavior, and introduces no new trust boundary.

## Scope

Changed only the existing cached-reference reuse line and one focused regression test. Broader schema support audits, model/framework changes, public API redesign, and unrelated cleanup are explicitly out of scope.

### DIFF
--- a/lib/openai/helpers/structured_output/json_schema_converter.rb
+++ b/lib/openai/helpers/structured_output/json_schema_converter.rb
@@ -109,7 +109,7 @@ module OpenAI
                 stored[OpenAI::Helpers::StructuredOutput::JsonSchemaConverter::RECURSIVE] = true
               end
 
-              pointers.first.except(OpenAI::Helpers::StructuredOutput::JsonSchemaConverter::NO_REF).tap do
+              pointers.first.slice(:$ref).tap do
                 pointers << _1
               end
             else
--- a/test/openai/helpers/structured_output_field_metadata_test.rb
+++ b/test/openai/helpers/structured_output_field_metadata_test.rb
@@ -1,0 +1,109 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
+class OpenAI::Test::StructuredOutputFieldMetadataTest < Minitest::Test
+  extend Minitest::Serial
+  include WebMock::API
+
+  class SyntheticAddress < OpenAI::BaseModel
+    required :city, String
+  end
+
+  class SyntheticOrder < OpenAI::BaseModel
+    required :billing, SyntheticAddress, doc: "Billing address only"
+    required :shipping, SyntheticAddress
+    required :pickup, SyntheticAddress
+  end
+
+  def before_all
+    super
+    WebMock.enable!
+  end
+
+  def after_all
+    WebMock.disable!
+    super
+  end
+
+  def setup
+    super
+    @client = OpenAI::Client.new(base_url: "http://localhost", api_key: "test-key")
+  end
+
+  def teardown
+    WebMock.reset!
+    super
+  end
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed both changed files and the metadata, inheritance, reference finalization, array/enum/union, and Chat/Responses conversion paths at 85a94511b23848b03bef39bbb01bc0bae0d492bc.

The cache reuse boundary is the right ownership boundary: copying only `$ref` prevents first-field metadata from propagating while retaining the shared reference string needed to finish recursive definitions. No extra copying/freezing abstraction is needed. Field/type contracts and repeated conversion remain intact; no correctness or structural findings.

Validation: 78 focused repository tests / 547 assertions and 4 independent ownership probes / 71 assertions passed. The added tests and all four probes reproduce the leak with the merge-base converter. Tested inherited/sibling models, nested metadata, reused model/enum/array/union types, nullable fields, recursion, and public request serialization. Converter remains 239 lines; new regression file is 109 lines.

Limits: used existing local gems rather than the unavailable locked bundle. Broader Sorbet checks encountered missing rbs/srb/sorbet-runtime tooling. Full suite, lint, budget checker, and live API were not run.

PR #656, opened 2026-09-06
fix: deduplicate string-keyed query overrides

jbeckwith-oai:
## Problem

`request_options[:extra_query]` documents string-keyed overrides, but a JSON-derived key did not replace the generated Symbol key at the request merge boundary. The native prepared URL contained duplicate wire keys.

```ruby
require "json"

client.files.list(
  purpose: :assistants,
  request_options: {extra_query: JSON.parse("{\"purpose\":\"batch\"}")}
)
```

Synthetic native `HTTPClient#execute` capture before this change:

```text
http://localhost/files?purpose=assistants&purpose=batch
```

After this change, including `page.next_page`:

```text
http://localhost/files?purpose=batch
http://localhost/files?purpose=batch&after=file-one
```

## Fix

At the existing request-query merge boundary, equivalent String and Symbol extra-query keys are remapped onto the generated query key representative before the existing deep merge. The paginated effective-query storage applies the same narrow reconciliation once so follow-up pages do not reintroduce the duplicate. This keeps override precedence while preserving the existing URI join behavior for base/path query concatenation.

## Compatibility and scope

The change preserves distinct keys, values, array/bracket encoding, caller-owned input hashes, cursor replacement, and base/path query concatenation. It does not change body merging, headers, retries, authentication, origin/path validation, global utilities, generated APIs, public APIs, dependencies, or budget/config metadata.

Reach confidence: string-keyed query configuration from JSON is a supported public workflow and the duplicate native wire URL is deterministic. Customer incidence and server parsing or rejection are unmeasured.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/query_override_keys_test.rb` — 3 runs, 6 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/base_query_isolation_test.rb` — 2 runs, 6 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec ruby test/openai/path_parameter_query_test.rb` — 17 runs, 36 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — RuboCop 2920 files clean, Sorbet no errors, 1286 RBS validated
- Serialized full suite with Steady mock — 1761 runs, 15173 assertions, 0 failures, 0 errors, 1 skip
- Trusted current-main public custom-code budget — 3450 / 4000 custom lines, 550 headroom
- `git diff --check` — clean

## Review and security

Four adversarial-review rounds used exactly two fresh read-only reviewers each. Round 1 found and repaired a base/path query concatenation regression; rounds 3 and 4 were consecutive clean rounds on the final unchanged candidate. A scoped security assessment found no new credential, origin, logging, deserialization, dependency, or CI surface: this only reconciles caller-supplied query key spelling immediately before existing serialization.

### DIFF
--- a/lib/openai/internal/transport/base_client.rb
+++ b/lib/openai/internal/transport/base_client.rb
@@ -414,20 +414,10 @@ module OpenAI
 
           path = OpenAI::Internal::Util.interpolate_path(uninterpolated_path)
 
-          query = req[:query].to_h
-          query_keys = query.each_key.with_object({}) do |key, keys|
-            keys[key.to_s] = key if key.is_a?(String) || key.is_a?(Symbol)
-          end
-
-          extra_query = opts[:extra_query].to_h.transform_keys do |key|
-            if key.is_a?(String) || key.is_a?(Symbol)
-              query_keys.fetch(key.to_s, key)
-            else
-              key
-            end
-          end
-
-          query = OpenAI::Internal::Util.deep_merge(query, extra_query)
+          query = OpenAI::Internal::Transport::RequestBodyMerge.merge(
+            req[:query].to_h,
+            opts[:extra_query].to_h
+          )
 
           url = OpenAI::Internal::Util.join_parsed_uri(
             @base_url_components,
--- a/lib/openai/internal/transport/base_client.rb
+++ b/lib/openai/internal/transport/base_client.rb
@@ -414,7 +414,20 @@ module OpenAI
 
           path = OpenAI::Internal::Util.interpolate_path(uninterpolated_path)
 
-          query = OpenAI::Internal::Util.deep_merge(req[:query].to_h, opts[:extra_query].to_h)
+          query = req[:query].to_h
+          query_keys = query.each_key.with_object({}) do |key, keys|
+            keys[key.to_s] = key if key.is_a?(String) || key.is_a?(Symbol)
+          end
+
+          extra_query = opts[:extra_query].to_h.transform_keys do |key|
+            if key.is_a?(String) || key.is_a?(Symbol)
+              query_keys.fetch(key.to_s, key)
+            else
+              key
+            end
+          end
+
+          query = OpenAI::Internal::Util.deep_merge(query, extra_query)
 
           url = OpenAI::Internal::Util.join_parsed_uri(
             @base_url_components,
--- a/lib/openai/internal/type/base_page.rb
+++ b/lib/openai/internal/type/base_page.rb
@@ -59,8 +59,21 @@ module OpenAI
         # @param page_data [Object]
         def initialize(client:, req:, response_metadata:, page_data:)
           options = req[:options].to_h
+          query = req[:query].to_h
+          query_keys = query.each_key.with_object({}) do |key, keys|
+            keys[key.to_s] = key if key.is_a?(String) || key.is_a?(Symbol)
+          end
+
+          extra_query = options[:extra_query].to_h.transform_keys do |key|
+            if key.is_a?(String) || key.is_a?(Symbol)
+              query_keys.fetch(key.to_s, key)
+            else
+              key
+            end
+          end
+
           query = OpenAI::Internal::Util
-            .deep_merge(req[:query].to_h, options[:extra_query].to_h)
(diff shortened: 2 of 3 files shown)


lib/openai/internal/transport/base_client.rb
```diff
@@ -414,7 +414,20 @@ def initialize(
...
 
-          query = OpenAI::Internal::Util.deep_merge(req[:query].to_h, opts[:extra_query].to_h)
+          query = req[:query].to_h
+          query_keys = query.each_key.with_object({}) do |key, keys|
+            keys[key.to_s] = key if key.is_a?(String) || key.is_a?(Symbol)
+          end
+
+          extra_query = opts[:extra_query].to_h.transform_keys do |key|
+            if key.is_a?(String) || key.is_a?(Symbol)
+              query_keys.fetch(key.to_s, key)
```
markstuart-oai: [P2] Reuse the canonical exact-key-first merge policy

This remaps even an override whose exact key already exists. With `query: {purpose: "symbol", "purpose" => "string"}` and `extra_query: {purpose: "new"}`, the public fake transport receives `?purpose=symbol&purpose=new`; base correctly sends `?purpose=new&purpose=string`. The override replaces the other entry and leaves its exact target unchanged. Reversing the input key order produces the symmetric regression.

`Internal::Transport::RequestBodyMerge` already implements the policy needed here: preserve exact matches, consult string/symbol counterparts only when the exact key is absent, and merge without mutating inputs. Copying a different reconciliation loop here and again in `BasePage#initialize` creates three policy implementations, with observable drift already. Please generalize/reuse that canonical policy at both query call sites and delete these duplicated blocks, retaining the separate base/path URI concatenation behavior. Add public wire-URL regressions for both original key spellings/insertion orders, including pagination. This is an existing-key precedence issue, not a request to remove legitimate repeated array or base/path query values.

markstuart-oai (commented):
Reviewed all three changed files and the request-option, generated-query serialization, URI joining, and pagination paths. The ordinary JSON-keyed override is fixed, but the duplicated reconciliation introduces an exact-key precedence regression and should reuse the existing merge policy before approval.

Validation: Ruby 4.0.6, 28 focused tests / 68 assertions pass using locally available gems; public fake-transport base/head controls reproduce both the intended fix and the inline regression. The new PR tests fail twice on the merge base as expected. No new 1,000-line crossing: BaseClient 1,012 → 1,025, BasePage 83 → 96, new test 87 lines. The full locked bundle was unavailable; a broader util run has the same nine CGI-related errors on base and head. No live API calls or source edits.

lib/openai/internal/transport/base_client.rb
jbeckwith-oai: Fixed in 963a7b85. Both query call sites now reuse RequestBodyMerge.merge, whose exact-key-first behavior overrides an existing exact String or Symbol key before consulting its counterpart. I added native HTTPClient URL regressions for both mixed-key insertion orders across the initial request and next_page, while retaining the JSON-key deduplication and base/path concatenation cases.

markstuart-oai (approved):
Re-reviewed all three changed files at 963a7b8539ecdf0e7201d0d8400aa2d7fde86a4d after the response to review 5126003473. The previous P2 is addressed: both query call sites now reuse the canonical exact-key-first merge policy, deleting the duplicated reconciliation loops. Existing exact String/Symbol keys retain precedence, while JSON-keyed overrides remain deduplicated and base/path query concatenation stays separate. No remaining substantive findings.

Validation: 29 focused tests / 70 assertions, the independent original regression (1 test / 2 assertions), and three-page probes across all three cursor-page types (3 tests / 210 assertions) pass on Ruby 4.0.6. Probes cover both spellings/insertion orders, nil/false/empty values, repeated arrays, frozen inputs, and cursor replacement. BaseClient is 1,015 lines (base 1,012; previous head 1,025); BasePage returns to 83 lines. No new 1k crossing or replacement helper/branch layer. Tests used available local gems; the locked bundle was unavailable, and no full-suite/lint/typecheck or live API validation was performed.

PR #657, opened 2026-09-06
fix: declare Chat stream finish exception types

jbeckwith-oai:
# PR draft

Title: fix: declare Chat stream finish exception types

## Problem

The gem already raises OpenAI::LengthFinishReasonError and
OpenAI::ContentFilterFinishReasonError from structured Chat completion streams,
but shipped Sorbet and RBS declarations omit those public constants and their
OpenAI::Helpers::Streaming::StreamError superclass. Typed applications cannot
rescue the existing errors, and Sorbet cannot inspect the completion carried by
the length error.

## User-facing usage

~~~ruby
begin
  stream.get_final_completion
rescue OpenAI::LengthFinishReasonError => error
  completion = error.completion
  puts completion.choices.first.finish_reason
rescue OpenAI::ContentFilterFinishReasonError
  # Handle the existing content-filter termination.
end
~~~

## Before / after

With a synthetic WebMock-backed structured Chat stream ending in
finish_reason: "length", runtime already raises
OpenAI::LengthFinishReasonError with a parsed completion. Before this change,
Sorbet reports unresolved OpenAI::LengthFinishReasonError and
OpenAI::ContentFilterFinishReasonError, then reports that completion does
not exist; rbs method OpenAI::LengthFinishReasonError completion cannot find
the class. After this change, the same public rescue path resolves in Sorbet
and RBS, and the focused runtime regression still observes the existing length
completion and content-filter error.

## Fix

- Add RBI declarations for the existing helper exception hierarchy, public root
  aliases, and precise parsed-completion reader/keyword constructor.
- Add RBS declarations for the same hierarchy and aliases, using the already
  declared OpenAI::Models::Chat::ChatCompletion supertype rather than creating
  a broader parsed-Chat RBS hierarchy.
- Add one focused regression for both type systems and WebMock-backed public
  Chat streams.
- Explicitly declare the existing zero-argument content-filter constructor in
  both type systems, with positive and negative constructor probes.

## Compatibility and scope

This is declaration-only for existing runtime classes. It does not change
exception behavior, stream parsing, refusal semantics, public APIs, generated
models, dependencies, or Task 1's sig/openai/streaming.rbs.

## Verification

- mise exec ruby@4.0.6 -- bundle exec rake typecheck
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_initial_finish_test.rb
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/helpers/chat_stream_error_types_test.rb
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/gem_packaging_test.rb
- TEST_API_BASE_URL=http://127.0.0.1:49137 mise exec ruby@4.0.6 -- bundle exec rake test
- git diff --check
- Trusted current-main public custom-code budget check: success, 3450/4000.
- Adversarial review: two consecutive clean rounds, each with two new
  independent read-only reviewers.

Full suite result after feedback: 1763 runs, 15187 assertions, 0 failures, 0 errors, 1 skip.

## Limitations

Customer incidence is unmeasured. The regression uses safe synthetic data and
does not make a live API request.

### DIFF
--- a/sig/openai/helpers/streaming/exceptions.rbs
+++ b/sig/openai/helpers/streaming/exceptions.rbs
@@ -1,0 +1,23 @@
+module OpenAI
+  module Helpers
+    module Streaming
+      class StreamError < StandardError
+      end
+
+      class LengthFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+        attr_reader completion: OpenAI::Models::Chat::ChatCompletion
+
+        def initialize: (
+          completion: OpenAI::Models::Chat::ChatCompletion
+        ) -> void
+      end
+
+      class ContentFilterFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+      end
+    end
+  end
+
+  class LengthFinishReasonError = Helpers::Streaming::LengthFinishReasonError
+
+  class ContentFilterFinishReasonError = Helpers::Streaming::ContentFilterFinishReasonError
+end
--- a/sig/openai/helpers/streaming/exceptions.rbs
+++ b/sig/openai/helpers/streaming/exceptions.rbs
@@ -13,6 +13,7 @@ module OpenAI
       end
 
       class ContentFilterFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+        def initialize: -> void
       end
     end
   end
--- a/rbi/openai/helpers/streaming/exceptions.rbi
+++ b/rbi/openai/helpers/streaming/exceptions.rbi
@@ -1,0 +1,27 @@
+# typed: strong
+
+module OpenAI
+  module Helpers
+    module Streaming
+      class StreamError < StandardError
+      end
+
+      class LengthFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+        sig { returns(OpenAI::Chat::ParsedChatCompletion) }
+        attr_reader :completion
+
+        sig do
+          params(completion: OpenAI::Chat::ParsedChatCompletion).returns(T.attached_class)
+        end
+        def self.new(completion:)
+        end
+      end
+
+      class ContentFilterFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+      end
+    end
+  end
+
+  LengthFinishReasonError = Helpers::Streaming::LengthFinishReasonError
+  ContentFilterFinishReasonError = Helpers::Streaming::ContentFilterFinishReasonError
+end
--- a/rbi/openai/helpers/streaming/exceptions.rbi
+++ b/rbi/openai/helpers/streaming/exceptions.rbi
@@ -18,6 +18,9 @@ module OpenAI
       end
 
       class ContentFilterFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+        sig { returns(T.attached_class) }
+        def self.new
+        end
       end
     end
   end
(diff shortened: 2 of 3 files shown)


sig/openai/helpers/streaming/exceptions.rbs
```diff
@@ -0,0 +1,23 @@
...
+      class LengthFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+        attr_reader completion: OpenAI::Models::Chat::ChatCompletion
+
+        def initialize: (
+          completion: OpenAI::Models::Chat::ChatCompletion
+        ) -> void
+      end
+
+      class ContentFilterFinishReasonError < OpenAI::Helpers::Streaming::StreamError
+      end
```
markstuart-oai, line 17: [P2] Declare the zero-argument content-filter constructor in both type systems

I independently verified the bot rationale: `OpenAI::ContentFilterFinishReasonError.new("message")` passes both Sorbet and Steep with these shipped declarations, but the real `lib/openai/helpers/streaming/exceptions.rb:19` initializer raises `ArgumentError: wrong number of arguments (given 1, expected 0)`. The empty class inherits the message-taking Exception constructor contract. Add `def initialize: () -> void` here and an explicit zero-argument `self.new` signature in the RBI, preserving the existing runtime behavior. Extend the probes to accept `.new` and reject `.new("message")`; the current content-filter coverage only rescues the class, so it misses this mismatch.

markstuart-oai (commented):
Reviewed all three changed files against the actual PR merge-base, plus runtime loading, both finish paths, public aliases/rescues, callers, type contracts, and the existing stream tests. The declarations belong beside the canonical streaming exceptions; no runtime wrappers, duplicated exception definitions, branching growth, or 1k-line crossings (new files: 27/23/184 lines). The RBS completion supertype is a reasonable scoped choice.

One P2 contract mismatch remains: I independently reproduced the bot finding below in both type checkers. Please make the content-filter constructor match runtime and cover rejected arguments.

Validation: Ruby 4.0.6; focused stream suites passed (54 tests, 288 assertions), project Sorbet/RBS typecheck passed (1,287 RBS files), and negative controls rejected an incorrect completion type/nonexistent method. Both checkers currently accept the invalid content-filter constructor, while runtime raises ArgumentError. No live API requests or full-suite run.

sig/openai/helpers/streaming/exceptions.rbs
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 04836664. ContentFilterFinishReasonError now declares its existing zero-argument constructor in both RBI and RBS, and the focused regression verifies .new succeeds while .new("message") is rejected by runtime, Sorbet, and RBS.

jbeckwith-oai: Fixed in 04836664. ContentFilterFinishReasonError now declares its existing zero-argument constructor in both RBI and RBS, and the focused regression verifies .new succeeds while .new("message") is rejected by runtime, Sorbet, and RBS.

markstuart-oai (approved):
Thanks for addressing the constructor contract. Re-reviewed all three current changed files and the runtime loading, aliases, inheritance, finish paths, and completion contracts after replies 3944775079/3944775313. The P2 from review 5126019490 is fixed: runtime, Sorbet, and Steep accept the zero-argument content-filter constructor and reject the message argument. Independent negative controls also verify that the length completion type is enforced.

No remaining substantive findings. The declarations stay in their canonical location without runtime wrappers or new branching; new files are 30/24/215 lines. Sharing the test subprocess harness remains optional cleanup.

Validation on Ruby 4.0.6: 56 focused tests / 298 assertions; independent probes 2 / 16; packaging test 1 / 32; project Sorbet and RBS validation (1,287 files); git diff --check. All passed. No full-suite run or live API requests.

PR #665, opened 2026-09-06
fix: preserve file semantics for pathless multipart IO

### LINKED ISSUE #664: Preserve file semantics for pathless multipart IO in Client#request
Follow-up to #630. The generated `FileInput` path wraps a pathless IO in `FilePart(filename: "upload")`, but the generic `Client#request` path calls `Util.encode_content` directly. After #630, a raw IO with `to_path == nil` inside a multipart hash reaches `write_multipart_chunk` without a filename. Common multipart parsers can then classify the part as a regular form value rather than a file.

It would be safer for the raw `Pathname | IO` multipart branch to apply the same fallback filename as `FileInput.dump`, with a regression through public `Client#request` using `IO.pipe`.

sylvesterkaczmarek:
## Summary

- give raw pathless multipart `IO` values the same `upload` fallback filename used by `FileInput.dump`
- preserve existing basenames for path-backed `IO` and `Pathname` values
- cover the behavior both at the multipart encoder boundary and through public `Client#request`

This keeps a pathless stream classified as a file part instead of a regular multipart form value.

Fixes #664.

## Validation

- `bundle _4.0.17_ exec ruby -Itest test/openai/internal/util_test.rb`
- `bundle _4.0.17_ exec ruby -Itest test/openai/http_client_test.rb`
- `bundle _4.0.17_ exec rubocop lib/openai/internal/util.rb test/openai/internal/util_test.rb test/openai/http_client_test.rb`
- `git diff --check`

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -554,11 +554,11 @@ module OpenAI
           in OpenAI::FilePart unless val.filename.nil?
             filename = escape_multipart_header_param(val.filename)
             y << "; filename=\"#{filename}\""
-            # `IO#to_path` is nil for a stream with no backing path, such as a pipe
-            # or a socket. Such a part is sent without a filename, which RFC 7578
-            # allows, rather than failing the whole request.
-          in Pathname | IO unless val.to_path.nil?
-            filename = escape_multipart_header_param(::File.basename(val.to_path))
+          in Pathname | IO
+            # Keep raw multipart IO consistent with FileInput.dump: pathless streams
+            # are still file parts and use the same stable fallback filename.
+            filename = val.to_path.nil? ? "upload" : ::File.basename(val.to_path)
+            filename = escape_multipart_header_param(filename)
             y << "; filename=\"#{filename}\""
           else
           end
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (approved):

markstuart-oai (approved):
Reviewed all three changed files and the surrounding FileInput/FilePart, multipart streaming, public request, and replayability contracts at 700227d8a665d313f55a46b0103dab2ed0ac0de5. No actionable correctness or structural findings. The fallback belongs in the existing raw-IO filename branch; it preserves explicit FilePart metadata and path-backed basenames without adding a helper or another processing layer. No file crosses 1,000 lines.

Validation: an independent Ruby encoder probe passed 7 tests / 69 assertions (pipes, sockets, binary bytes, boundaries, path-backed IO/Pathname, cursor and ownership behavior, explicit/unnamed parts, and FileInput fallback). The pipe/socket regression probes fail against the actual base encoder because the filename is absent. All three changed files pass Ruby syntax checks; git diff --check passes. The repository utility and HTTP-client suites could not start because async is missing and bundle check reports unsatisfied dependencies; public-client integration was reviewed statically, not executed. No live API or full-suite validation claimed.

sylvesterkaczmarek:
Thanks for the thorough review and merge. The checks around raw IO, path-backed inputs, multipart streaming, and replayability were especially useful confirmation that the fallback stays confined to the intended filename branch.

PR #666, opened 2026-09-07
fix: match the multipart request content type case-insensitively

thegoodengineer:
## Problem

`encode_content` dispatches multipart bodies on an inline pattern, the only one in that `case` that is not a named, anchored, case-insensitive constant:

https://github.com/openai/openai-ruby/blob/a63eda82/lib/openai/internal/util.rb#L637

Its siblings a few lines above are both `%r{\A...[ \t]*(?:;|\z)}i`:

```ruby
JSON_CONTENT  = %r{\Aapplication/(?:[a-z0-9.-]+\+)?json[ \t]*(?:;|\z)}i
JSONL_CONTENT = %r{\Aapplication/(?:x-[nl]djson|(?:x-)?jsonl)[ \t]*(?:;|\z)}i
```

Media type names are case-insensitive, so `Multipart/Form-Data` names the same type. It did not match:

```ruby
file = OpenAI::FilePart.new(StringIO.new("x"), filename: "a.txt")

OpenAI::Internal::Util.encode_content({"content-type" => "Multipart/Form-Data"}, {file: file})
# => ["Multipart/Form-Data", {file: #<OpenAI::FilePart:0x...>}]
```

The body falls through to the pass-through branch, so no boundary is added and the parts are never encoded. Nothing raises: the request goes out with a content type promising multipart and a body that is the `Hash`'s `to_s`.

The unanchored prefix match was wrong in the other direction too, since `multipart/form-dataX` matched and was encoded as multipart with a boundary appended to that media type.

## Fix

Add `MULTIPART_CONTENT` next to the other two, with the same shape, and match against it:

```ruby
MULTIPART_CONTENT = %r{\Amultipart/form-data[ \t]*(?:;|\z)}i
```

| `content-type` | before | after |
| --- | --- | --- |
| `multipart/form-data` | encoded | encoded |
| `Multipart/Form-Data` | **not encoded** | encoded |
| `MULTIPART/FORM-DATA` | **not encoded** | encoded |
| `multipart/form-data; boundary=zzz` | encoded | encoded |
| `multipart/form-dataX` | **encoded** | not encoded |
| `multipart/mixed` | not encoded | not encoded |

`sig` and `rbi` declare the new constant alongside the existing two.

## Tests

`test_multipart_content_type_matching_ignores_case` asserts the boundary is set and the part is written for all three spellings. `test_multipart_content_type_requires_a_full_media_type` covers the over-match. Both fail on `main`.

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -449,6 +449,8 @@ module OpenAI
       JSON_CONTENT = %r{\Aapplication/(?:[a-z0-9.-]+\+)?json[ \t]*(?:;|\z)}i
       # @type [Regexp]
       JSONL_CONTENT = %r{\Aapplication/(?:x-[nl]djson|(?:x-)?jsonl)[ \t]*(?:;|\z)}i
+      # @type [Regexp]
+      MULTIPART_CONTENT = %r{\Amultipart/form-data[ \t]*(?:;|\z)}i
 
       class << self
         # @api private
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Reviewed with two clean rounds of scope-bounded adversarial review.

PR #673, opened 2026-09-07
fix: honor custom idempotency headers for retries

### LINKED ISSUE #671: Connection-error retry check ignores a custom configured idempotency header name
`BaseClient` lets a subclass configure a custom idempotency header name via `idempotency_header:` on initialization, and correctly writes that header under the configured name when sending requests. But the connection-error retry-safety check only ever looks for the literal header name `"idempotency-key"`, so a subclass using a custom header name loses the safe-retry path it configured.

**Where**

`lib/openai/internal/transport/base_client.rb`:

```ruby
# initialize
@idempotency_header = idempotency_header&.to_s&.downcase
...

# build_request — writes the header correctly under the configured name
if @idempotency_header &&
    !headers.key?(@idempotency_header) && ...
  headers[@idempotency_header] = opts.fetch(:idempotency_key) { generate_idempotency_key }
end
...

# request_retryable_after_connection_error? — only recognizes the hardcoded default name
private def request_retryable_after_connection_error?(request, error:)
  return true if false.equal?(error.request_may_have_been_sent?)
  return true if Net::HTTP::IDEMPOTENT_METHODS_.include?(request.fetch(:method).to_s.upcase)

  request.fetch(:headers).any? do |name, value|
    name.to_s.casecmp?("idempotency-key") && !value.to_s.empty?
  end
end
```

**Scenario**

A `BaseClient` subclass is constructed with e.g. `idempotency_header: "x-request-key"`. A non-idempotent request (e.g. `POST`) is sent with a real key attached under `x-request-key`, and the TCP connection drops after the request bytes may already have reached the server (`request_may_have_been_sent?` is not `false`, and the method isn't idempotent). `request_retryable_after_connection_error?` looks only for the literal `idempotency-key` header, finds nothing, and treats the request as unsafe to retry — even though the whole point of supplying a custom-named idempotency key was to make that retry safe. The connection error is raised immediately instead of the SDK retrying as configured.

`OpenAI::Client` itself never passes `idempotency_header:`, so this is currently dormant for the shipped client — but `BaseClient` is `@api private` reusable transport infrastructure other Stainless-generated clients in this org build on, and the bug is live/reachable for any subclass that does configure a custom header name.

I'll open a PR that has this check read the header name from `idempotency_header` (defaulting to `"idempotency-key"`) instead of hardcoding the literal.

jbeckwith-oai:
Summary:
- use the configured idempotency header when deciding whether an ambiguous connection error is safe to retry
- preserve the existing idempotency-key fallback when no custom header is configured
- add focused coverage for a POST retry with a stable custom header

Tests:
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/internal/transport/base_client_retry_test.rb
- mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/client_test.rb
- mise exec ruby@4.0.6 -- bundle exec rubocop lib/openai/internal/transport/base_client.rb test/openai/internal/transport/base_client_retry_test.rb

Closes #671

### DIFF
--- a/lib/openai/internal/transport/base_client.rb
+++ b/lib/openai/internal/transport/base_client.rb
@@ -512,8 +512,9 @@ module OpenAI
           return true if false.equal?(error.request_may_have_been_sent?)
           return true if Net::HTTP::IDEMPOTENT_METHODS_.include?(request.fetch(:method).to_s.upcase)
 
+          idempotency_header = @idempotency_header || "idempotency-key"
           request.fetch(:headers).any? do |name, value|
-            name.to_s.casecmp?("idempotency-key") && !value.to_s.empty?
+            name.to_s.casecmp?(idempotency_header) && !value.to_s.empty?
           end
         end
 
--- a/test/openai/internal/transport/base_client_retry_test.rb
+++ b/test/openai/internal/transport/base_client_retry_test.rb
@@ -9,6 +9,18 @@ class OpenAI::Test::BaseClientRetryTest < Minitest::Test
     end
   end
 
+  class CustomIdempotencyClient < OpenAI::Internal::Transport::BaseClient
+    def initialize(**kwargs)
+      super(
+        base_url: "https://example.test/v1",
+        idempotency_header: "x-request-key",
+        **kwargs
+      )
+    end
+
+    private def auth_headers(**_context) = {}
+  end
+
   def setup
     super
     @client = RetryClient.new(api_key: "test-key")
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed both changed files, #671, and the surrounding request construction, header normalization, transport error, and retry contracts. The configured header now controls the existing safety predicate; keys remain stable across attempts, and the default idempotency-key fallback and conservative no-key behavior are preserved. No actionable correctness or structural findings.

Independent validation on Ruby 4.0.6: retry, client, and HTTP client tests passed (108 tests, 491 assertions), plus nine request-level probes covering generated/explicit stable keys, casing, empty/nil keys, default fallback, and unsent/idempotent requests. These ran directly with an existing gem cache: the locked Bundler invocation could not start because syntax_tree-rbs was not checked out. Both changed files passed syntax checks; full suite and lint were not run.

PR #676, opened 2026-09-07
fix: preserve bare CR SSE line boundaries

### LINKED ISSUE #670: decode_lines merges a bare-CR-terminated line into the following LF-terminated line
`OpenAI::Internal::Util.decode_lines` (used to split raw HTTP response chunks into lines before SSE parsing) mis-handles a specific line-ending combination: a line terminated with a bare `\r` (not part of `\r\n`) immediately followed by a separate line terminated with a bare `\n`. The two lines get merged into one.

**Where**

`lib/openai/internal/util.rb`, `decode_lines` (around line 796):

```ruby
while (match = re.match(buffer, cr_seen&.to_i || offset))
  case [match.captures.first, cr_seen]
  in ["\r", nil]
    cr_seen = match.end(1)
    next
  in ["\r" | "\r\n", Integer]
    line = buffer.slice!(..(cr_seen.pred))
    line.force_encoding(encoding) unless encoding.nil?
    y << line
  else
    line = buffer.slice!(..(match.end(1).pred))
    line.force_encoding(encoding) unless encoding.nil?
    y << line
  end
  offset = 0
  cr_seen = nil
end
```

The `cr_seen` state exists so a trailing `\r` at the edge of a chunk isn't judged until the next byte is known (needed to correctly detect a `\r\n` pair split across two chunks). Once `cr_seen` is set, the code correctly resolves it when the *next* terminator found is another `\r` or a `\r\n` (the `["\r" | "\r\n", Integer]` branch) — it just re-emits the pending line and lets the following loop iteration reprocess the newly found terminator fresh.

But when the next terminator found is a bare `\n`, it falls into the final `else` branch, which slices the buffer from the very start all the way through that `\n` — silently swallowing the earlier pending `\r` boundary and merging what should be two lines into one.

**Reproduction**

```ruby
OpenAI::Internal::Util.decode_lines(["A\rB\n"]).to_a
# actual:   ["A\rB\n"]        (one line)
# expected: ["A\r", "B\n"]    (two lines, per the SSE line-parsing algorithm this method
#                              implements: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream)
```

This corrupts `decode_sse` (which assumes input has already been correctly split into lines): a server/proxy emitting `id: 1\rdata: hello\n` would have that decoded as a single bogus line `"id: 1\rdata: hello"` instead of two separate SSE fields, silently dropping the `data:` field.

`test/openai/internal/util_test.rb::test_mixed_decode_lines` covers repeated `\r\r` and adjacent CRLF-split-across-chunks cases, but not a bare `\r` followed later by an unrelated `\n`-terminated line, so this slipped through.

I'll open a PR with a fix plus a regression test for this exact case.

jbeckwith-oai:
## Summary
- preserve a pending bare CR boundary before a later LF terminator in `decode_lines`
- keep adjacent CRLF, including CRLF split across chunks, as a single terminator
- add focused regression cases for the reported input and chunk boundaries

## Test plan
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/internal/util_test.rb --name '/test_(decode_lines|mixed_decode_lines)$/'`
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/internal/util_test.rb`
- `git diff --check 821d9b3f018040a3a946f714e120f009541c9743`

## Review
- adversarial review converged after two consecutive clean rounds, each with two fresh independent reviewers
- dedicated SSE parsing safety review found no new limits, logging, dependency, API, buffering-growth, or payload-handling risk

Fixes #670

### DIFF
--- a/lib/openai/internal/util.rb
+++ b/lib/openai/internal/util.rb
@@ -896,7 +896,7 @@ module OpenAI
                 in ["\r", nil]
                   cr_seen = match.end(1)
                   next
-                in ["\r" | "\r\n", Integer]
+                in [terminator, Integer] if terminator != "\n" || match.begin(1) != cr_seen
                   line = buffer.slice!(..(cr_seen.pred))
                   line.force_encoding(encoding) unless encoding.nil?
                   y << line
--- a/test/openai/internal/util_test.rb
+++ b/test/openai/internal/util_test.rb
@@ -982,7 +982,11 @@ class OpenAI::Test::UtilSseTest < Minitest::Test
       %W[\r\r \r] => %W[\r \r \r],
       %W[\r \n] => %W[\r\n],
       %W[\r\r\n] => %W[\r \r\n],
-      %W[\n\r] => %W[\n \r]
+      %W[\n\r] => %W[\n \r],
+      ["A\rB\n"] => ["A\r", "B\n"],
+      ["A\r", "B\n"] => ["A\r", "B\n"],
+      ["A\rB", "\n"] => ["A\r", "B\n"],
+      ["A\r", "\nB\n"] => ["A\r\n", "B\n"]
     }
     cases.each do |enum, expected|
       lines = OpenAI::Internal::Util.decode_lines(enum)


markstuart-oai (approved):
Reviewed both changed files, issue #670, SSE/JSONL callers, encoding and fused-enumerator contracts. No actionable correctness, security, or maintainability findings. The existing pending-CR guard now distinguishes an adjacent LF from a later LF without adding parser states, branches, or helpers. File sizes: util.rb 1072→1072; util_test.rb 1145→1149.

Independent Ruby 4.0.6 validation: 335,924 exhaustive line/chunk cases (including empty chunks), mixed-ending SSE integration across every two-cut partition, UTF-8/binary byte splits and input preservation, JSONL integration, syntax checks, and diff whitespace checks passed. The three existing SSE test methods passed in isolation: 66 assertions. Reproduced the reported failure on the base revision.

Limit: the normal utility test suite and bundled lint could not be validated in this environment; the bundle lacks syntax_tree-rbs, and direct test-helper loading lacks async. Isolated checks loaded the actual unmodified parser; no repository files were edited.

PR #677, opened 2026-09-07
fix: avoid unused Responses stream snapshot copies

jbeckwith-oai:
## Summary

Responses streams built full text and function-argument snapshots for every delta even when callers consumed only .text or requested the final response. These paths now reuse private accumulation buffers and skip unused snapshot construction. A buffer stops being reused as soon as an event exposes it, preserving retained snapshots and wire events.

No public options, model shapes, payload limits, or transport behavior change. Normal .each consumption still materializes the full per-event snapshots that callers request; this optimization does not remove that inherent cost.

## Validation

- Added deterministic public-entrypoint regressions for interleaved text/tool deltas and final-response consumption, plus retained-snapshot and consumption-mode coverage. The performance regressions fail on the original implementation and pass with this change.
- Helper suites, unknown/resumed Responses events, and Responses streaming: **223 tests, 1,405 assertions passed**.
- bundle exec rake lint: passed (RuboCop, formatting, directives, Sorbet and RBS validation).
- Two independent reviewers per round; two consecutive clean adversarial review rounds.

The change stays inside the handwritten Responses helper, its RBI declaration, and focused tests. Buffer ownership was reviewed for input/event aliasing and cleanup; no new parsing, network, authentication, logging, or dependency behavior is introduced.

### DIFF
--- a/lib/openai/helpers/streaming/response_stream.rb
+++ b/lib/openai/helpers/streaming/response_stream.rb
@@ -19,18 +19,26 @@ module OpenAI
         end
 
         def until_done
-          each { |_event| next }
+          text.each { |_delta| nil }
           self
         end
 
         def text
           OpenAI::Internal::Util.chain_fused(@iterator) do |yielder|
+            @text_only = true
             @iterator.each do |event|
+              # Restore normal event snapshots while caller code is running.
+              @text_only = false
               case event
               when OpenAI::Models::Responses::ResponseTextDeltaEvent
                 yielder << event.delta
               end
+
+              @text_only = true
             end
+
+          ensure
+            @text_only = false
           end
         end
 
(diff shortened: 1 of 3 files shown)


markstuart-oai (approved):
Reviewed all three changed files and the surrounding stream/enumerator, model conversion, ownership, consumer, and resumed/unknown-event contracts at f132fe229ec01599d2f6afd3f0e546fb6435a60c against d9a2271531acea12bc3b46293bb90159d3975d59. No actionable correctness or maintainability findings.

The private-buffer ownership set earns its place: exposed snapshots retain their mutable identity, incoming event strings remain isolated, and normal event consumption retains its snapshot behavior. I found no clear behavior-preserving simplification that removes this bookkeeping. No file crosses 1,000 lines (helper 257→284; RBI 111→115; tests 298→429).

Fresh validation on Ruby 4.0.6 with the locked bundle: 223 tests / 1,405 assertions passed; full `bundle exec rake lint` passed. Independent base/head probes matched early-exit, re-entrancy, enumerator-switching, duplication, and snapshot identity behavior. For 2,000 interleaved text/tool deltas, measured explicit prefix-copy bytes fell from 400,200,000 to 200 for text/final consumption; normal `.each` remained unchanged. Both new performance tests fail against the base implementation. These measurements cover accumulator copies, not total process memory or live-network performance. The author's prior review rounds were not used as independent validation.

PR #679, opened 2026-09-07
fix: restore conflict-free exported RBIs for Tapioca

### LINKED ISSUE #678: Exported RBIs conflict since v0.81.0: `tapioca gem openai` produces an empty RBI
### Summary

Since v0.81.0, the RBI files shipped in `rbi/` conflict with each other, so `tapioca gem openai` cannot use them. Combined with the deliberate early `return` in `lib/openai.rb` that skips reflection under `tapioca gem`, the result is a **289-byte, completely empty RBI** — every `OpenAI::*` type disappears from the consuming application.

`rbi/openai/helpers/realtime/extensions.rbi` reopens `OpenAI::Client` **without its superclass**:

```ruby
# rbi/openai/helpers/realtime/extensions.rbi
module OpenAI
  class Client                                              # <- no superclass
```

while `rbi/openai/client.rbi` declares it with one:

```ruby
# rbi/openai/client.rbi
module OpenAI
  class Client < OpenAI::Internal::Transport::BaseClient
```

`RBI::Rewriters::Merge` (used by Tapioca with `Keep::NONE`) treats the differing superclass as an unresolvable conflict.

### What Tapioca does with it

`Tapioca::Gemfile::GemSpec#exported_rbi_tree` merges every `rbi/**/*.rbi` with `Keep::NONE`, and `Tapioca::Commands::AbstractGem#merge_with_exported_rbi` **discards the whole exported tree** as soon as one conflict is found:

```ruby
unless tree.conflicts.empty?
  say_error("\n\n  RBIs exported by `#{gem.name}` contain conflicts and can't be used:", :yellow)
  ...
  return file
end
```

`file` is the reflection-generated tree, which is empty here because `lib/openai.rb` returns early under `tapioca gem`:

```ruby
tapioca_dsl = ARGV.any?(/dsl/) || caller.any?(%r{tapioca[\\/]loaders[\\/]dsl})
if Object.const_defined?(:Tapioca) &&
    caller.chain([$PROGRAM_NAME]).chain(ARGV).any?(/tapioca/) &&
    !tapioca_dsl
  return
end
```

So the two mechanisms cancel each other out: reflection is intentionally disabled *because* the gem ships manifests, and the manifests are then rejected. There is no fallback left.

### Reproduction

```
$ bundle exec tapioca gem openai

    Conflicting definitions for `::Client`
    Found at:
      .../gems/openai-0.85.0/rbi/openai/client.rbi:4:2-311:5
      .../gems/openai-0.85.0/rbi/openai/helpers/realtime/extensions.rbi:4:2-48:5
  Compiled openai (empty output)
      force  sorbet/rbi/gems/openai@0.85.0.rbi

$ wc -c sorbet/rbi/gems/openai@0.85.0.rbi
289 sorbet/rbi/gems/openai@0.85.0.rbi
```

The generated file is Tapioca's "THIS IS AN EMPTY RBI FILE" placeholder. `srb tc` then fails on every use of the SDK:

```
lib/.../openai_adapter.rb:NN: Unable to resolve constant `OpenAI` https://srb.help/5002
```

Standalone repro, no Tapioca needed — only the `rbi` gem and an installed `openai`:

```ruby
require "rbi"

gem_dir = Gem::Specification.find_by_name("openai").full_gem_path
rewriter = RBI::Rewriters::Merge.new(keep: RBI::Rewriters::Merge::Keep::NONE)
Dir.glob("#{gem_dir}/rbi/**/*.rbi").sort.each { |f| rewriter.merge(RBI::Parser.parse_file(f)) }

rewriter.tree.conflicts.each do |c|
  puts c
  puts "  left:  #{c.left.loc}"
  puts "  right: #{c.right.loc}"
end
```

### Affected versions

Merging the full exported tree, per version:

| openai | `rbi/**/*.rbi` | conflicts |
|---|---|---|
| 0.71.0 – 0.80.0 | 1204 – 1226 | 0 |
| 0.83.0 | 1260 | 1 (`::Client`) |
| 0.84.0 | 1260 | 1 |
| 0.85.0 | 1288 | 1 |

`rbi/openai/helpers/realtime/extensions.rbi` was added in aba0f05c9 ("feat: add Realtime WebSocket core", #443) on 2026-08-18, so the first affected release is **v0.81.0** (v0.80.0 predates it). I confirmed the missing superclass at the `v0.81.0`, `v0.82.0` and `v0.85.0` tags, and it is still present on `main`.

`::Client` is the only conflict in the whole tree — the other files reopening shared namespaces (`Errors`, `Resources::Realtime`) declare no superclass on either side, so they merge cleanly.

### Suggested fix

Emit the superclass whenever `OpenAI::Client` is reopened in an exported RBI, i.e. in `rbi/openai/helpers/realtime/extensions.rbi`:

```diff
 module OpenAI
-  class Client
+  class Client < OpenAI::Internal::Transport::BaseClient
```

Since these files are generated, the durable fix is in the generator: any reopened class in the exported RBIs should carry the same superclass as its primary definition (or none of them should).

A regression test would be cheap and would have caught this — merging `rbi/**/*.rbi` with `RBI::Rewriters::Merge::Keep::NONE` and asserting `conflicts.empty?` is the whole check, and it runs in a few seconds.

### Environment

- openai 0.81.0 – 0.85.0 (verified locally on 0.83.0, 0.84.0, 0.85.0)
- tapioca 0.19.1 and 0.19.2 (identical `abstract_gem.rb` / `gemfile.rb`), rbi 0.3.13 / 0.4.0 / 0.4.3, sorbet 0.6.13486
- Ruby 4.0.5, Linux

The conflict reproduces on every combination above, so it is not sensitive to the Tapioca, `rbi` or Sorbet version.

jbeckwith-oai:
## Summary

Fixes #678. Tapioca rejects the entire exported RBI tree when a reopened class disagrees with its primary superclass declaration. Because the SDK intentionally skips reflection during `tapioca gem`, that rejection leaves consumers with an empty RBI.

Declare the existing `OpenAI::Internal::Transport::BaseClient` superclass in both handwritten Client RBI extensions: Realtime and Responses WebSocket. The latter conflict becomes visible after fixing the former. Runtime loading and public APIs are unchanged.

Add a regression that merges every shipped RBI with `RBI::Rewriters::Merge::Keep::NONE`, reporting conflicting file locations. Add `rbi` only to the development/test bundle and update both lockfiles without changing existing dependency versions.

## Validation

- Regression failed before correction; fixing only Realtime exposed the Responses WebSocket conflict.
- Full-tree regression and loader tests: 5 tests, 12 assertions, no failures.
- `bundle exec rake typecheck:sorbet` passed.
- RuboCop and rubyfmt checks passed for all changed Ruby/RBI files.
- Fresh-context adversarial review converged; dependency review found no new transitive dependencies or install hooks. Bedrock release markers are preserved.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ group :development, :test do
   gem "minitest-hooks", ">= 1.5.4"
   gem "minitest-mock", "~> 5.27"
   gem "minitest-proveit"
+  gem "rbi", "~> 0.4.3"
   gem "sorbet-runtime"
   gem "webmock"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -118,6 +118,9 @@ GEM
     rb-fsevent (0.11.2)
     rb-inotify (0.11.1)
       ffi (~> 1.0)
+    rbi (0.4.3)
+      prism (~> 1.0)
+      rbs (>= 4.0.1)
     rbs (4.2.0)
       logger
       prism (>= 1.6.0)
@@ -198,6 +201,7 @@ DEPENDENCIES
   minitest-proveit
   openai!
   rake
+  rbi (~> 0.4.3)
   rbs
   rubocop
   sorbet (< 0.6.12760)
--- a/gemfiles/bedrock.gemfile.lock
+++ b/gemfiles/bedrock.gemfile.lock
@@ -134,6 +134,9 @@ GEM
     rb-fsevent (0.11.2)
     rb-inotify (0.11.1)
       ffi (~> 1.0)
+    rbi (0.4.3)
+      prism (~> 1.0)
+      rbs (>= 4.0.1)
     rbs (4.1.3)
       logger
       prism (>= 1.6.0)
@@ -215,6 +218,7 @@ DEPENDENCIES
   minitest-proveit
   openai!
   rake
+  rbi (~> 0.4.3)
   rbs
   rubocop
   sorbet (< 0.6.12760)
--- a/rbi/openai/helpers/realtime/extensions.rbi
+++ b/rbi/openai/helpers/realtime/extensions.rbi
@@ -1,7 +1,7 @@
 # typed: strong
 
 module OpenAI
-  class Client
+  class Client < OpenAI::Internal::Transport::BaseClient
     # @api private
     sig do
       params(
--- a/rbi/openai/helpers/responses_websocket/extensions.rbi
+++ b/rbi/openai/helpers/responses_websocket/extensions.rbi
@@ -1,7 +1,7 @@
 # typed: strong
 
 module OpenAI
-  class Client
+  class Client < OpenAI::Internal::Transport::BaseClient
     # @api private
     sig do
       params(
--- a/test/openai/exported_rbi_test.rb
+++ b/test/openai/exported_rbi_test.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rbi"
+
+require_relative "test_helper"
+
+class OpenAI::Test::ExportedRBITest < Minitest::Test
+  def test_exported_signatures_merge_without_conflicts
+    paths = Dir.glob(File.expand_path("../../rbi/**/*.rbi", __dir__))
+    refute_empty(paths)
+
+    # Tapioca discards the entire exported tree if this merge has any conflicts.
+    merger = RBI::Rewriters::Merge.new(keep: RBI::Rewriters::Merge::Keep::NONE)
+    paths.each { merger.merge(RBI::Parser.parse_file(_1)) }
+
+    conflicts = merger.tree.conflicts.map { "#{_1}: #{_1.left.loc} and #{_1.right.loc}" }
+    assert_empty(conflicts)
+  end
+end


markstuart-oai (approved):
Reviewed all six changed files and the surrounding Client declarations, extension callers, loader behavior, packaging, and test discovery. Both superclass declarations match the canonical Client contract; keeping these signatures in their feature extensions is the direct fix, without adding runtime branches or another abstraction.

Independently verified that the regression covers all 1,299 shipped RBI files, using the same parser and Keep::NONE merge as Tapioca 0.19.2. The exported-RBI and loader tests pass (5 tests, 12 assertions), as do Sorbet, RuboCop, and rubyfmt. Full-tree negative controls reproduce a Client conflict on base and with either fix omitted; head has zero conflicts and retains all three extension method signatures. The dependency is development/test-only, existing locked versions and Bedrock release markers are preserved, and no changed file crosses 1,000 lines.

No actionable correctness or structural findings. Validation used Ruby 4.0.6 with the frozen primary bundle; I did not run a complete consumer Tapioca CLI invocation or the Bedrock suite.

PR #680, opened 2026-09-08
examples: cover Responses WebSocket workflows and lifecycle

jbeckwith-oai:
## Summary

Ruby 0.86.0 ships Responses WebSocket support, but its README only demonstrates a single turn. Add runnable workflows and a linked guide for sequential tools, interleaved lanes, stored-ID reconnect, full stateless replay, terminal outcomes, and operation deadlines. The examples retain application ownership of recovery and never retry uncertain writes.

Package the guide and workflow so both links work from the installed gem.

Add deterministic loopback tests through the default Responses transport, including UTF-8 fragmentation, premature EOF, idle-read cancellation/cleanup, and a synthetic event exceeding 32 MiB. No SDK runtime, public API, dependency, schema, or compiler behavior changes.

## Validation

- Complete Responses WebSocket suite plus shared transport tests: 75 tests, 363 assertions; no failures/errors/skips, using Ruby 4.0.6 and the locked bundle.
- Deadline mutation check: removing the 100 ms operation timeout now fails the monotonic bound at the outer 10-second harness timeout; the normal workflow suite passes in about 0.3 seconds.
- Gem packaging regression tests: 2 tests, 40 assertions; all README-relative links resolve in the built gem.
- RuboCop, rubyfmt, example inventory, and diff whitespace checks pass.
- Custom-code budget: 3,451 / 4,000 lines (549 remaining); no policy changes.
- Adversarial review: seven rounds total, with two fresh consecutive clean rounds after the deadline-test correction; security-sensitive example/transport behavior reviewed.
- Temporary schema-field generation probe: two clean three-way generation passes, model/RBI/RBS propagation, unique helper signature blocks, no helper loss, and 4,309 identical files between passes (excluding generation metadata). The same 75-test suite passes on the regenerated copy.
- No live API calls performed locally. The example is registered in the existing live-example inventory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,10 @@ does not automatically reconnect or replay an ambiguous write. When the server
 closes a connection, open a new connection and continue with
 `previous_response_id` when the response was stored.
 
+See the [Responses WebSocket guide](responses-websocket.md) and
+[runnable workflows](examples/responses/websocket_workflows.rb) for tool turns,
+multiplexing, stored continuation, stateless replay, and operation deadlines.
+
 ### Pagination
 
 List methods in the OpenAI API are paginated.
--- a/test/openai/responses_websocket/workflows_test.rb
+++ b/test/openai/responses_websocket/workflows_test.rb
@@ -241,6 +241,7 @@ class OpenAI::Test::ResponsesWebSocketWorkflowsTest < Minitest::Test
       end
     end
 
+    started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
     with_server(handler) do |client|
       client.responses.connect do |connection|
         %w[planner critic].each do |lane|
@@ -253,6 +254,13 @@ class OpenAI::Test::ResponsesWebSocketWorkflowsTest < Minitest::Test
             Workflows.completed_responses(connection, lanes: %w[planner critic])
           end
         end
+
+        assert_operator(
+          Process.clock_gettime(Process::CLOCK_MONOTONIC) - started_at,
+          :<,
+          5,
+          "The operation deadline must fire before the 10-second harness timeout."
+        )
       end
 
       assert(peer_closed.dequeue)
--- a/test/openai/responses_websocket/workflows_test.rb
+++ b/test/openai/responses_websocket/workflows_test.rb
@@ -1,0 +1,315 @@
+# frozen_string_literal: true
+
+require "async/http/server"
+require "async/queue"
+require "async/websocket/adapters/http"
+require "async/websocket/server"
+require "io/endpoint/bound_endpoint"
+require "stringio"
+
+require_relative "../test_helper"
+require_relative "../../../examples/responses/websocket_workflows"
+
+class OpenAI::Test::ResponsesWebSocketWorkflowsTest < Minitest::Test
+  extend Minitest::Serial
+
+  Workflows = OpenAI::Examples::ResponsesWebSocketWorkflows
+
+  # Cross 32 MiB without imposing an API maximum. Keep this probe large and
+  # sequential; never shrink it or raise client limits merely to pass CI.
+  PAYLOAD_BYTES = (32 * 1024 * 1024) + 1
+
+  def test_multiplex_example_waits_for_both_interleaved_lanes
+    handler = lambda do |socket|
+      creates = [read_event(socket), read_event(socket)]
+      assert_equal(%w[planner critic], creates.map { |event| event.fetch("stream_id") })
+      assert_equal(["response.create"], creates.map { |event| event.fetch("type") }.uniq)
+      write_event(socket, type: "response.future_event", stream_id: "critic", extra: "new field")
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -112,6 +112,10 @@ does not automatically reconnect or replay an ambiguous write. When the server
 closes a connection, open a new connection and continue with
 `previous_response_id` when the response was stored.
 
+See the [Responses WebSocket guide](responses-websocket.md) and
+[runnable workflows](examples/responses/websocket_workflows.rb) for tool turns,
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 9e93f2ffea8d7c82e63e5c45597d8268651df7db. Added the workflow to the gemspec packaged examples and the guide to extra_rdoc_files. GemPackagingTest now passes (2 tests, 40 assertions), including all README-relative links in the built gem. RuboCop, rubyfmt, and the custom-code budget pass; two fresh independent review rounds are clean. The fix is pushed and exact-head CI is running.

test/openai/responses_websocket/workflows_test.rb
```diff
@@ -0,0 +1,315 @@
...
+
+    with_server(handler) do |client|
+      client.responses.connect do |connection|
+        %w[planner critic].each do |lane|
+          connection.response.create(model: "example-model", input: "Hello", stream_id: lane)
+        end
+
+        ready.dequeue
+        assert_raises(OpenAI::Errors::ResponsesConnectionError) do
+          Async::Task.current.with_timeout(0.1) do
```
markstuart-oai, line 253: [P3] Distinguish the operation deadline from the harness timeout

This assertion also accepts the outer `with_server` 10-second timeout: the default transport wraps both timeouts as `ResponsesConnectionError`. I replaced only `Async::Task.current.with_timeout(0.1) do` with `begin` in an in-memory copy, and this test still passed (1 test, 3 assertions, about 10.05 seconds). Please add a monotonic elapsed-time assertion with generous slack below the harness timeout, or independently detect harness expiry, so deleting/breaking the operation deadline makes the test fail while preserving the cleanup/no-replay checks.

markstuart-oai (approved):
Reviewed all six files at 9e93f2ffea8d7c82e63e5c45597d8268651df7db against actual base 74584a053845e5b7184480d5e9d297201cb9743c, including the unchanged Responses/shared transport lifecycle, credential handshake, neighboring examples, inventory and packaging contracts. Recovery stays application-owned, uncertain writes are not replayed, and there is no runtime/API/dependency/schema/compiler change. The small shared completion reader and reconnect example keep the structure proportionate; no file crosses 1,000 lines.

Reproduced 75 tests / 361 assertions for the complete Responses WebSocket and shared transport suites, plus the built-gem README-link test (1 test / 32 assertions), inventory, scoped RuboCop/rubyfmt, and diff whitespace checks. The earlier packaging finding is fixed. The separate installed-gem X.509 smoke subprocess is blocked by this environment's relocated Ruby load path; I did not reproduce generation, the budget measurement, live examples, or the author's review rounds.

One non-blocking test-strengthening comment: removing the inner operation deadline in an in-memory negative control still lets the test pass after the harness's 10-second timeout. No author files were modified.

test/openai/responses_websocket/workflows_test.rb
jbeckwith-oai: Fixed in 0a19100e7b07e2569f7f114d9f2f5196978374f8. Added a monotonic five-second elapsed bound measured before the harness starts, giving the 100 ms operation deadline generous slack while rejecting the outer ten-second timeout. Verified your mutation: the old test passes after 10.00s without the operation timeout; the corrected test fails at the new assertion after 10.00s. Normal workflow tests pass (10/64); the complete focused suite passes (75/363). Lint, format, budget, and two fresh independent review rounds pass. Pushed; monitoring exact-head CI.

markstuart-oai (approved):
R2 review at 0a19100e7b07e2569f7f114d9f2f5196978374f8 against base 74584a053845e5b7184480d5e9d297201cb9743c, following my prior review 5136291889 and the author response 3953512871.

The deadline-test finding (3953490759) is fixed. I reproduced the same in-memory mutation removing the 100 ms operation timeout: it now fails the new monotonic assertion at 10.07 seconds. The five-second bound gives generous slack while distinguishing the ten-second harness timeout, without introducing another helper or mode.

Fresh validation: complete Responses WebSocket/shared transport suite passes (75 tests, 363 assertions); built-gem README-relative-link test passes (1 test, 32 assertions); example inventory, scoped RuboCop/rubyfmt, and diff whitespace checks pass. Inspected every new line and confirmed the remaining five PR files and previously reviewed dependencies are unchanged. Recovery, lifecycle, credential handling and packaging conclusions remain sound; no new findings or structural regressions.

No live calls, generation, budget measurement, or full-repository suite were run. The separate installed-gem X.509 smoke remains unverified because its isolated subprocess clears the load path required by this environment's relocated Ruby. Author sources were not modified.

PR #684, opened 2026-09-08
ci: shard Ruby tests across parallel jobs

jbeckwith-oai:
The Ruby test matrix determines CI completion: PR #683 spent 6m47s–8m47s in each `Run tests` step while all other CI jobs finished within 53 seconds. Split the primary test suite into three disjoint groups of whole files on each of Ruby 3.3, 3.4, and 4.0. Every existing test still runs on every version, and existing serial scheduling—including large-payload cases—is preserved.

`TEST_SHARD=1/3 ./scripts/test` reproduces a group locally; omitting `TEST_SHARD` still runs the complete primary suite. Invalid configurations and empty groups fail visibly. The existing `ci-required` gate requires all nine matrix jobs to succeed. Permissions, action pins, Bedrock testing, and the 15-minute timeout from #683 are unchanged.

### Validation

- Local Ruby 4.0.6 baseline: 1,872 tests in 254s; zero failures/errors, one platform skip.
- Three local shards: 555 / 671 / 650 tests in 22s / 93s / 143s; zero failures/errors and the same platform skip. Combined count is 1,876, including four new sharding tests.
- Regression coverage verifies exhaustive/disjoint coverage against the repository inventory, all supported Rubies, discovery-order independence, invalid configuration, and empty-selection failure.
- Full lint/formatting, Sorbet, and RBS validation passed; custom-code budget passed at 3,451 / 4,000 lines.
- Independent adversarial review and CI security review completed before pushing. Hosted CI timings will verify the actual wall-time benefit; local measurements are diagnostic and runner performance varies.

Tradeoff: six additional runner jobs add setup overhead and consume more concurrency. This reduces feedback latency while preserving coverage; it does not eliminate the repeated whole-tree typechecking and RBI-merge work identified during profiling.

PR #683, opened 2026-09-08
ci: allow 15 minutes for Ruby test jobs

apcha-oai:
Ruby test jobs can run close to the current 10-minute limit, leaving too little headroom for setup and runner variance. Increase the `test-ruby` matrix job timeout to 15 minutes as an interim measure while test-suite performance is investigated separately.

This applies to Ruby 3.3, 3.4, and 4.0. Other job timeouts and test coverage are unchanged.

Validation: reviewed the one-line workflow diff and ran `git diff --check`.

PR #659, opened 2026-09-06
fix: declare built-in workload identity providers

jbeckwith-oai:
## Problem

The README documents three built-in workload identity subject-token providers, but the shipped Sorbet and RBS graphs did not declare them. Static users could not configure Kubernetes, Azure, or GCP workload identity through the existing `SubjectTokenProvider` contract even though the runtime classes already worked.

## User-facing fix

```ruby
provider = OpenAI::Auth::SubjectTokenProviders::K8sServiceAccountTokenProvider.new
identity = OpenAI::Auth::WorkloadIdentity.new(
  provider: provider,
  identity_provider_id: "ip_synthetic",
  service_account_id: "sa_synthetic"
)
```

The same typed configuration now works with `AzureManagedIdentityTokenProvider` and `GCPIDTokenProvider`. The new RBI and RBS files declare the existing constructor keywords/default shapes, interface inclusion, and `token_type` / `get_token` contracts.

## Synthetic before/after

- Before: an offline Sorbet probe reported three unresolved `OpenAI::Auth::SubjectTokenProviders` constants; RBS could not resolve the Kubernetes provider.
- After: focused shipped-declaration probes accept all three providers flowing into `WorkloadIdentity`, and negative controls still reject `token_path: 123`.

## Scope and compatibility

- Adds only `rbi/openai/auth/subject_token_providers.rbi`, `sig/openai/auth/subject_token_providers.rbs`, and one focused type test.
- No runtime authentication, token exchange, transport, error, dependency, generated-model, or public API behavior changed.
- Tests use synthetic identifiers, do not call `get_token`, and do not read token files or contact metadata services.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/auth/subject_token_provider_types_test.rb` — 5 runs, 19 assertions, 0 failures, 0 errors.
- `mise exec ruby@4.0.6 -- bundle exec rake lint` — RuboCop 2,925 files, RBS validation, and Sorbet example typecheck passed.
- `TEST_API_BASE_URL=http://127.0.0.1:49137 mise exec ruby@4.0.6 -- bundle exec rake test` — 1,777 runs, 15,259 assertions, 0 failures, 0 errors, 1 skip.
- Trusted freshly fetched-main public custom-code budget check — isolation success; 3,449 / 4,000 custom lines.
- Adversarial review — two consecutive clean rounds, each with two new independent read-only reviewers.
[...]

PR #658, opened 2026-09-06
fix: declare workload identity error types

jbeckwith-oai:
## Problem

The runtime already raises `OpenAI::Errors::SubjectTokenProviderError` and `OpenAI::Errors::OAuthError` during workload-identity authentication, but the shipped RBI and RBS omitted both classes. Typed applications could hit unresolved-constant errors when rescuing these supported failures, and could not type-check reads of their existing metadata.

## User-facing example

```ruby
begin
  client.models.list
rescue OpenAI::Errors::SubjectTokenProviderError => error
  warn "#{error.provider}: #{error.cause&.message}"
rescue OpenAI::Errors::OAuthError => error
  warn "OAuth #{error.status}: #{error.error_code}"
end
```

## Change

- Declares the two existing runtime classes in `rbi/openai/errors.rbi` and `sig/openai/errors.rbs` with their existing inheritance and keyword constructors.
- Declares `provider`, nullable `cause`, `error_code`, and the concrete OAuth status reader using the existing OAuth error-code union.
- Adds one focused offline regression that exercises public `Client(...workload_identity...).models.list` failures and positive/negative checks against the shipped Sorbet and RBS declarations.

Synthetic before/after:

- Before: the Sorbet probe reports unresolved `SubjectTokenProviderError` and `OAuthError` constants plus missing `provider`/`error_code`; RBS reports both constants missing.
- After: the same positive probes pass; deliberate `status` type mismatches remain rejected. Offline runtime probes observe `provider == "kubernetes"` with an `Errno::ENOENT` cause, and a WebMock 401 exchange with `status == 401` and `error_code == :invalid_grant`.

## Compatibility and scope

This is a type-only correction: runtime error classes, authentication behavior, providers, transports, and public APIs are unchanged. It intentionally does not add the separately scoped provider declarations or perform broader auth-type cleanup. All test values are synthetic and live network access is disabled.

Security assessment: the diff only declares existing authentication error types and adds offline synthetic tests; it changes no credential acquisition, token exchange, transport, logging, dependencies, CI, or release behavior. `@openai/sdks-team` review is requested because the declarations describe authentication-facing errors.

## Verification

- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/auth/workload_identity_error_types_test.rb` — 6 runs, 18 assertions, 0 failures
- `mise exec ruby@4.0.6 -- bundle exec ruby -Itest test/openai/auth/workload_identity_test.rb` — 30 runs, 130 assertions, 0 failures
[...]

PR #686, opened 2026-09-08
ci: reduce test overhead and unshard to conserve runners and reduce flakes

apcha-oai:
Ruby CI repeatedly copied and loaded the full RBS signature tree for ten tiny consumer checks, and a mock-process startup race could add a 60-second wait. This change batches the compatibility fixtures, fixes the mock handshake, and returns the Ruby matrix to one shard per version for simplicity and fewer opportunities for flakes from repeated job setup.

## Why return to one shard

Unsharding is intentional. Three-way sharding brought elapsed CI time back toward baseline, but multiplied the Ruby matrix from three jobs to nine. That consumes more runners and creates more opportunities for flakes through additional dependency installation, mock startup, and cleanup. We prefer to remove redundant test work and keep one complete-suite job per Ruby version; sharding should be a last resort if further test improvements cannot provide acceptable feedback time.

This is a deliberate tradeoff: individual unsharded jobs take longer than shards, while CI uses fewer runners and has fewer independent setup/cleanup paths. The current sequential-RBI implementation completes public jobs in 3m59s–4m43s and internal jobs in 5m11s–5m35s. The original under-five-minute target is not met on the internal runners, and we intentionally retain one shard rather than restore nine jobs to meet it.

## Changes

- Check all ten positive/negative RBS consumers in one Steep invocation. Each stays a separate source file; exact negative diagnostics are checked with `--with-expectations`. Runtime and declaration assertions remain in their original suites.
- Publish the waiting mock's PID from initialized Ruby and require the interrupted launcher to exit within five seconds.
- Run one complete-suite shard for each supported Ruby (3.3, 3.4, 4.0). Keep the sharding helper and verify that the configured shard covers every test file exactly once.
- Preserve Tapioca's canonical sequential RBI merge and add the mixed typed/untyped counterexample from [review](https://github.com/openai/openai-ruby/pull/686#discussion_r3960338093). Pairwise merging was withdrawn because the operation is not associative: a later typed declaration can hide an earlier method conflict when files are regrouped.

Rebased onto main `31bbd60de2775f90f46dcaddda8089d5e322bc5b`. The upstream 15-minute timeout from #683 is unchanged; this PR changes the three-way sharding added in #684. No SDK runtime, signature, dependency, or generated-surface changes. Large-payload tests retain their original sizes and assertions.

## Root cause and remaining cost

- Ten separate Steep subprocesses accounted for about 205 seconds in the internal Ruby 3.4 baseline profile. They were introduced incrementally in #554, #642, #654, #657, #658, and #659. The consolidated check measured 8.7 seconds locally and 22–24 seconds in the hosted RBS-only experiment.
- The exported RBI test from #679 repeatedly indexes growing merged output across about 1,300 files. Hosted profiling measured 97–135 seconds. **That cost remains:** this PR retains the ordered merge for correctness.
- The mock fixture published its PID before Ruby initialized. A TERM during interpreter startup could be lost; one profiled run waited 60.4 seconds for natural exit. Publishing the PID from the initialized process closes that race.
- Cached bundle setup took about 4 seconds and mock startup about 8 seconds in a representative baseline job; test loading was 1.5–1.8 seconds. Dependency setup was not the dominant regression in those runs.

Internal Ruby 3.4 test execution grew from 2.9 seconds on July 15 (460 tests), to 31.1 seconds on August 14 (723 tests), to 498.4 seconds on September 8 (1,872 tests). Expanded coverage explains some growth; the measured overhead above identifies specific additional costs.

## Validation

- The review's four-file mixed typed/untyped regression fails against the pairwise helper (expected one conflict, got zero), then passes with sequential merging.
- Both exported-RBI tests pass locally: 2 tests, 6 assertions, 50.8 seconds. Changed-file RuboCop, rubyfmt, and whitespace checks pass.
- The one-shard workflow regression passes: 4 tests, 23 assertions.
- Removing one expected negative diagnostic and introducing one error in a valid fixture makes the consolidated RBS check fail with exactly one missing and one unexpected diagnostic.
- All required checks pass at `c03d0ed004366b0e283581a74bacf9c7a75b6dbc` in both repositories. Each complete-suite job ran 1,868 tests with no failures, errors, or skips. Earlier 2m39s–4m04s timings used the withdrawn pairwise optimization and do not describe this revision.

## Current hosted validation

Whole-job times include setup and cleanup; every row uses one shard.

| Repository | Ruby | Whole job | Test execution |
| --- | --- | --- | --- |
| Internal | 3.3 | 5m35s | 312.8s |
| Internal | 3.4 | 5m34s | 309.3s |
| Internal | 4.0 | 5m11s | 289.9s |
| Public | 3.3 | 3m59s | 215.9s |
| Public | 3.4 | 4m43s | 260.6s |
| Public | 4.0 | 4m29s | 244.1s |

Runs: [internal](https://github.com/openai/openai-ruby-internal/actions/runs/34255090430), [public](https://github.com/openai/openai-ruby/actions/runs/34255090446).

## Measurement sources

- Baseline main `21b8a70d48e2623e32aa01f95503adcf3fa2ba75`: [internal](https://github.com/openai/openai-ruby-internal/actions/runs/34188687079), [public](https://github.com/openai/openai-ruby/actions/runs/34188680416). Whole Ruby jobs were 8m38s–9m26s internally and 6m49s–7m12s publicly.
- Attribution: [baseline per-test/subprocess profile](https://github.com/openai/openai-ruby-internal/actions/runs/34245674742), [RBS consolidation only](https://github.com/openai/openai-ruby-internal/actions/runs/34251931166).
- Historical Ruby 3.4: [July 15](https://github.com/openai/openai-ruby-internal/actions/runs/29431093250/job/87405969955), [August 14](https://github.com/openai/openai-ruby-internal/actions/runs/31843976362/job/94906550271).

SDK-team review is requested. Measurements use each repository's normal runner selection and are reported separately because observed runtimes differ.

Linked counterpart: https://github.com/openai/openai-ruby-internal/pull/97


.github/workflows/ci-checks.yml
```diff
@@ -99,17 +99,17 @@ jobs:
...
+    name: test (ruby ${{ matrix.ruby-version }}, shard ${{ matrix.shard }}/1)
     permissions:
       contents: read
     runs-on: ${{ inputs.runner }}
     strategy:
       fail-fast: false
       matrix:
         ruby-version: ['3.3', '3.4', '4.0']
-        shard: [1, 2, 3]
+        shard: [1]
```
jbeckwith-oai, line 110: Blocking: this exact head changes main’s newly merged `shard: [1, 2, 3]` matrix to `[1]` (and `/3` to `/1`), undoing #684 and increasing the suite’s critical-path wall time. That directly conflicts with the PR description’s “no ... job-selection changes” statement and its claim that the current PR uses three-way sharding; the cited final checks at `9db65bc` are not this head. Please preserve the three-way matrix, or explicitly make unsharding part of the approved scope, update the description, and provide exact-current-head public/internal validation for that tradeoff.

jbeckwith-oai (changes requested):
Reviewed the complete exact-head 23-file diff against the test-efficiency outcome and the stated non-goals. The consolidated RBS harness preserves the ten separate consumer files and exact negative diagnostics, the pairwise RBI merge retains within-pair, cross-pair, and odd-tail conflict coverage, and the initialized-Ruby PID publication closes the lost-TERM fixture race with a bounded launcher wait. `git diff --check` passes, no runtime/signature/dependency/generated-surface changes are present, and all completed exact-head lint/type/package/RBS/Bedrock/CodeQL/Castiron checks are green. One blocking scope/CI regression remains: this head undoes the just-merged three-way Ruby test matrix even though the PR description says job selection is unchanged and cites sharded validation from an older head. Restore the three-way matrix, or explicitly re-scope and revalidate the unsharding decision on the current head.

.github/workflows/ci-checks.yml
apcha-oai: intentional; we should avoid sharding as that introduces more flakes (e.g. GH network flake issues; etc.)

markstuart-oai (commented):
Reviewed all 23 changed files against main/base `31bbd60d`, including the surrounding RBS/RBI and launcher contracts. One substantive issue remains: the balanced RBI merge is not equivalent to the original ordered merge and can silently lose a conflict (reproduction inline). Collecting every level's conflict metadata does not fix conflicts erased before that level is merged.

The RBS consolidation otherwise removes duplicated setup cleanly: all ten consumer sources remain separate, and an independent mutation produced exactly one missing and one unexpected diagnostic. Both original Client superclass regressions are still detected. No file crosses 1,000 lines, and no SDK runtime/signature/dependency/generated-surface changes are present.

Fresh local Ruby 4.0.6 validation: 23 focused tests / 115 assertions passed; the interrupt test passed (1 test / 3 assertions). The broader launcher run had one readiness-timeout assertion failure, so I am not claiming a clean full launcher suite or full-suite/older-Ruby/platform validation.

I also verified the three-to-one shard change. The author has now explicitly documented its rationale and separated current-head validation from historical timing evidence. The optimized consumer/RBI/launcher code matches `9db65bc`; control `da9304781` differs from that historical head only by workflow dispatch. Historical timings are not fresh measurements of this review.

test/openai/exported_rbi_test.rb
```diff
@@ -10,10 +10,41 @@ def test_exported_signatures_merge_without_conflicts
...
+
+  def merge_conflicts(trees)
+    conflicts = []
+    # Each merge indexes its entire accumulated output. A left fold over all
+    # 1,300 files repeatedly indexes growing prefixes; adjacent pairs keep the
+    # same file order while limiting that work to logarithmically many levels.
+    while trees.size > 1
+      trees = trees.each_slice(2).map do |pair|
+        merger = RBI::Rewriters::Merge.new(keep: RBI::Rewriters::Merge::Keep::NONE)
+        pair.each { merger.merge(_1) }
```
markstuart-oai: [P2] Preserve ordered merge conflict semantics

`RBI::Rewriters::Merge` is not associative, so preserving file order while changing the grouping can make this regression test pass for an export the original merge rejects. With locked RBI 0.4.3, parse these as four separate trees in order:

```ruby
class A; def m(x); end; end
class A; def m(x); end; end
class A; def m(y); end; end
class A; sig {params(x: String).void}; def m(x); end; end
```

The previous single merger records `Conflicting definitions for ::A#m(x)` when it reaches tree 3; this helper returns `[]`. In the second pair, merging the typed declaration into the untyped method replaces its parameters/signature, erasing the conflict before the two pairs meet. Saving intermediate conflict arrays cannot recover a conflict that was never emitted.

Please retain the canonical ordered merge semantics (the sequential merge is a safe fallback) and add this mixed typed/untyped regression before introducing an equivalent optimization. The superclass-only probe does not establish equivalence for method declarations.

jbeckwith-oai (approved):

jbeckwith-oai (changes requested):
Follow-up exact-head review found one remaining in-scope documentation regression. This PR intentionally changes CI from three shards per Ruby to one complete-suite job, but CONTRIBUTING.md still states that CI partitions the suite into three groups and tells contributors to reproduce CI with TEST_SHARD=1/3, 2/3, or 3/3. That guidance was added by the base commit being reversed here, so it becomes false at this head. Please update the contributor documentation to describe the one-shard CI behavior while retaining the helper optional local sharding instructions if desired. The current title/body, sequential RBI fix, consolidated RBS fixtures, mock-handshake change, exact-head CI, and 23-file patch otherwise match the approved scope. Local focused execution was unavailable because this checkout lacks the locked bundle; exact-head required CI is green.

.github/workflows/ci-checks.yml
jbeckwith-oai: Agreed this is good and should stay

jbeckwith-oai (approved):
Agent version of me, please don't request changes

.github/workflows/ci-checks.yml
markstuart-oai, line 110: [P3] Align contributor instructions with the intentional one-shard workflow

The one-shard tradeoff is now explicit and the configured inventory is complete, but `CONTRIBUTING.md:152–156` still says CI partitions each Ruby suite into three groups and presents `TEST_SHARD=1/3`, `2/3`, and `3/3` as reproducing CI. Please describe the complete-suite CI job there and label those commands as optional local sharding. The helper can remain available without documenting it as the current CI configuration.

markstuart-oai (approved):
Follow-up to review 5144715509 and the explicitly requested re-review: the RBI finding is fixed. The exact four-file counterexample now runs through the same canonical sequential merger as the exported-tree check; the pairwise loop and intermediate-conflict bookkeeping are deleted. This restores the required semantics with a simpler implementation.

Reconciled all 23 changed files against main/base `31bbd60d`; only the RBI test changed since my prior review. Fresh Ruby 4.0.6 checks passed: 7 tests / 32 assertions across exported RBI, consolidated RBS, and shard coverage. Independent probes also confirmed that the previous helper misses the exact counterexample, the current helper reports it, and restoring either Client superclass omission still produces a conflict. Inventory confirms all 251 non-Bedrock test files, including the unchanged large-payload suite, run in each configured 1/1 job. No file crosses 1,000 lines.

The intentional one-job-per-Ruby tradeoff and unchanged timeout are now explicit. Public run 34255090446 is successful at this head (3m59s–4m43s); I also checked its Ruby 4.0 log: 1,868 tests, zero failures/errors/skips. Withdrawn pairwise timings are not evidence for this revision.

Earlier diagnostic-mutation and interrupt checks remain applicable to unchanged code. I did not rerun the full local suite, older Ruby versions, or other platforms; the earlier broader-launcher readiness-timeout failure remains a validation limit. One minor contributor-documentation correction remains inline.

PR #687, opened 2026-09-08
ci: publish Castiron statuses despite stale PR base metadata

jbeckwith-oai:
## Summary

GitHub can retain an older `base.sha` on an open PR after main advances. The Castiron handler treated that metadata as PR identity, so rerunning a valid PR could leave both required budget statuses missing while the publisher appeared green.

Resolve the current PR by its verified repository, branch, and head identity. Continue evaluating current-main policy against trusted Git objects and checking the actual evaluation revisions before publication. Publish failing statuses and fail the publisher job when evaluation is unavailable, stale, or fails. Recheck main before writing the report comment.

## Validation

- Castiron suite passed: 55 tests, with the optional compiler contract test skipped (`CASTIRON_TEST_BIN` is unset), including stale PR base metadata, stale heads, evaluation base changes, failure visibility, and comment publication freshness.
- Ruff, actionlint for both Castiron workflows, and `git diff --check` passed.
- Security review and two consecutive clean adversarial-review rounds, with two independent fresh-context reviewers per round.

The trusted workflow runs from main, so the updated publisher takes effect after this PR merges. Budget policy, accounting, and permissions are unchanged.

### DIFF
--- a/.github/workflows/castiron-custom-code-comment.yml
+++ b/.github/workflows/castiron-custom-code-comment.yml
@@ -157,7 +157,7 @@ jobs:
               for (const number of [...new Set(pulls.map(pull => pull.number))].sort((a, b) => a - b)) {
                 if (!Number.isInteger(number) || number <= 0) return;
                 const {data: pr} = await github.rest.pulls.get({...context.repo, pull_number: number});
-                if (pr.state === 'open' && pr.head.sha === head && pr.base.sha === base &&
+                if (pr.state === 'open' && pr.head.sha === head &&
                     pr.head.ref === run.head_branch && pr.head.repo?.id === headRepository.id &&
                     pr.head.repo?.full_name === headRepository.full_name &&
                     pr.base.ref === branch &&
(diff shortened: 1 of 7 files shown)


markstuart-oai (approved):
Reviewed all seven changed files at fe571bae6877032015c578161073dabf5e772e0c, including the surrounding trusted evaluator, Git-object verification, publisher workflows, and tests. No substantive findings.

The change cleanly separates PR identity from evaluation freshness: repository/ref/head checks remain exact, while the budget evaluator and publication checks use current main independently of PR base metadata. Missing, stale, and failed evaluations produce failing statuses and a visible publisher failure; report comments recheck main after pagination. The existing resolver owns the change, with no new modes or helper layers. Production scripts plus workflows grow by only 3 lines / 240 bytes, and no file crosses 1,000 lines.

Validation: the full Castiron suite ran 55 tests successfully with one optional compiler-contract skip (CASTIRON_TEST_BIN unset); six additional isolated probes passed, covering the stale-metadata status matrix, identity mismatches, a real newer-main budget decrease, comment retry idempotency, replacement-ref isolation, and mocked status transport failures. Reporter checksum and git diff --check passed; observed PR checks were green. Local Ruff was blocked by the package-registry proxy, and actionlint was unavailable. No live publisher or workflow was invoked.

PR #689, opened 2026-09-09
feat(api): Add service account key expiration support

apcha-oai:
## Summary

Allow project service account API keys to be created with an expiration period and expose their expiration timestamps in returned models.

## Changes

- Add the optional `expires_in_seconds` parameter when creating a project service account.
- Expose the optional nullable `expires_at` field on project API keys.
- Expose the optional nullable `expires_at` field on API keys returned when creating project service accounts.


lib/openai/models/responses/response_function_web_search.rb
```diff
@@ -30,20 +30,22 @@ class ResponseFunctionWebSearch < OpenAI::Internal::Type::BaseModel
...
-        #   more information.
+        #   [web search guide](https://developers.openai.com/api/docs/guides/tools-web-search)
+        #   for more information.
         #
-        #   @param id [String] The unique ID of the web search tool call.
+        #   @param id [String]
+        #     The unique ID of the web search tool call.
         #
-        #   @param action [OpenAI::Models::Responses::ResponseFunctionWebSearch::Action::Search, OpenAI::Models::Responses::ResponseFunctionWebSearch::Action::OpenPage, OpenAI::Models::Responses::ResponseFunctionWebSearch::Action::Find] An object describing the specific action taken in this web search call.
+        #   @param action [OpenAI::Models::Responses::ResponseFunctionWebSearch::Action::Search, OpenAI::Models::Responses::ResponseFunctionWebSearch::Action::OpenPage, OpenAI::Models::Responses::ResponseFunctionWebSearch::Action::FindInPage]
```
markstuart-oai: [P3] Keep the documented action type aligned with the public class

This changes the constructor documentation to `Action::FindInPage`, but the runtime union still resolves `find_in_page` to `Action::Find` (line 66), and both RBI and RBS expose `Find`. `Action.const_defined?(:FindInPage, false)` is false. Please preserve `Action::Find` in the generated parameter documentation so editor guidance does not point callers to a nonexistent constant.

lib/openai/models/admin/organization/projects/service_account_create_params.rb
```diff
@@ -27,12 +27,31 @@ class ServiceAccountCreateParams < OpenAI::Internal::Type::BaseModel
...
-            # @!method initialize(project_id:, name:, create_service_account_only: nil, request_options: {})
+            # @!attribute expires_in_seconds
+            #   Number of seconds until the initial API key expires. If omitted or null, the key
+            #   does not expire unless the effective organization or project policy requires an
+            #   expiration. When a policy sets a maximum lifetime, this value must be provided
+            #   and must not exceed that limit. A non-null value cannot be used when
+            #   `create_service_account_only` is true.
+            #
+            #   @return [Integer, nil]
+            optional :expires_in_seconds, Integer, nil?: true
```
markstuart-oai, line 38: [P3] Cover expiration on the initial service-account creation path

Please extend `service_account_api_key_expiration_test.rb` with a mocked `service_accounts.create` request that asserts `expires_in_seconds` reaches the JSON body and the nested `response.api_key.expires_at` survives decoding, including omitted/null values. The existing focused test exercises `service_accounts.api_keys.create`, while the changed resource tests only cover project-key reads. I verified the new path works with offline probes; keeping that coverage in the suite would protect these newly exposed fields and satisfy the repository's focused serialization-test guidance.

markstuart-oai (approved):
Reviewed the expiration models, request serialization, response decoding, type signatures, and generated documentation changes. The new fields use the existing conversion path without adding special-case policy logic. Offline probes passed for omitted/null/bounded lifetimes and omitted/null/integer expiration responses. Two small follow-ups below: retain regression coverage for the initial-create path and correct a generated documentation type name.

Eight model files cross 1,000 lines through documentation expansion alone: beta/beta_tool, evals/run_cancel_response, evals/run_create_response, evals/run_list_response, evals/run_retrieve_response, responses/response_create_params, responses/responses_client_event, and responses/tool. Their non-comment executable tokens are unchanged, and the model and constructor help remains organized around existing fields and types. I am waiving decomposition for these specific crossings: splitting runtime types solely to relocate this documentation would add indirection without removing concepts.

jbeckwith-oai (dismissed):
Two in-scope blockers remain on this exact head. The newly added initial service-account key expiration path still has no focused request/response coverage: the existing expiration test exercises service_accounts.api_keys.create, not service_accounts.create (see the existing inline thread). Also, the shared moderation-model documentation now labels a link as “available models” but points it back to the moderation guide in api_reference/openapi.transformed.yml; that link should target the models documentation. Please correct the source and regenerate.

PR #701, opened 2026-09-15
examples: add signed SIP webhook lifecycle worker

jbeckwith-oai:
## Summary

Adds a runnable signed-webhook SIP worker that uses the existing Ruby webhook, Realtime call-control, and sideband APIs. The worker authenticates a configured application tenant, verifies the raw delivery, deduplicates event/call attempts, and owns cleanup only after successful accept. A definitive 409 never hangs up another handler's call; ambiguous accept failures remain quarantined for reconciliation. Bounded hangup preserves the primary error.

The runbook explains the dedicated-project tenant boundary, Realtime versus Live contract, private stdin framing, and the limits of process-local claims. Response/playback terminal events do not end the call. The example is excluded from general live E2E because it requires trusted webhook and SIP infrastructure. An example-local RBI declares the optional Async constants for the existing standalone typecheck. No SDK API, generated code, media, WebRTC, dependency, or CI workflow changes.

## Validation

- Dedicated deterministic suite: 20 tests, 236 assertions.
- Realtime and webhook regression suites: 258 tests, 1,783 assertions.
- Ruby 4.0.6 locked bundle; Rubyfmt, focused RuboCop, standalone Sorbet typecheck, directive validation, and example inventory pass.
- No live SIP execution; tests use fake signatures, HTTP responses, and sockets.


examples/realtime/sip_webhook.rb
```diff
@@ -0,0 +1,211 @@
...
+                end
+                # response.done and output_audio_buffer.stopped finish a turn,
+                # not the call. Keep observing until EOF, error, or our deadline.
+              end
+            end
+          end
+
+          :sideband_closed
+        ensure
+          cleanup(call_id, $!) if owned
```
markstuart-oai: [P2] Track this invocation's failure instead of ambient `$!`

Ruby's `$!` also retains an exception being handled by a caller's enclosing `rescue`. If an embedding application calls `worker.handle(envelope)` from such a rescue, accept and sideband can complete normally while this passes the unrelated exception into cleanup. A failed hangup is then swallowed at line 170. I reproduced the same signed delivery and successful accept/EOF with a failing hangup: outside a rescue it raises; inside a rescue for an unrelated error it returns `:sideband_closed` while state is `:cleanup_unconfirmed`. That falsely reports the documented successful-cleanup outcome and hides the need for reconciliation. Track normal completion or capture only an exception raised by this lifecycle, preserving cancellation handling, and cover the caller-rescue case.

markstuart-oai (commented):
The worker keeps signature verification, tenant authority, atomic local claims, and call ownership in clear boundaries, and the runbook accurately limits process-local deduplication and ambiguous-accept recovery. One P2 remains: an unrelated handled exception in an embedding caller can suppress a failed hangup and produce a false successful outcome. See the inline reproduction.

Validation: 21 lifecycle tests / 282 assertions passed, including actual Async cancellation and elapsed cleanup deadlines. Separate synthetic tenant-isolation, signature/framing, inventory and standalone Sorbet checks passed. The additional caller-rescue reproduction exposes the finding. Verified CI logs at the reviewed head/base merge: 1,977 tests on each of Ruby 3.3/3.4/4.0, no failures/errors, plus Bedrock, lint, types and package checks. No live SIP/provider/API execution; carrier behavior and model availability remain unverified.

examples/realtime/sip_webhook.rb
jbeckwith-oai: Fixed in ebe81cbdf853672547de60511801ccac5be126f9. The lifecycle now records normal completion and passes no primary error to hangup cleanup on that path, so an unrelated caller-rescue `$!` cannot hide a failed hangup. Exceptional exits still preserve this invocation's primary error or cancellation.

Added two focused regressions: successful accept/EOF followed by failing hangup inside an unrelated caller rescue (confirmed failing before the fix), and this invocation's Async cancellation plus failing hangup inside an enclosing rescue. Dedicated tests: 20 tests / 236 assertions; Realtime and webhook regression suites: 258 / 1,783. Sorbet, Rubyfmt, RuboCop and directive validation pass. Two fresh independent paired review rounds are clean, including security review. Exact-head CI is running.

markstuart-oai (approved):
Re-reviewed the two-file correction and author response. The prior cleanup finding is resolved: normal completion explicitly passes no primary error, so a caller's unrelated rescue cannot suppress failed hangup; exceptional exits still preserve this invocation's error or cancellation. The focused completion flag makes that distinction directly, without changing call ownership or broadening scope. No new findings.

The original signed-delivery reproduction now raises the hangup failure both inside and outside a caller rescue. Validation also passed 20 dedicated tests / 236 assertions and three real Async/deadline probes / 74 assertions. Verified current CI logs at the reviewed head/base merge: 1,979 tests on each of Ruby 3.3/3.4/4.0 with no failures/errors, plus Bedrock, lint, types and package checks. The prior review's process-local ownership and live-validation limits still apply; no live SIP/provider/API test was run.

PR #702, opened 2026-09-15
ci: adopt generated Ruby lint policy

jbeckwith-oai:
## Summary

The SDK ships a generated RuboCop policy but does not inherit it. Inherit that policy and enforce `Lint/NumberConversion`, retaining its exact webhook timestamp compatibility exception.

Replace the 14 handwritten number-conversion offenses with strict parsing or type-preserving conversion. Decimal refresh-buffer strings retain their decimal interpretation, numeric inputs retain their behavior, and the streaming line decoder no longer converts an already-integer offset. Unsafe number-conversion autocorrection stays disabled.

## Generator impact

No generator change is needed. The inherited fragment is byte-identical to the existing canonical projection. The root config and all modified auth, helper, internal, script, and test paths are excluded from generation. Generated webhook code remains unchanged: malformed timestamps retain the established `InvalidWebhookSignatureError` / "Webhook timestamp is too old" behavior. This addresses the policy-adoption follow-up from #486 without changing that contract.

Baseline: 15 `Lint/NumberConversion` offenses (14 handwritten plus the existing generated exception). Final: 0 offenses, with only `lib/openai/resources/webhooks.rb` excluded from number conversion. No new suppression or generated-file edit.

## Deferred cop

`Lint/UnusedPrivateMethod` is explicitly disabled. In pinned RuboCop 1.90.0 it requires project indexing and the optional `rubydex` gem; an enabled flag alone does not enforce it. A probe with an unreferenced private method was inspected but reported zero offenses with the current index-free bundle. Adding indexing and its dependency requires a separate rollout, so this PR makes no unused-private-method enforcement claim.

Policy digests, clean-generation infrastructure, plugins, and metrics are separate follow-ups.

## Validation

- `bundle exec rake lint`: 3,454 Ruby/RBI files clean; Sorbet and 1,530 RBS validations pass, including formatter and directive checks.
- `./scripts/test`: 1,962 tests / 16,167 assertions, zero failures or errors, two skips, on Ruby 4.0.6.
- `bundle exec rake build:gem`: passes.
- NumberConversion policy regressions fail before adoption and pass afterward; they verify effective enablement and real enforcement outside the exact exclusion. The temporary source-tree fixture includes both policy files.
- Two consecutive clean independent adversarial-review rounds, including security-sensitive auth, timeout, and streaming paths.


.rubocop.yml
```diff
@@ -181,6 +187,8 @@ Lint/UnreachablePatternBranch:
   Enabled: true
 Lint/UnmodifiedReduceAccumulator:
   Enabled: true
+Lint/UnusedPrivateMethod:
+  Enabled: true
```
markstuart-oai: [P2] Make the unused-private-method check actually run, or defer its adoption

In the pinned RuboCop 1.90.0, this cop immediately returns unless a project index exists. The effective `AllCops/UseProjectIndex` remains `false`, and `rubydex` is absent from the bundle, so `Enabled: true` still provides no enforcement. I confirmed an inspected file containing an unreferenced private method exits successfully with zero offenses. The new test only asserts the Enabled flag, so it misses this. Either configure the required index/dependency and add a behavioral regression that detects a dead private method, or keep this cop explicitly deferred alongside the other index-dependent cops and narrow the rollout claims to NumberConversion.

markstuart-oai (commented):
The NumberConversion adoption is scoped correctly: the effective configuration preserves only the generated webhook exception, disables unsafe autocorrection, and the runtime conversions preserve supported numeric inputs. One P2 remains in the advertised UnusedPrivateMethod enforcement; see the inline comment.

Validation: checked the exact pinned RuboCop implementation and reproduced the missing offense with an unused private method; compared effective base/head configuration and generated-policy ownership. Ran 210 focused runtime tests (one opt-in audio skip) with documented local Ruby launcher/load-path adjustments, plus 22 webhook tests. The README dispatch subprocess could not run in this container because its stripped environment cannot load libruby. Verified CI logs for the reviewed head/base merge: 1,962 tests on each of Ruby 3.3/3.4/4.0, with no failures or errors; lint, types, package, and Bedrock checks also passed. No live API calls.

.rubocop.yml
jbeckwith-oai: Fixed in aab9de148b0831c14342b136a3b01bdc0283c4f3. Confirmed with the exact RuboCop 1.90.0 bundle: Enabled=true, UseProjectIndex=false, rubydex absent; an inspected file containing an unreferenced private method reported zero offenses. Lint/UnusedPrivateMethod is now explicitly deferred with its indexing/dependency prerequisite documented. I removed the misleading Enabled assertion and narrowed the PR description/tests to NumberConversion, whose tests exercise actual enforcement outside the generated webhook exception. Adding project indexing and rubydex is outside this rollout. No generator change is needed. Full lint/format/type checks, package build, and all 1,962 tests / 16,167 assertions pass (two skips); two fresh consecutive paired adversarial rounds, including security review, are clean.

markstuart-oai (approved):
Re-reviewed the correction and author response. The prior finding is resolved: UnusedPrivateMethod is explicitly deferred with its index/rubydex prerequisites documented, and the tests and PR description now claim only NumberConversion enforcement. The effective configuration retains NumberConversion, disables unsafe autocorrection, and preserves exactly the generated webhook exception. No new findings.

The two-file correction leaves the previously reviewed runtime, dependencies and generated ownership unchanged. Focused policy tests pass (2 tests / 9 assertions), including real offenses outside the exception. Verified current CI logs for the reviewed head/base merge: 1,962 tests on each of Ruby 3.3/3.4/4.0 with no failures/errors, plus Bedrock, lint, types and package checks. The prior runtime review and its documented local-environment limitations still apply; no live API validation was performed.

PR #709, opened 2026-09-17
chore(api): clarify incoming SIP call ID usage

jbeckwith-oai:
## Summary

Clarifies how to use incoming SIP call identifiers with the Realtime and Live APIs. This update changes documentation only.

## Changes

- Document that incoming call IDs should be passed unchanged to Realtime accept and reject requests.
- Direct Live API callers to the separate session ID from the incoming transport event.


marcuswood-oai (approved):
Looks good. Reviewed the full diff: the call ID guidance matches the source OpenAPI update, with no SDK runtime or type changes. CI is green at this head. Local tests not run for this documentation change.

markstuart-oai (approved):
Reviewed `dd3e832f3e5e05c0670c41007f404e559b890951`. No actionable findings. The Ruby and RBI documentation consistently says to pass the incoming Realtime call ID unchanged, and use the separate Live event's session ID for Live accept/reject. This agrees with upstream OpenAPI `514a730ba3ebf5d9cede34e8c7af3e45a52ebc3c` and the existing endpoint implementations.

Validation: independent review of the complete diff; both Ruby/RBI non-comment sources are identical and syntax checks passed; all four descriptions match the schema. Parsed YAML changes are the description, valid JSON example, and two Go-specific generation annotations, alongside updated generation metadata. Exact-head CI passed. No local runtime suite, rendered documentation build, or live SIP/API call was performed.

PR #714, opened 2026-09-17
ci: run independent Ruby post-format checks concurrently

markstuart-oai:
## Summary

Run four independent, read-only post-format checks concurrently: RuboCop inspection, suppression-directive validation, rubyfmt verification, and RBS formatting verification. Each check runs in a separate POSIX process so working-directory changes cannot affect sibling checks.

Keep writing formatters serial, replay diagnostics in task order, and propagate check failures. Clean up child process groups on completion or cancellation. Initialize child records and signal state before entering their rescue/ensure regions, so exception handlers only access initialized state. Preserve autocorrect aliases, subsequent type checks after lint failure, and the existing serial Windows path.

## Why use a subprocess wrapper?

RuboCop's Rake task runs its CLI in-process, and configuration loading can temporarily change the process-wide working directory. Separate processes isolate that state. The checks also launch descendants, so cancellation must clean up process groups rather than just the immediate children.

- **Rake `multitask`:** [runs prerequisites in Ruby threads](https://ruby.github.io/rake/Rake/MultiTask.html), which share the working directory. Directly parallelizing the existing tasks would retain that race. Subprocess-backed prerequisites are viable, but still need process-group cleanup and output/error handling.
- **The `parallel` gem:** already present through RuboCop, and supports isolated process workers. Its [interrupt handler kills worker PIDs](https://github.com/grosser/parallel/blob/master/lib/parallel.rb), rather than their entire process groups. Descendant cleanup and grouped diagnostics would still require additional handling.
- **`Process.spawn`:** Ruby's standard API supplies process isolation, per-child working directories, output redirection and process groups. The helper uses it for four fixed checks and adds lifecycle and diagnostic handling without introducing a worker-pool dependency.

The alternatives were assessed through documentation and source inspection, not comparative benchmarks. This explains the current choice without claiming a smaller equivalent wrapper is impossible.

## Validation

Regression coverage exercises overlapping execution, formatter ordering, working-directory isolation, failure reporting, launch failures, descendant cleanup, and cancellation during launch, cleanup and signal-handler restoration. The scheduling and formatting-policy suites passed on this branch: **26 tests / 413 assertions**, zero failures, errors or skips. Windows was not locally exercised; it retains the serial path. Public CI will validate this change against the current base.


jbeckwith-oai (approved):
Reviewed exact head `25e07b20ae07748a8341ce0d341d5a8d7fee39a4`. The POSIX coordinator runs the four read-only checks in isolated process groups, replays diagnostics deterministically, preserves the Windows serial path and autocorrect aliases, propagates failures while still allowing typechecks, and closes child groups across normal completion, launch failure, and cancellation. All three prior threads are resolved and exact-head required CI is green. I also verified Ruby syntax locally; the focused suite could not be rerun here because the locked gems are not installed.

PR #718, opened 2026-09-18
fix(api)!: correct model types and network policy docs

markstuart-oai:
Carry property descriptions into Ruby, RBI and RBS attribute and initializer documentation. For hosted environments and template creation/update, this preserves the GA versus alpha/beta network defaults and distinguishes omitted update values from explicit `null` resets.

Also clarifies Realtime item-creation events and agent-session cancellation and deletion.

**Breaking change:** removes unsupported `gpt-5.1-mini` values from generated model types. This model was never released, and the removal is intentional.

Validation: the corresponding generated candidate passes Ruby 3.3, 3.4 and 4.0 tests, lint, RBI/RBS validation, package checks, CodeQL and the existing custom-code budget. Public checks run on this PR's updated head.


lib/openai/models/beta/agents/environments/template_create_params.rb
```diff
@@ -156,7 +156,7 @@ class Network < OpenAI::Internal::Type::BaseModel
               module Access
                 extend OpenAI::Internal::Type::Enum
 
-                # Allows unrestricted network access, matching an omitted network policy.
+                # Allows unrestricted network access.
```
jbeckwith-oai, line 162: [P2] Preserve per-operation network defaults in the generated field docs

Removing the generic enum-level default is appropriate, but `TemplateCreateParams.network` and its initializer parameter still only say “Network access for an OpenAI-hosted environment.” This hunk therefore removes the omission-default hint without exposing the replacement guidance from this PR's API reference: create/hosted-environment defaults differ between GA and alpha/beta, and template updates distinguish omission (preserve) from explicit `null` (reset). Please carry those existing property descriptions into the Ruby attribute/initializer docs and corresponding RBI/RBS surfaces, including the hosted-environment and update models. Restoring the old blanket enum claim would misdescribe updates; the field-specific wording is needed so callers can determine what omitting or resetting a network policy actually does.

jbeckwith-oai (commented):
Reviewed the complete 18-file diff at `3f9549a5d04a481b525f05b80d7040235f925117` for source-contract accuracy, compatibility, and the networking documentation implications.

Every non-comment line in the changed Ruby, RBI, and RBS files is identical to the base. The API-reference delta matches the corresponding public OpenAPI changes, and its recorded hash is valid. GitHub reports a successful exact-head check rollup. I ran read-only source/reference checks, not the full SDK test suite or a live backend probe.

One P2 documentation-propagation issue is noted inline; no additional in-scope findings.

dpiet-oai (approved):

PR #725, opened 2026-09-19
feat(api): add safety case retrieval

jbeckwith-oai:
## Summary

Adds support for retrieving safety case details by ID.

## Changes

- Add client.safety.cases.retrieve for retrieving a safety case by ID
- Expose the case identifier, creation time, entity identifier, nullable reason, and warning or deactivation notice


markstuart-oai (approved):
Reviewed exact head `33b8956e01c4d73e4263d075c98c923ac7b910f7`. Reviewed `33b8956e01c4d73e4263d075c98c923ac7b910f7` against `8d1179aab1cebd3b394d5c91cc48ba447be1993a`. No actionable findings. Service loading and RBI/RBS signatures agree; the new method uses existing path encoding, bearer-only authentication, options and response conversion. Public-entrypoint coverage checks sync/async retrieval, nullable reasons, unknown fields and credential separation. No structural regression or new 1,000-line crossing.

Validation: exact-source static review plus successful hosted Ruby 3.3/3.4/4.0 tests, type checks, lint, package and Castiron checks. No local execution.

Hosted check runs at this head have passed or been skipped, and both commit-status checks pass. Some integrations have queued check suites without runs; I am not treating those as completed checks.

PR #726, opened 2026-09-19
chore(api): document response management resources

jbeckwith-oai:
## Summary

Adds documentation for response-management resources and their Sorbet interfaces.

## Changes

- Describe response, input-item, and input-token resources in the client and Sorbet interfaces


markstuart-oai (approved):
Reviewed all changed files at `678881b6ce0c`; no actionable findings. The OpenAPI delta only adds the Responses resource description and tags existing stable/beta token-count and compaction operations. Checked every Ruby/RBI pair; the new summaries align and all other source and signatures remain unchanged.

Verified hosted Ruby 3.3/3.4/4.0 tests, packaging, lint, RBI/RBS checks and Castiron baseline/budget checks at this head. Validation here was static comparison of pinned source plus hosted results; I did not rerun tests or generation locally.

PR #729, opened 2026-09-20
fix(api)!: represent multi-graders as named grader maps

markstuart-oai:
## Summary

Changes multi-grader inputs to keyed collections of graders.

## Changes

- Use dictionary keys as variables in multi-grader formulas.

## Migration and validation

Pass a named grader hash, for example `graders: {quality: grader}`, and refer to the same key in `calculate_output: "quality"`. The single-value union is now `MultiGrader::Grader` (previously `MultiGrader::Graders`); its RBS alias likewise changes from `graders` to `grader`.

Pre-publication validation of this updated candidate passed the complete Ruby 3.3, 3.4 and 4.0 suites, Bedrock tests, lint, RBI/RBS checks and package build. A local full-import consumer passed 26 assertions, including all five grader variants, typed round trips and three captured SDK requests; existing session-reset controls passed 16 assertions. No live API calls or package releases were made.

The committed regression tests cover both string and symbol dictionary keys, all five typed values, round trips, exact Run/Validate request envelopes, and invalid flat/array controls. They pass 3 tests / 27 assertions on this candidate and fail against the unchanged public baseline. Formatting, focused lint, Sorbet and RBS checks pass locally.

Public CI passed Ruby 3.3/3.4/4.0, Bedrock, lint, RBI/RBS validation and packaging checks, including the new named-map regression tests.


dpiet-oai (approved):

dpiet-oai (approved):

jbeckwith-oai (changes requested):
not as a breaking change, nah

markstuart-oai:
In response to [the nonbreaking-release review](https://github.com/openai/openai-ruby/pull/729#pullrequestreview-5274230859):

Agreed — this should not ship as a breaking release. The current diff really does remove `MultiGrader::Graders` (including its RBI/RBS types) and replace the single-grader field contract with a named map, so just removing `!` from the title would hide an incompatibility.

I’m keeping this PR in draft with your change request outstanding. It needs an additive compatibility revision that retains the existing public names and legacy input behavior while supporting correctly named grader maps, plus regression coverage for both paths. The old flat payload must not be silently given an invented grader name or represented as an API-valid request. No compatibility fix or release is claimed yet.

